PR #88, opened 2026-04-03
docs: workflow improvements design proposal

engalar:
## Summary

- Proposes ALTER WORKFLOW with graph-aware operations for incremental workflow modification
- Adds workflow cross-reference tracking in catalog `refs` table for SHOW CALLERS/CALLEES support

## Motivation

Workflow support has full CREATE/DESCRIBE/DROP coverage, but two gaps remain:
1. No incremental modification — any change requires full `CREATE OR MODIFY` rebuild
2. Workflows are invisible to cross-reference queries (`SHOW CALLERS`, `SHOW IMPACT`)

## Design Highlights

**ALTER WORKFLOW** supports two categories of operations:
- **Property ops**: SET DISPLAY, DESCRIPTION, EXPORT LEVEL, activity-level PAGE/TARGETING
- **Graph ops**: INSERT AFTER (linear), INSERT OUTCOME/PATH/BRANCH (at split points), DROP ACTIVITY/OUTCOME/PATH/BRANCH — each with well-defined BSON transformations and merge point discovery

**Cross-References** extends the existing catalog builder to track workflow→microflow, workflow→workflow, workflow→entity, and workflow→page references.

See `docs/plans/2026-04-03-workflow-improvements-design.md` for full proposal including BSON transformation details, merge point algorithm, ANTLR grammar, and 6-phase implementation plan.

## Test plan

- [ ] Review design document completeness
- [ ] Validate graph operation BSON transformations against Studio Pro behavior
- [ ] Confirm merge point discovery algorithm handles nested splits

### DIFF
--- a/docs/11-proposals/PROPOSAL_workflow_improvements.md
+++ b/docs/11-proposals/PROPOSAL_workflow_improvements.md
@@ -1,0 +1,466 @@
+# Workflow Improvements: ALTER WORKFLOW + Cross-References
+
+**Date:** 2026-04-03
+**Status:** Proposal
+**Author:** @engalar
+
+## Problem
+
+Workflow support in mxcli has full CREATE/DESCRIBE/DROP/SHOW coverage with 13 activity types and BSON round-trip fidelity. Two significant gaps remain:
+
+1. **No ALTER WORKFLOW** — any workflow change requires a full `CREATE OR MODIFY` rebuild. For small edits (change a task page, add an outcome, insert an activity), this is disproportionate effort and error-prone.
+
+2. **No cross-reference tracking** — the catalog `refs` table does not track workflow references. `SHOW CALLERS OF Module.SomeMicroflow` will not show workflows that call it. Impact analysis is incomplete.
+
+## Design
+
+### Part 1: ALTER WORKFLOW
+
+#### 1.1 Property Operations
+
+Modify workflow-level and activity-level metadata without touching the flow graph:
+
+```sql
+-- Workflow-level properties
+ALTER WORKFLOW Module.MyWorkflow
+  SET DISPLAY 'New Display Name';
+
+ALTER WORKFLOW Module.MyWorkflow
+  SET DESCRIPTION 'Updated description';
+
+ALTER WORKFLOW Module.MyWorkflow
+  SET EXPORT LEVEL API;
+
+ALTER WORKFLOW Module.MyWorkflow
+  SET DUE DATE '[%CurrentDateTime%] + 7 * 24 * 60 * 60 * 1000';
+
+ALTER WORKFLOW Module.MyWorkflow
+  SET OVERVIEW PAGE Module.NewOverviewPage;
+
+ALTER WORKFLOW Module.MyWorkflow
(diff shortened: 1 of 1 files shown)


ako (commented):
## Review

### Part 2 (Cross-References) looks ready

The cross-reference addition to the catalog builder is straightforward, independent, and high-value. It plugs workflows into the existing `SHOW CALLERS/CALLEES/IMPACT` infrastructure with no new commands needed. P1 implementation priority is right.

### Part 1 (ALTER WORKFLOW) - concerns

**Merge point algorithm is fragile.** The simplified algorithm assumes each branch is a linear chain (`while current has exactly one outgoing edge`). This breaks for nested splits — if a branch contains a Decision or ParallelSplit, the inner split's outgoing edges would cause the traversal to stop prematurely. The algorithm needs to handle nested structured blocks, either by tracking split/merge depth or using proper post-dominator analysis.

**Activity references by bare name are ambiguous.** The grammar uses `IDENTIFIER` to reference activities (`SET ACTIVITY userTask1`, `INSERT AFTER activityName`). Workflow activity names are user-assigned display labels, not guaranteed unique. Two activities could share the same name. The design should specify behavior for duplicates — error? First match? — or use a different addressing scheme (e.g., positional index, qualified path like `parallelSplit1/branch1/userTask1`).

**DROP PATH by 0-based index is brittle.** `DROP PATH 2 ON parallelSplit1` depends on knowing the internal ordering of parallel paths. If a prior DROP shifts indices, scripts that drop multiple paths need careful reverse-ordering. Consider using a name or label instead, similar to DROP OUTCOME which uses a string name.

**INSERT AFTER doesn't show how the new activity connects to existing context.** The example `INSERT AFTER activityName CALL MICROFLOW validate Module.Validate` uses inline activity syntax, but the grammar rule references `workflowActivity` which isn't defined in this document. Should clarify whether this reuses the same activity syntax from CREATE WORKFLOW or is a subset.

**No REPLACE operation.** ALTER PAGE has REPLACE for swapping one widget for another. A `REPLACE ACTIVITY oldName WITH ...` would be a natural addition — currently you'd need DROP + INSERT AFTER the predecessor, which requires knowing the predecessor name.

### Minor issues

- Author is `@anthropics/claude-code` — should be the human author
- The `workflowBlock` grammar rule uses `'{'` and `'}'` literal characters instead of `LBRACE`/`RBRACE` tokens — should use the existing lexer tokens for consistency
- No mention of how `DESCRIBE WORKFLOW` output changes to support roundtripping ALTER operations (e.g., should DESCRIBE show activity names that can be used with ALTER?)

### What looks good

- Clean separation of property ops (simple, low-risk) from graph ops (complex, high-risk)
- BSON transformation diagrams for each operation are clear and implementable
- "Operations NOT supported" section is honest about scope limits
- Testing strategy covers the right edge cases
- Phase ordering is sensible — cross-refs first, then properties, then graph ops by complexity

### Verdict

Part 2 (cross-references) is solid and can proceed to implementation. Part 1 (ALTER WORKFLOW) needs the merge point algorithm and activity addressing concerns resolved before implementation — these are design-level issues that would be expensive to fix after building. The property operations (P2/P3) are fine as-is since they don't touch the graph.

PR #164, opened 2026-04-09
fix: run mx update-widgets before mx check to prevent false CE0463

### LINKED ISSUE #121: [error] [CE0463] "The definition of this widget has changed. Update this widget by right-clicking it and selecting 'Update widget', or select 'Update all widgets' to update all widgets in the app."
❯ &"D:\Program Files\Mendix	.6.4\modeler\mx.exe" --help
List of available commands:

  check                       Check the app for issues
  convert                     Convert Mendix project
  create-project              Create a new Mendix project
  update-widgets              Update widgets in Mendix project to match widget definitions in mpks
  collect-native-deps         Collect native dependencies
  show-version                Show the version of Studio Pro that was last used to edit the app
  show-app-version            Show the Application Version of the app
  set-app-version             Sets the Application Version of the app
  merge                       Merge mpr files
  diff                        Diff mpr files
  show-module-version         Show the version of the module specified
  set-module-version          Set the version of the module specified
  create-project-package      Export a project package
  create-solution-package     Export a solution package
  create-module-package       Export a module package
  module-import               Import module package into passed .mpr file
  dump-mpr                    Export a model as JSON data
  analyze-mpr                 Analyze the storage used by an MPR file
  rename-design-properties    Update renamed Design Properties
  show-java-version           Show the configured Java version of the app
  sync-java-dependencies      Synchronize managed Java dependencies
  translate                   Export and import translatable texts for an App. Note Add-On and Solution modules will not be exported.
  git-merge                   Merge conflicting mpr file
  export-security-overview    Export the security overview
  show-private-values         Show private values
  delete-private-values       Delete private values

---
_Copied from engalar/mxcli#36_

engalar:
## Summary
- Run `mx update-widgets` automatically before `mx check` in `docker check` and `docker build` commands to normalize pluggable widget definitions against installed `.mpk` files
- Added `--no-update-widgets` flag to opt out
- Failure of update-widgets is non-fatal (warns and continues)

## Context
CE0463 ("The definition of this widget has changed") fires when widget Object properties don't match Type PropertyTypes. This is a known issue with mxcli-generated pluggable widgets. Running `mx update-widgets` normalizes the Objects as a workaround.

A follow-up PR will address the root cause in the widget template engine.

Closes #121

## Test plan
- [x] Unit tests verify `update-widgets` runs before `check` by default
- [x] Unit tests verify `--no-update-widgets` skips the step
- [x] `go build ./...` and `go vet` pass
- [ ] Manual test with a project that has pluggable widgets

### DIFF
--- a/cmd/mxcli/docker.go
+++ b/cmd/mxcli/docker.go
@@ -139,14 +139,16 @@ Examples:
 		outputDir, _ := cmd.Flags().GetString("output")
 		dryRun, _ := cmd.Flags().GetBool("dry-run")
 		skipCheck, _ := cmd.Flags().GetBool("skip-check")
+		noUpdateWidgets, _ := cmd.Flags().GetBool("no-update-widgets")
 
 		opts := docker.BuildOptions{
-			ProjectPath: projectPath,
-			MxBuildPath: mxbuildPath,
-			OutputDir:   outputDir,
-			DryRun:      dryRun,
-			SkipCheck:   skipCheck,
-			Stdout:      os.Stdout,
+			ProjectPath:       projectPath,
+			MxBuildPath:       mxbuildPath,
+			OutputDir:         outputDir,
+			DryRun:            dryRun,
+			SkipCheck:         skipCheck,
+			SkipUpdateWidgets: noUpdateWidgets,
+			Stdout:            os.Stdout,
 		}
 
 		if err := docker.Build(opts); err != nil {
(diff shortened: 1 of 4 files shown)


ako (approved):
Clean fix with the right approach — `mx update-widgets` normalizes pluggable widget Objects against `.mpk` PropertyTypes, preventing false CE0463. PR correctly notes this is a workaround pending a root-cause fix in the widget template engine.

**What's good:**
- `SkipUpdateWidgets` option threaded through both `BuildOptions` and `CheckOptions`
- `--no-update-widgets` flag on both `docker build` and `docker check`
- Failure is non-fatal (warns and continues) — correct, since worst case is the original CE0463
- Tests use fake shell scripts to verify command order (update-widgets before check) and the opt-out flag

**Minor:**
- Flag naming `--no-update-widgets` is slightly inconsistent with neighboring `--skip-check`. `--skip-update-widgets` would match better.
- No test for the failure path (`update-widgets` exits non-zero → warning printed, check still runs)
- No test for the `Build()` path even though the same logic was added there

LGTM.

PR #165, opened 2026-04-09
fix: association navigation expression missing target entity and extra spaces

### LINKED ISSUE #120: Association navigation expression: missing target entity and extra spaces
## Bug

When using association navigation in microflow expressions, mxcli generates expressions with two issues:

### 1. Missing target entity qualifier

**Generated:**
```
$Order / Module.Order_Customer / Name
```

**Expected (valid Mendix expression):**
```
$Order/Module.Order_Customer/Module.Customer/Name
```

The Mendix expression engine requires the full path: `$var/Association/TargetEntity/Attribute`. mxcli omits the target entity name, causing **CE0117** ("Error(s) in expression").

Since the association definition already contains the target entity, mxcli should auto-resolve it during code generation.

### 2. Extra spaces around path separators

**Generated:**
```
$Order / Module.Order_Customer / Name
```

**Expected:**
```
$Order/Module.Order_Customer/Module.Customer/Name
```

The spaces around `/` are cosmetic but may also contribute to Mendix expression parser failures.

### Root cause

The MDL-to-Mendix expression translator does not:
1. Look up the association's target entity to insert it into the path
2. Strip whitespace around `/` path separators

### Repro

```mdl
CREATE PERSISTENT ENTITY MyModule.Customer (Name: String(200));
CREATE PERSISTENT ENTITY MyModule.Order (OrderNumber: String(50));
CREATE ASSOCIATION MyModule.Order_Customer
  FROM MyModule.Order TO MyModule.Customer TYPE Reference;

CREATE MICROFLOW MyModule.TestNav()
RETURNS String AS $Result
BEGIN
  RETRIEVE $Order FROM MyModule.Order LIMIT 1;
  DECLARE $CustName String = $Order/MyModule.Order_Customer/Name;
  RETURN $CustName;
END;
/
```

`DESCRIBE MICROFLOW` output shows `$Order / MyModule.Order_Customer / Name` — fails CE0117 on `mx check`.

### Expected fix

mxcli should generate: `$Order/MyModule.Order_Customer/MyModule.Customer/Name`

---
_Copied from engalar/mxcli#38_

engalar:
## Summary

Fixes #120 — association navigation expressions in microflows generated incorrect Mendix expressions causing CE0117.

- **Extra spaces**: `$Order / Module.Order_Customer / Name` — parser misidentified path as division operators
- **Missing target entity**: `$Order/Module.Order_Customer/Name` — Mendix requires `$Order/Module.Order_Customer/Module.Customer/Name`

## Changes

- **Parser fix** (`visitor_microflow_expression.go`): `tryBuildAttributePath` now handles `IdentifierExpr` and `QualifiedNameExpr` as path components, correctly parsing association navigation as `AttributePathExpr`
- **Builder fix** (`cmd_microflows_builder.go`): Added `exprToString()` → `resolveAssociationPaths()` pipeline on `flowBuilder` that auto-inserts the target entity after association segments using the existing `lookupAssociation` infrastructure
- **Call site updates** (5 builder files): Replaced `expressionToString()` with `fb.exprToString()` for model-aware expression serialization
- **Tests** (`bugfix_test.go`): 3 new test functions covering parsing, path resolution, and no-spaces serialization

## Test plan

- [x] `go build ./...` — clean
- [x] `go test ./...` — 19/19 packages pass
- [x] New tests verify: `$Order/Test.Order_Customer/Name` parses as `AttributePathExpr` (not `BinaryExpr`)
- [x] New tests verify: `expressionToString` produces no extra spaces
- [ ] Integration test with real MPR: create entities + association + microflow with association navigation, run `mx check`

### DIFF
--- a/mdl/executor/bugfix_test.go
+++ b/mdl/executor/bugfix_test.go
@@ -422,3 +422,140 @@ func validateMicroflowFromMDL(t *testing.T, input string) []string {
 
 	return ValidateMicroflowBody(stmt)
 }
+
+// TestAssociationNavParsing verifies that $Var/Module.Assoc/Attr parses as
+// AttributePathExpr (not nested BinaryExpr with "/" operator).
+// Issue #120: extra spaces around path separators.
+func TestAssociationNavParsing(t *testing.T) {
+	input := `CREATE MICROFLOW Test.MF_Nav()
+RETURNS String AS $Result
+BEGIN
+  DECLARE $CustName String = $Order/Test.Order_Customer/Name;
+  RETURN $CustName;
+END;`
+
+	prog, errs := visitor.Build(input)
+	if len(errs) > 0 {
+		t.Fatalf("Parse error: %v", errs[0])
+	}
+
+	stmt := prog.Statements[0].(*ast.CreateMicroflowStmt)
+	declStmt := stmt.Body[0].(*ast.DeclareStmt)
+
+	// The expression should be an AttributePathExpr, not a BinaryExpr
+	pathExpr, ok := declStmt.InitialValue.(*ast.AttributePathExpr)
+	if !ok {
+		t.Fatalf("Expected AttributePathExpr, got %T", declStmt.InitialValue)
+	}
+
+	if pathExpr.Variable != "Order" {
+		t.Errorf("Variable = %q, want %q", pathExpr.Variable, "Order")
+	}
+	if len(pathExpr.Path) != 2 {
+		t.Fatalf("Path length = %d, want 2", len(pathExpr.Path))
+	}
+	if pathExpr.Path[0] != "Test.Order_Customer" {
+		t.Errorf("Path[0] = %q, want %q", pathExpr.Path[0], "Test.Order_Customer")
+	}
+	if pathExpr.Path[1] != "Name" {
+		t.Errorf("Path[1] = %q, want %q", pathExpr.Path[1], "Name")
+	}
+
+	// Serialized form should have no extra spaces
+	got := expressionToString(pathExpr)
+	want := "$Order/Test.Order_Customer/Name"
+	if got != want {
+		t.Errorf("expressionToString() = %q, want %q", got, want)
+	}
+}
+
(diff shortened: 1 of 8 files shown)


ako (approved):
Two clear bugs solved with minimal, focused changes.

### Bug 1: Extra spaces around `/`

Root cause: parser didn't recognize qualified names as path components, falling through to multiplicative operator rule and producing nested `BinaryExpr{Op: "/"}`. Fix in `tryBuildAttributePath` correctly handles `IdentifierExpr` and `QualifiedNameExpr`.

### Bug 2: Missing target entity qualifier

Root cause: Mendix expressions require `$Order/Mod.Assoc/Mod.Customer/Name` but AST only stores `[Mod.Assoc, Name]`. Fix introduces `fb.exprToString()` → `resolveAssociationPaths()` that walks the expression tree and inserts the target entity for each association segment.

### What's good

- Comprehensive call-site update — all 27+ uses of `expressionToString` in builder files switched to `fb.exprToString`
- "Already resolved" heuristic prevents double-insertion when path is already qualified
- Nil-reader fallback keeps tests/standalone use working
- Recursive expression walking handles nested expressions (BinaryExpr, FunctionCallExpr, IfThenElseExpr, etc.)

### Concerns

**`TestResolveAssociationPaths` doesn't actually test resolution.** It sets `reader: nil` which short-circuits the function, then only verifies the length stays the same. The `assoc_then_attr` case never verifies `Test.Customer` is actually inserted. A fake reader (or real MPR) would close this gap — the PR test plan already acknowledges this with the unchecked integration test box.

**Inverse navigation may pick the wrong entity.** `lookupAssociation` returns `childEntityQN`, but if someone navigates from the child side (`$Customer/Mod.Order_Customer/...`), the target should be `parentEntityQN`. Worth verifying whether `lookupAssociation` handles direction or if this is an unhandled edge case.

LGTM — the fix is correct for the common case. Recommend addressing the test gap and verifying inverse navigation in a follow-up.

PR #166, opened 2026-04-10
fix: augment nested ObjectType properties from .mpk to prevent CE0463

engalar:
## Summary
- Extended `AugmentTemplate` to recursively sync nested ObjectType PropertyTypes (e.g., DataGrid2 column properties) with `.mpk` definitions
- Fixes the root cause of CE0463 when installed widget version differs from embedded template version
- Handles add/remove of nested properties in both Type schema and Object values

## Root Cause
When the project's `.mpk` widget version differs from the embedded template version, `AugmentTemplate` only synced top-level properties. Nested ObjectType properties (like DataGrid2 column properties: tooltip, exportValue, showContentAs) were left with the original template's schema, causing Type/Object property count mismatches that trigger CE0463.

## What Changed
- `augment.go`: Added `augmentNestedObjectType()` that processes `.mpk` children for Object-type properties
- Fixed early return that skipped nested processing when top-level had no changes
- `augment_test.go`: Added test for nested ObjectType property addition

## Relationship to #164
PR #164 adds `mx update-widgets` as a workaround before `mx check`. This PR fixes the underlying cause so the workaround is needed less often.

Refs mendixlabs/mxcli#121

## Test plan
- [x] Existing augmentation tests pass (8/8)
- [x] New nested ObjectType augmentation test passes
- [x] `go build ./...` and `go vet` pass
- [ ] Integration test with DataGrid2/ComboBox/Gallery on mismatched widget versions

### DIFF
--- a/sdk/widgets/augment.go
+++ b/sdk/widgets/augment.go
@@ -87,8 +87,17 @@ func AugmentTemplate(tmpl *WidgetTemplate, def *mpk.WidgetDefinition) error {
 		}
 	}
 
-	// Nothing to add/remove
-	if len(missing) == 0 && len(stale) == 0 {
+	// Check if nested augmentation is needed (skip early return if so)
+	hasNestedChildren := false
+	for _, p := range def.Properties {
+		if len(p.Children) > 0 {
+			hasNestedChildren = true
+			break
+		}
+	}
+
+	// Nothing to add/remove at top level, and no nested children to process
+	if len(missing) == 0 && len(stale) == 0 && !hasNestedChildren {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


ako (approved):
Substantial fix targeting the root cause that PR #164 was working around. The approach is correct.

### What's good

- **Recursive sync of nested ObjectType PropertyTypes** correctly handles add/remove for both Type schema and Object Properties
- **Early-return fix is critical**: previously `if len(missing) == 0 && len(stale) == 0 { return nil }` would skip nested processing when top-level was already in sync. Now checks `hasNestedChildren`.
- **Exemplar-based cloning** with `createPropertyPair` fallback
- **TypePointer matching** correctly identifies which Object Properties to update for stale removal

### Concerns

**Test coverage is thin.** Only one test case (add missing). Missing:
- Removing stale nested properties (the ~50 lines of removal logic)
- Multiple WidgetObject instances (e.g., DataGrid with 3 columns — does the new property get added to all 3?)
- No-op case when nested is already in sync (regression guard for the early-return change)

**Recursion depth.** The function syncs one level of nesting. Worth confirming this is sufficient for all Mendix widget specs — does any widget have nested-nested ObjectTypes?

**Unhandled: type change between versions.** If `.mpk` changes a child from `string` → `textTemplate`, the function neither removes nor updates the type. Probably out of scope but worth noting.

**Minor:** The `break` in `augmentNestedObjectType`'s objProps loop after finding matched TypePointer would benefit from a comment explaining why (only one property matches a given TypePointer).

LGTM — fix is correct for the documented scenario. Recommend expanding test coverage for stale removal and multi-instance cases in a follow-up.

PR #171, opened 2026-04-10
feat: support full user targeting on workflow USER TASK

### LINKED ISSUE #169: Support TARGET USERS clause on workflow USER TASK definitions
## Problem

When creating workflows via MDL, there is no way to configure the "Target users using" property on `USER TASK` definitions. This means workflow user tasks cannot be claimed by anyone at runtime — the `WorkflowUserTask_TargetUsers` association remains empty.

Currently, the only workaround is to open Studio Pro and manually set the "Target users using" property on each user task after creating the workflow via mxcli.

## Current behavior

`DESCRIBE WORKFLOW` output shows no target user configuration on user tasks:

```sql
USER TASK ReviewTask 'Review Request'
    PAGE MyModule.TaskPage
    DESCRIPTION 'Please review and decide'
    OUTCOMES
      'Approve' { ... }
      'Reject' { ... };
```

Attempting to add `TARGET USERS` fails with a parse error (`'Target' is a reserved keyword`).

## Expected behavior

Support a `TARGET USERS` clause on `USER TASK` definitions, for example:

```sql
USER TASK ReviewTask 'Review Request'
    PAGE MyModule.TaskPage
    DESCRIPTION 'Please review and decide'
    TARGET USERS XPATH '[System.UserRoles = ''[%UserRole_Manager%]'']'
    OUTCOMES
      'Approve' { ... }
      'Reject' { ... };
```

Or with a microflow:

```sql
    TARGET USERS MICROFLOW MyModule.GetTargetUsers
```

## Additional context

- `GRANT EXECUTE ON WORKFLOW` is supported and works — this sets workflow-level access rules
- But user task-level targeting is separate and equally important for tasks to be actionable
- Without this, every workflow created via mxcli requires a Studio Pro visit to configure task assignment
- Mendix 11.x workflow model supports XPath and microflow-based user targeting on user tasks

engalar:
Fixes #169

## Summary
- Parser now reads both legacy `UserSource` and current `UserTargeting` BSON fields, recognizing all five `$Type` variants (`MicroflowUserTargeting`, `XPathUserTargeting`, `MicroflowGroupTargeting`, `XPathGroupTargeting`, `NoUserTargeting`)
- Added `MicroflowGroupSource` and `XPathGroupSource` SDK types with full BSON serialization/deserialization
- MDL grammar extended with optional `USERS`/`GROUPS` modifier: `TARGETING [USERS|GROUPS] MICROFLOW/XPATH`
- DESCRIBE output now correctly displays all targeting types including group targeting

## Syntax

```sql
-- User targeting (existing, backward compatible)
TARGETING MICROFLOW Module.GetUsers
TARGETING XPATH '[System.UserRoles = ''[%UserRole_Manager%]'']'

-- Explicit user targeting (new)
TARGETING USERS MICROFLOW Module.GetUsers
TARGETING USERS XPATH '[...]'

-- Group targeting (new)
TARGETING GROUPS MICROFLOW Module.GetGroups
TARGETING GROUPS XPATH '[GroupType = ''Reviewers'']'
```

## Test plan
- [x] 6 new parser tests covering all targeting types (user microflow/xpath, group microflow/xpath, no targeting, legacy UserSource)
- [x] `mxcli check` validates all syntax variants
- [x] Full test suite passes
- [x] Build succeeds


ako (approved):
Clean fix with thoughtful backward compatibility.

### What's good

**Backward compatibility done right.** Parser reads both legacy `UserSource` and current `UserTargeting` fields, mapping old and new `$Type` variants to the same SDK types — projects from any Mendix version round-trip.

**Grammar extension is non-breaking.** `TARGETING (USERS | GROUPS)?` keeps existing scripts working.

**Full pipeline coverage** across all 7 layers (SDK, parser, writer, grammar, AST, visitor, executor, formatter).

**Good test coverage** — 6 parser tests covering all targeting variants including the legacy field.

### Concerns

**`XPathConstraint` fallback may be reversed.** Parser checks `XPath` first, then falls back to `XPathConstraint`, but the writer always emits `XPathConstraint`. If real Mendix BSON uses `XPathConstraint`, the `XPath` check is dead code; if both exist in different versions, the priority should match what's actually used.

**Group source types may be missing fields.** `MicroflowGroupSource` only stores `Microflow`. If group targeting has additional fields (e.g., context variable like user targeting may have), they'd be silently lost. Worth verifying against a real Studio Pro export.

**`MULTI USER TASK` grammar duplication.** Identical targeting blocks in both `USER TASK` and `MULTI USER TASK` rules — could be factored into a `userTargetingClause` rule. Not blocking.

**Visitor's group detection is coarse.** `len(utCtx.AllGROUPS()) > 0` flags both microflow and xpath as group if either uses GROUPS. Probably impossible in practice (only one targeting clause allowed) but worth a guard.

**No DESCRIBE roundtrip test.** Tests cover parsing only. A test that creates → describes → re-parses would catch formatter mismatches.

LGTM — fix is correct, concerns are minor improvements.

ako (commented):
## PR Review: feat: support full user targeting on workflow USER TASK

**Overall**: Clean, well-scoped PR that adds group targeting support for workflow user tasks. The implementation follows existing patterns correctly and the test coverage is solid. A few items to address:

### Issues

**1. DESCRIBE output for existing user targeting types misses `USERS` modifier**

The DESCRIBE formatter (`cmd_workflows.go:426-433`) outputs `TARGETING MICROFLOW` and `TARGETING XPATH` for existing user-targeting types, but the new syntax supports `TARGETING USERS MICROFLOW`. While backward-compatible (the grammar makes `USERS`/`GROUPS` optional), this creates an asymmetry: group targeting always outputs `TARGETING GROUPS MICROFLOW` but user targeting never outputs `TARGETING USERS MICROFLOW`. If a user reads DESCRIBE output, they might not realize the `USERS` modifier exists. Consider whether DESCRIBE should emit the explicit form for clarity once group targeting is in play.

**2. Writer serializes `UserTargeting` but parser reads both `UserSource` and `UserTargeting` — field name inconsistency**

The writer (`writer_workflow.go`) serializes into the `UserTargeting` BSON field (the current name), which is correct. The parser reads both `UserSource` (legacy) and `UserTargeting` (current) — also correct. However, there's no comment or constant explaining which Mendix versions use which field name. The PR description mentions "Mendix 10.12+" for `UserTargeting` — a brief comment in `parser_workflow.go` at the fallback would help future maintainers.

**3. `NoUserTargeting` not explicitly handled in `parseUserSource`**

The `default` case in `parseUserSource` (`parser_workflow.go:598`) returns `&workflows.NoUserSource{}` which implicitly handles `Workflows$NoUserTargeting`. This works, but adding it as an explicit case alongside `NoUserSource` would make the code self-documenting and prevent accidental regression if someone later adds logging for the `default` case.

**4. Grammar allows both `TARGETING USERS MICROFLOW` and `TARGETING USERS XPATH` on the same task**

The grammar rule has two separate optional clauses:
```
(TARGETING (USERS | GROUPS)? MICROFLOW qualifiedName)?
(TARGETING (USERS | GROUPS)? XPATH STRING_LITERAL)?
```
This means a user could write *both* on the same task. The visitor code would set `Kind` twice (once to microflow, then overwritten by xpath). This is pre-existing behavior, not introduced by this PR, but worth noting.

**5. Missing BSON fields for group sources**

`MicroflowGroupTargeting` and `XPathGroupTargeting` — the parser doesn't read `XPath` as a fallback field for `MicroflowGroupTargeting` (unlike the existing `MicroflowBasedUserSource` which reads `MicroflowExpression` as fallback). Are there any additional BSON fields on these types? If you've verified against real Studio Pro output, this is fine — but worth confirming.

### Positives

- **Backward compatibility**: Legacy `UserSource` field is still read, with `UserTargeting` as fallback — existing MPR files continue to work.
- **Test coverage**: 6 well-structured tests covering all targeting types including legacy.
- **Full-stack wiring**: Grammar, visitor, AST, executor (read + write), parser, writer, SDK types, and LSP completions all updated consistently.
- **Clean scope**: Single feature, single commit — follows the PR checklist.

### Checklist items to verify

- [ ] `make grammar` was run (generated files are included, looks correct)
- [ ] No `mdl-examples/doctype-tests/` test files for the new syntax variants — the PR checklist says MDL syntax changes should include working examples
- [ ] No version-gating for group targeting — if this requires Mendix 10.12+, should `checkFeature()` be called?
- [ ] Quick reference (`docs/01-project/MDL_QUICK_REFERENCE.md`) not updated with the new `GROUPS` modifier

ako:
## Follow-up review after ef0f3e5

The new commit adds `GROUPS` to the `keyword` rule in `MDLParser.g4` — correct fix for `TestKeywordRuleCoverage`. No issues with this change.

The substantive items from my [previous review](https://github.com/mendixlabs/mxcli/pull/171#pullrequestreview) still apply:

1. **DESCRIBE asymmetry** — user targeting outputs `TARGETING MICROFLOW` while group targeting outputs `TARGETING GROUPS MICROFLOW`. Consider emitting `TARGETING USERS MICROFLOW` for symmetry.
2. **Version comment** — the `UserSource` → `UserTargeting` field name transition (Mendix 10.12+) deserves a comment in `parser_workflow.go`. *(Partially addressed — the comments are there now but a version note would help.)*
3. **Explicit `NoUserTargeting` case** — still handled implicitly by the `default` branch in `parseUserSource`.
4. **Missing `mdl-examples/doctype-tests/`** — no working MDL examples for the new syntax variants.
5. **No version-gating** — if group targeting requires Mendix 10.12+, should `checkFeature()` guard it?
6. **Quick reference not updated** — `docs/01-project/MDL_QUICK_REFERENCE.md` doesn't mention the `GROUPS` modifier.

Otherwise LGTM — the core implementation is solid.

engalar:
image

ako:
## Review

**Overall: solid work, small fixes needed.** Clean backward-compat handling for both legacy `UserSource` and current `UserTargeting` BSON fields, all 5 `$Type` variants covered, full pipeline wired (lexer → grammar → AST → visitor → executor read+write → tests). The CI failure is partly self-inflicted (new test) and partly pre-existing on main.

### What I like
- **Bidirectional support**: parser handles both `Workflows$MicroflowBasedUserSource` (legacy) and `Workflows$MicroflowUserTargeting` (10.12+), writer emits the modern names. Default `NoUserTargeting` is also written when no source is set — that's the correct serialization.
- **Grammar change is surgical**: one new `GROUPS` token, optional `(USERS | GROUPS)?` modifier on existing TARGETING clause. `USERS` was already a token so no new keyword shadow risk there.
- **Parser test coverage**: 6 tests across all 5 `$Type` variants plus the legacy field name — exactly the surface area that needs guarding against silent regressions.

### Concerns

1. **CI failure on the new test (`workflow-user-targeting.mdl`)** — fails with `[CE1613] "The selected entity 'System.UserGroup' no longer exists."` in Mendix 11.9.0. The targeting microflow `WFTarget.ACT_GetGroups` returns `List of System.UserGroup`, but that entity is gone in 11.x. Use whatever the modern workflow group entity is (likely `System.WorkflowUserTaskGroup` or check Mendix docs for the 11.x system module). The other two failing tests (`14-project-settings-examples.mdl`, `keyword-as-identifier.mdl`) are pre-existing failures on main — not blocking this PR.

2. **Visitor's `isGroupTargeting` is one boolean across both TARGETING clauses** (`visitor_workflow.go:440`). Grammar lets you write `TARGETING USERS MICROFLOW X` and `TARGETING GROUPS XPATH Y` independently; `len(AllGROUPS()) > 0` then poisons the MICROFLOW classification. Won't bite valid inputs but is fragile. Either reject mixed forms in the executor or compute group-ness per clause.

3. **No mutual exclusion of TARGETING clauses.** Grammar permits both `TARGETING ... MICROFLOW` and `TARGETING ... XPATH` on the same task; only one can be valid in Mendix. Add an executor-side check with a clear error.

4. **ALTER WORKFLOW grammar doesn't support USERS/GROUPS** (`MDLParser.g4:2834-2835`). You can CREATE a task with `TARGETING GROUPS MICROFLOW` but can't ALTER a user-targeted task into a group-targeted one. Either extend the alter rule symmetrically or note as a follow-up.

5. **DESCRIBE roundtrip not explicitly verified.** PR summary says DESCRIBE displays group targeting — but is the output re-executable MDL? Worth a roundtrip test (write → describe → re-execute → verify equal).

6. **Inconsistent SDK naming**: legacy types are `MicroflowBasedUserSource` / `XPathBasedUserSource`, new ones are `MicroflowGroupSource` / `XPathGroupSource` (no "Based"). Either add "Based" for consistency or strip it from the legacy ones in a follow-up rename. Also the `UserSource` interface name now hosts group sources too — semantically a misnomer; consider `TaskTargeting` / `Targeting` to match the modern BSON term. Bigger refactor — defer.

7. **No writer roundtrip test for group targeting.** Read path is well tested; write → read → assert would catch BSON drift on the group variants.

### Verdict
Approve once (1) is fixed. (2)–(4) should be addressed before merge or filed as follow-ups; (5)–(7) are improvements.

ako:
HI @engalar i cant merge this at the moment as is has merge conflicts. Can you update the PR?

PR #174, opened 2026-04-10
fix: add OPEN to keyword rule for enum value parsing

engalar:
## Summary

- `OPEN` was a lexer token (used by `OPEN USER TASK`, `OPEN WORKFLOW`) but missing from the parser's `keyword` rule
- This caused 3-part enum values like `Module.ENUM_IssueStatus.Open` to fail parsing in `CREATE`/`CHANGE` statements within microflow bodies
- The fix adds `OPEN` to the keyword alternatives in `MDLParser.g4` line 3592

## Reproduction

```sql
CREATE OR REPLACE MICROFLOW MyModule.Test ()
RETURNS Nothing
BEGIN
  $Issue = CREATE MyModule.OrderIssue (
    IssueNumber = 'TEST',
    Status = MyModule.ENUM_IssueStatus.Open,
    IssueType = MyModule.ENUM_IssueType.DeliveryDelay
  );
END;
```

**Before:** `line 6:47 mismatched input '.' expecting ')'`
**After:** `Syntax OK`

## Test plan

- [x] `mxcli check` passes for enum values with `Open` (previously reserved keyword)
- [ ] Existing tests still pass
- [ ] `OPEN USER TASK` and `OPEN WORKFLOW` syntax unaffected


ako (approved):
Surgical, correct fix. The diagnosis is right — `OPEN` was added as a lexer token in #159 but never wired into the `keyword` rule, breaking enum values like `Module.ENUM_Status.Open`. This is exactly the friction I flagged in the #159 review about common-word tokens.

### Concerns

**Test plan boxes are unchecked.** `[ ] Existing tests still pass` and `[ ] OPEN USER TASK syntax unaffected` should be verified before merge. The change is generally safe (allows OPEN as identifier in identifier-position contexts) but worth confirming.

**No regression test added.** A doctype-test with an enum value named `Open` would prevent regression — even one line in an existing test file would do it.

**Same class of issue likely affects other tokens from #159.** That PR added `DATA`, `RECORDS`, `NOTIFY`, `PAUSE`, `UNPAUSE`, `ABORT`, `RETRY`, `RESTART`, `LOCK`, `UNLOCK`, `REASON` — common words that could appear as enum values, attribute names, etc. I'd recommend either:
1. Adding all of them to the `keyword` rule in this PR (zero risk, broad coverage)
2. Filing a follow-up to audit and add regression tests

LGTM for the immediate fix.

engalar:
Superseded by #186 which adds all 168 missing keyword tokens (including OPEN).

PR #179, opened 2026-04-10
fix: support @Module.Const syntax in expressions

### LINKED ISSUE #178: Constant reference round-trip broken: DESCRIBE outputs @Module.Const but parser rejects @ syntax
## Problem

`DESCRIBE MICROFLOW` outputs constant references using `@Module.ConstantName` syntax, but this syntax is rejected by the parser when used in `CREATE MICROFLOW` or `CREATE OR REPLACE MICROFLOW`.

This breaks the round-trip guarantee: you cannot take the output of DESCRIBE and feed it back into CREATE.

### Reproduction

```sql
-- 1. Create a microflow that references a constant (using the working syntax)
CREATE MICROFLOW MyModule.Test ()
BEGIN
  DECLARE $val String = getConstantValue('MyModule.MyConstant');
  RETURN;
END;

-- 2. DESCRIBE outputs:
--    DECLARE $val String = @MyModule.MyConstant;

-- 3. Copy-paste the DESCRIBE output into a new CREATE statement → parse error:
--    extraneous input '@' expecting { ... }
```

### Expected behavior

Either:
1. **Parser should accept `@Module.Const` syntax** (matching DESCRIBE output), OR
2. **DESCRIBE should output `getConstantValue('Module.Const')`** (matching what the parser accepts)

Option 1 is preferred — `@Module.Const` is cleaner and the DESCRIBE output should be valid MDL.

### Workaround

Use `getConstantValue('Module.ConstantName')` instead of `@Module.ConstantName` when writing MDL scripts.

### Impact

Anyone who uses DESCRIBE output as a starting point for modifying microflows (a common workflow) will hit a cryptic parse error. Especially confusing because the `@` syntax looks intentional and correct.

engalar:
Fixes #178

## Summary
- Added `@QualifiedName` as a valid `atomicExpression` in the MDL grammar for constant references
- New `ConstantRefExpr` AST node type for clean representation
- Visitor converts `@ Module.Const` parse tree to `ConstantRefExpr`
- `expressionToString()` outputs `@Module.Const` for BSON serialization
- Backward compatible: `getConstantValue('Module.Const')` still works

## Round-trip restored
```sql
-- DESCRIBE outputs:
DECLARE $val String = @MyModule.MyConstant;

-- Can now be fed back into CREATE without modification ✓
```

## Test plan
- [x] `mxcli check` accepts `@Module.Const` syntax
- [x] `mxcli check` still accepts `getConstantValue('Module.Const')` syntax  
- [x] Full test suite passes
- [x] Build succeeds


ako (approved):
Surgical, minimal fix that restores DESCRIBE roundtrip for microflows referencing constants.

### What's good

- Full-stack change in 5 files: 1 grammar line, 1 AST type, 1 visitor case, 2 formatter cases
- Backward compatible — `getConstantValue('Mod.Const')` still works
- Reuses existing `AT` token (no lexer changes)

### Concerns

**Potential parser ambiguity with `@annotation`.** The grammar already uses `AT IDENTIFIER` for annotations like `@Documentation`. Adding `AT qualifiedName` to `atomicExpression` should be fine since contexts don't overlap (annotations are statement-level, expressions are body-level), but a test confirming this would close the gap.

**No regression tests.** The PR's main value is DESCRIBE roundtrip but there's no test verifying parse → describe → re-parse for a microflow with `@Module.Const`. A doctype-test entry would prevent regression.

**Two `expressionToString` functions in different files** both got the new case. This kind of duplication is a maintenance hazard but outside the scope of this PR.

**`QualifiedNameExpr` change in `cmd_diff_mdl.go`** is unrelated to the constant reference fix — looks like a separate bug being piggybacked. Should be its own commit ideally.

LGTM.

PR #186, opened 2026-04-10
fix: add 168 missing lexer tokens to keyword rule

engalar:
## Summary
- The `keyword` rule in `MDLParser.g4` allows lexer tokens to be used as identifiers (entity/attribute/enum names via `qualifiedName`). **168 word-type tokens were missing**, causing parse failures when user-defined names matched keywords like `Data`, `Filter`, `Match`, `Empty`, `Open`, `Container`, `Node`, `Activity`, etc.
- Reorganized the rule by category with alphabetical ordering for easy auditing
- Removed 3 duplicate entries (`STRUCTURES`, `PAGING`, `EXECUTE`) and 1 phantom token (`UI`)

## Before
```sql
-- Parse error: "extraneous input '.' expecting {',', ')'}"
$Issue = CREATE MyModule.OrderIssue (
    Status = MyModule.ENUM_IssueStatus.Open
);
```

## After
```sql
-- All keyword-named enum values now parse correctly
$Issue = CREATE MyModule.OrderIssue (
    Status = MyModule.ENUM_IssueStatus.Open,     -- was broken
    Kind = MyModule.ENUM_Kind.Data,               -- was broken
    Mode = MyModule.ENUM_Mode.Filter,             -- was broken
    State = MyModule.ENUM_State.Empty,            -- was broken
    Type = MyModule.ENUM_Type.Container,          -- was broken
    Phase = MyModule.ENUM_Phase.Match,            -- was broken
    Method = MyModule.ENUM_M.Get                  -- was broken
);
```

## Scope
Only the `keyword` rule in `MDLParser.g4` changed (+ regenerated parser files). No executor, visitor, or AST changes.

Follow-up from PR #174 code review which identified this systemic gap.

## Test plan
- [x] `make test` — all tests pass (including previously failing `TestQuotedIdentifierInWidgetAttribute` and `TestShowPageMicroflowStyleArgsInWidget` which were broken during development by accidentally dropping `LAYOUT`)
- [x] Smoke tested 15+ keyword-as-enum-value cases (Data, Filter, Match, Empty, Container, Open, Node, Activity, Condition, Get, Post, etc.)
- [x] No original keyword tokens dropped (verified via `comm -23`)
- [x] No duplicate tokens in final rule (verified via `uniq -d`)


ako (approved):
This is exactly the broad fix I recommended in the PR #174 review.

### What's good

- **Solves the systemic problem**: 168 word-type tokens missing from `keyword` is a significant gap. This affects enum values, attribute names, entity names, anywhere `qualifiedName` is used.
- **Categorical organization** with 18 groups + alphabetical ordering makes future audits trivial — easy to find where to add new tokens
- **Cleanup**: removes 3 duplicates (`STRUCTURES`, `PAGING`, `EXECUTE`) and 1 phantom (`UI`)
- **Pure grammar change** — no executor/visitor/AST, lowest possible risk
- **Verification approach documented**: `comm -23` for missing tokens, `uniq -d` for duplicates

### Concerns

**No regression tests added.** "Smoke tested 15+ cases" is good for confidence but doesn't prevent regression. A test file with `CREATE ENUMERATION Test.E (Open, Data, Filter, Match, Empty, Container, Node, ...)` would catch any future drift.

**No automated lexer/keyword sync check.** A `make` target that grep's word-type tokens from `MDLLexer.g4` and verifies each is in the `keyword` rule (or explicitly excluded) would prevent this from happening again. Worth a follow-up issue.

**PR #174 becomes a partial duplicate** since this includes the `OPEN` fix. Should be closed when this merges.

LGTM.

ako (approved):
Both review concerns fully addressed in the two new commits:

1. **Automated sync test** (`keyword_coverage_test.go`) parses `MDLLexer.tokens` and verifies every word-type token is in the `keyword` rule — this prevents the class of bug from ever recurring. Also catches stale/extra entries. Exactly what I recommended.

2. **Regression test file** (`keyword-as-identifier.mdl`) with 117 lines of enumerations using keyword-named values (`Open`, `Data`, `Filter`, `Match`, `Get`, `Post`, `Activity`, `Layout`, `Header`, etc.).

3. **Unified identifier rules** to use the full keyword list — good consistency improvement.

LGTM.

PR #195, opened 2026-04-13
fix: prefer Studio Pro over CDN downloads on Windows

engalar:
## Summary
- On Windows, CDN-downloaded mxbuild is a Linux ELF binary that cannot execute natively. Reorder resolution priority so Studio Pro installations (`D:\Program Files\Mendix\{version}\`) are found before cached CDN downloads.
- Unify all Windows path discovery through `windowsProgramDirs()` — eliminates hardcoded `C:\` drive letters, uses `PROGRAMFILES`/`PROGRAMW6432`/`SystemDrive` env vars.
- Fix `filepath.Join` bug where `SystemDrive` ("D:" without separator) produced relative paths like "D:Program Files".

## Test plan
- [x] `TestWindowsProgramDirs_UsesEnvVars` — all returned paths are absolute
- [x] `TestWindowsProgramDirs_NoDuplicates` — no duplicate entries
- [x] `TestWindowsProgramDirs_SystemDriveFallback` — fallback works when PROGRAMFILES vars are unset
- [x] `TestMendixSearchPaths_NoHardcodedDriveLetter` — no hardcoded C: paths on non-C: systems
- [x] `TestJdkSearchPaths_NoHardcodedDriveLetter` — same for JDK paths
- [x] `TestResolveStudioProDir_FakeInstall` — finds fake Studio Pro layout
- [x] `TestResolveMxBuild_PrefersStudioProOverCache` — Studio Pro wins over cached Linux binary
- [x] `mxcli new TestApp --version 11.6.4` — uses Studio Pro directly, no CDN download
- [x] `mxcli docker run` — full build + container startup with Studio Pro mxbuild.exe

### DIFF
--- a/cmd/mxcli/cmd_new.go
+++ b/cmd/mxcli/cmd_new.go
@@ -57,12 +57,21 @@ Examples:
 			os.Exit(1)
 		}
 
-		// Step 1: Download MxBuild
-		fmt.Printf("Step 1/4: Downloading MxBuild %s...\n", mendixVersion)
-		mxbuildPath, err := docker.DownloadMxBuild(mendixVersion, os.Stdout)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Error downloading MxBuild: %v\n", err)
-			os.Exit(1)
+		// Step 1: Resolve MxBuild and mx binary.
+		// On Windows, prefer Studio Pro installation (ships both mxbuild.exe and mx.exe).
+		// Fall back to CDN download on other platforms or when Studio Pro is not installed.
+		fmt.Printf("Step 1/4: Resolving MxBuild %s...\n", mendixVersion)
+		var mxbuildPath string
+		if studioDir := docker.ResolveStudioProDir(mendixVersion); studioDir != "" {
+			mxbuildPath = filepath.Join(studioDir, "modeler", "mxbuild.exe")
+			fmt.Printf("  Using Studio Pro: %s\n", studioDir)
+		} else {
+			var err error
+			mxbuildPath, err = docker.DownloadMxBuild(mendixVersion, os.Stdout)
+			if err != nil {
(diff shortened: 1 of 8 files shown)


ako:
Do you think it's a good idea to have claude code run on a windows system without a devcontainer?

engalar:
> Do you think it's a good idea to have claude code run on a windows system without a devcontainer?

To clarify: this PR still uses Docker Compose to run the Mendix app — that doesn't change. The difference is **where the build happens**.

**Current (broken on Windows):** mxcli downloads Linux mxbuild from CDN → can't execute → build fails → nothing to put in Docker.

**With this PR:** mxcli uses Studio Pro's native mxbuild.exe to build the PAD package on the host → the built artifact runs in Docker Compose as before.

The build step (mxbuild) must run on the host because it needs access to the `.mpr` file, Java, and the full Mendix model. The runtime step (Docker Compose) is already containerized. This is the natural split — build natively where the tools are, run in containers where isolation matters.

**Why not build inside a devcontainer instead?**

1. **Performance** — mxbuild reads the `.mpr` SQLite database and compiles Java. Doing this across the WSL2/9P filesystem bridge is significantly slower than native.

2. **Complexity** — a devcontainer for building needs: mxbuild (Linux), JDK 21, the `.mpr` file mounted in, and Docker socket access for the subsequent `docker compose up`. That's Docker-in-Docker territory — fragile and hard to debug (we spent significant time in this session debugging Docker proxy issues through the WSL2 network stack).

3. **Studio Pro already installed** — mxbuild.exe is literally sitting in `Program Files\Mendix\{version}\modeler\`. Downloading a separate Linux copy to use inside a container is redundant.

4. **macOS/Linux unaffected** — on those platforms the CDN download works natively, no change needed. This PR only adds a "check Studio Pro first" step on Windows.

PR #204, opened 2026-04-15
fix: microflow layout issues (ConnectionIndex, redundant Merge, DESCRIBE roundtrip)

### LINKED ISSUE #132: Nested IF/ELSE generates ugly microflow layout with excessive decision splits
## Problem

When a microflow uses a single `RETURN` at the end with a result variable (instead of multiple `RETURN` statements in each branch), mxcli still generates nested IF/ELSE decision diamonds. This produces a microflow layout in Studio Pro with many crossing/diverging sequence flows that is hard to read.

## Example

```mdl
CREATE OR REPLACE MICROFLOW OrderManagement.ChatContext_ActionMicroflow (
  $ChatContext: ConversationalUI.ChatContext
)
RETURNS Boolean AS $Success
FOLDER 'Chat'
BEGIN
  DECLARE $Success Boolean = false;

  $Request = CALL MICROFLOW ConversationalUI.ChatContext_Preprocessing("ChatContext" = $ChatContext);

  IF $Request != empty THEN
    RETRIEVE $Agent FROM AgentCommons.Agent
      WHERE Title = 'SupplyChainOrderAgent'
      LIMIT 1;

    IF $Agent != empty THEN
      $Response = CALL JAVA ACTION AgentCommons.Agent_Call_WithHistory(...);

      IF $Response/GenAICommons.Response_Message != empty THEN
        $Message = CALL MICROFLOW ConversationalUI.ChatContext_UpdateAssistantResponse(...);
        SET $Success = true;
      END IF;
    ELSE
      LOG ERROR NODE 'OrderManagement' 'SupplyChainOrderAgent not found';
    END IF;
  END IF;

  RETURN $Success;
END;
```

This MDL is logically clean — one entry, one exit, a single `RETURN $Success` at the end. But the generated microflow in Studio Pro has 3 nested decision diamonds with sequence flows that zigzag across the canvas, making it visually cluttered.

## Expected behavior

Nested IF blocks that converge to a single exit point should produce a clean layout where the "false" branches merge back into the main flow without excessive crossings. Ideally the layout engine would:

1. Keep the happy path on the main horizontal/vertical axis
2. Route false/empty branches as short side-paths that rejoin the main flow
3. Minimize sequence flow crossings

## Screenshot

_(To be added manually)_

## Environment

- mxcli: 0.1.0
- Mendix: 11.6.4

---
_Copied from engalar/mxcli#47_

### Comments from original issue

**@engalar** (2026-04-08T01:59:23Z):
Image

engalar:
Fixes #132

## Summary

- **ConnectionIndex int64/int32 mismatch**: Writer stored `OriginConnectionIndex`/`DestinationConnectionIndex` as `int64`, but parser read with `.(int32)` type assertion — silently failed, all anchors defaulted to 0 (Top). Connections pointed at wrong edges after any read-modify-write cycle.
- **Redundant Merge nodes**: Guard checks (`IF cond THEN RETURN; END IF`) created `ExclusiveMerge` with in-degree=1, wasting layout space. Now skipped when only one branch flows into the merge.
- **DESCRIBE roundtrip**: Fixed `$` prefix on RETURN expressions, trailing newline in DECLARE constants, and guard pattern detection (IF/RETURN/END IF without ELSE). DESCRIBE output is now re-parseable.

## Insight

`ALTER MICROFLOW` preserves `@position` from original activities, preventing Studio Pro from applying auto-layout. DESCRIBE should omit positions to let the layout engine or Studio Pro compute optimal positions.

## Test plan

- [x] ConnectionIndex roundtrip verified with `Tool_SendOrderToSupplier` (28 flows, correct R/L/B/T anchors)
- [x] Guard chain (3 sequential guards): 0 redundant Merge nodes (was 3)
- [x] DESCRIBE → parse → execute roundtrip: `Tool_SendOrderToSupplier4` created successfully
- [x] Studio Pro layout verified: auto-layout produces clean, readable microflows


ako:
## Review

**Overall: solid fixes on real bugs, two concerns.** The three headline fixes (ConnectionIndex type, redundant Merge elimination, DESCRIBE RETURN prefix) are all legitimate. Plus this PR fixes the `settingsSection` grammar gap I flagged on PR #171. But: scope creep and a couple of rough edges.

### What I like
- **ConnectionIndex int32 fix is right** (`parser_microflow.go:900`, `writer_microflow.go:919`). Writer was emitting `int64` while parser type-asserted `int32` — every read-modify-write silently dropped to 0 (Top anchor). Parser now uses existing `extractInt` (handles int32/int64/int/float64), writer emits `int32` to match Studio Pro. Backward-read-compatible with any files the old writer produced.
- **`needMerge` logic is correctly derived**: `bothReturn → 0 inputs, no merge, terminate`; `hasElseBody → merge iff neither returns`; `no-else → merge iff THEN continues`. Matches the case analysis cleanly.
- **`nextFlowCase` dispatch** (`cmd_microflows_builder_control.go:236-244`) correctly labels the deferred split edge for all three no-merge cases (else+thenReturns, else+elseReturns, no-else).
- **Grammar fix for `MODEL`**: adding `| MODEL` to `settingsSection` fixes the pre-existing `14-project-settings-examples.mdl` failure that was breaking main's CI. Bonus value.

### Concerns

1. **Scope creep — undocumented deletions in `keyword-as-identifier.mdl`.** The PR removes `Owner`, `Type`, `Object` from both the enumeration and the entity in the test, without any mention in the PR description. That's a test specifically designed to *assert* these words work as identifiers. If they don't, the fix should be in the grammar/reserved-word handling; if they truly can't, the removal deserves a paragraph explaining which reserved-word list changed and why. Silently narrowing a test hides a regression. Either split this out or explain it.

2. **`!strings.ContainsAny(returnVal, "+'\"()")` heuristic is incomplete** (`cmd_microflows_format_action.go:~28`). It misses common operators and punctuation: `-`, `*`, `/`, `<`, `>`, `=`, `!`, `,`, `[`, `]`, space. So `RETURN count - 1` becomes `RETURN $count - 1`, `RETURN var1 = var2` becomes `RETURN $var1 = var2`. Either widen the char set or invert: only prefix when the whole value matches `^[A-Za-z_][A-Za-z0-9_]*$` (a single bare identifier).

3. **Dead if/else in position calc** (`cmd_microflows_builder_control.go:~229`):
   ```go
   if !hasElseBody {
       fb.posX = max(afterSplit, afterBranch)
   } else {
       fb.posX = max(afterSplit, afterBranch)
   }
   ```
   Both branches identical — collapse to one line. Suggests the author intended different positions originally; either implement the divergence or drop the if.

4. **Test coverage is manual-only.** PR cites one microflow (`Tool_SendOrderToSupplier`), one guard chain, one Studio Pro layout check. No automated tests for: ConnectionIndex roundtrip, guard-pattern DESCRIBE, RETURN-prefix heuristic, `needMerge` branch matrix. This is exactly the kind of fix that should have a regression test alongside it — these bugs were silent before, they'll be silent again if reintroduced.

5. **Merge conflict (DIRTY)** — needs rebase before merge.

### Verdict
Approve after (1) is clarified/split, (2) fixed, (3) cleaned up. (4) and (5) before merge.

PR #214, opened 2026-04-16
Feedback on agent document proposal

tinaengel-mx:


docs/11-proposals/PROPOSAL_agent_document_support.md
tinaengel-mx, line 255: I don't think entity & variables are super relevant for an overview

tinaengel-mx, line 45: afaik customblobdocuments only have these properties exposed 
        $ID: "",
        name: "",
        excluded: false,
        documentType: agentDocumentTypeId, //i.e. agenteditor.agent
        contents: createNewAgent(), //json

tinaengel-mx: see my comment at the top. afaik not all of these properties are exposed for custom blob documents

ako:
## Correction to my earlier review

On re-reading, I was too quick to treat Tina's edits as corrections to the BSON. The original proposal was written by inspecting a real project's BSON; I should have verified before accepting changes that contradict it.

Checked the shipping parser:

- `sdk/mpr/parser_customblob.go:26` — wrapper `$Type` is `CustomBlobDocuments$CustomBlobDocument` (matches original).
- Line 59 — parser reads `raw["CustomDocumentType"]`, not `DocumentType`.

So **the field name in the actual BSON is `CustomDocumentType`**. The edit on line 19 to `DocumentType` is almost certainly a slip (and only one of four occurrences was changed). **Revert that one.**

Stepping back, the edits in this PR are three different kinds of change that need different handling:

1. **BSON field-name edits** (e.g. `CustomDocumentType → DocumentType`): reject unless a BSON snapshot or parser line proves otherwise. The shipping parser is the ground truth here.

2. **New empirical claims about schema content** — `endpoint` / `authenticationMicroflow` on ConsumedMCPService; `UsageType: "Task" | "Chat"` vs the original `"Task" | "Conversational"`; the explanation of empty `type`/`name`/... as decoded from the key constant. These could be correct in a newer Mendix version or a different sample project than the original author examined, but they contradict the original's "observed in test3 project" baseline. **Need a real BSON snapshot from a current project before landing.** If the BSON actually contains these, great. If not, they belong in a "spec/intent" section, not in the observed-schema section.

3. **Design-intent additions** — microflow tool return-type contract (`String`, primitive inputs + `GenAICommons.Request`/`Tool`), tool-name regex, `ToolChoice: Tool` requiring `toolChoiceToolName`, auth microflow returning `List of System.HttpHeader`, `minSimilarity` range 0–1. These are product knowledge the empirical-only original couldn't have. **These should land** — they're the highest-value contribution in the PR. Just mark them clearly as "design constraints" rather than "observed schema."

My original "what's valuable" list conflated (2) and (3). Only (3) is unambiguously valuable without more evidence.

Net recommendation: revert the `DocumentType` edit, get a fresh BSON snapshot to validate the category-2 claims, and keep the category-3 additions.

docs/11-proposals/PROPOSAL_agent_document_support.md
tinaengel-mx: I can only speak for the extensions-api, what's readable from a dump-mpr might be different

tinaengel-mx, line 632: AgentEditorCommons itself has many dependencies; see https://docs.mendix.com/appstore/modules/genai/genai-for-mx/agent-editor/#dependencies

tinaengel-mx, line 772: These are actions that require an agent object (not agent document) as input. In the toolbox in the agent editor section, there is now also a call agent acitvity, which is the preferred way when using the agent editor. It is called: AgentEditorCommons.Agent_CallDocument_WithoutHistory

A with history operation will be added soon as well.

tinaengel-mx, line 783: it's probably best to let the extension handle the variable calculation. We can trigger that on chnage and on opening of the agent tab

tinaengel-mx, line 849: of add to existing after startup logic

tinaengel-mx, line 837: of course this is still possible, because all objects (including agents) will be created in the ASU, but it will be easier to just use the native agent editor call agent activities that require and Agent docment as input. This will remove the need to do a database retrieve of the agent object before passing it to the agent commons call agent actions

tinaengel-mx: only valid values will we Task or Chat; Chat will likely be introduced in v1.1.0

tinaengel-mx, line 1053: only available at runtime, will be added to extension later

tinaengel-mx, line 1074: only available at runtime, will be added to extension later

tinaengel-mx, line 2003: let me know if there are still open questions regarding microflow tools. I added a lot of info above

tinaengel-mx, line 2007: as decribed earlier, we have an will continue to introdcue decided toolbox actions in AgentEditorCommons module that accept an agent document as input param. This will simplify microflows containing call agent acitivites even though the "old" from AgentCommons actions will still work

tinaengel-mx, line 2009: currently we require a restart of the app after every change to be able to use the testing playground and male sure that agent object in the database are in sync with agent documents

tinaengel-mx, line 2011: correct we ship samples (template agents) but agents can be created in any module

tinaengel-mx, line 2013: we can handle renames of referenced documents from the extension, however we are not yet able to recover references when a document is deleted and recreated with the same name. This will result in a consistency error and the user needs to reselect the referenced document (model, constant, microflow, KB, MCP)

tinaengel-mx, line 2015: not correct. Keys are impoted right after selection of a constant. We also listen to changes to the constant values or the runtime configs. Additionally users can trigger a call to the resource by pressing the test key button, this might update the model name and ID if that was chnages in the portal in the meantime

ako:
## Review: PR #214 — Feedback on agent document proposal

**Verdict: Approve with suggestions** — excellent domain corrections that significantly improve the accuracy of the proposal. Changes are well-targeted and backed by apparent first-hand knowledge of the agent document schema.

---

### Summary

This PR corrects and extends `PROPOSAL_agent_document_support.md` with accurate field names, schema details, and MDL syntax based on the actual BSON structure of Mendix AI Agent documents. The changes are documentation-only.

---

### What's correct and valuable

**BSON field corrections:**
- `CustomDocumentType` → `DocumentType`: important fix — using the wrong field name would cause `TypeCacheUnknownTypeException` in Studio Pro
- `Excluded` field clarified as user-controllable toggle, not a pre-release artifact
- `type` field on Model now shows `"Text generation"` — this matters for roundtrip serialization

**ConsumedMCPService schema additions:**
- Adding `endpoint` (String constant ref) and `authenticationMicroflow` (optional) is a significant correction. The original proposal had these as runtime-only config, but they're actually stored in the MPR document. This affects what the executor must write.
- Updating `CATALOG.CONSUMED_MCP_SERVICES` to include `endpoint_constant` and `authentication_microflow` columns keeps the catalog schema consistent.

**Agent struct additions:**
- `temperature`, `topP`, `toolChoiceToolName` — important for prompt engineering workflows
- `minSimilarity` on KBToolRef — necessary for knowledge base retrieval tuning

**Syntax fixes:**
- `MCP SERVICE` → `CONSUMED MCP SERVICE`: the block keyword should match the document type name
- `UsageType: Conversational` → `UsageType: Chat`: aligns with actual enum value

**Typo fix:**
- `"Microlfow"` → `"Microflow"` in sample JSON

---

### Suggestions

1. **Model metadata source**: The proposal now says metadata is "decoded from key, not Portal-pushed" — it would be useful to document the key format (or a reference to where it's decoded) so implementors know where to look.

2. **`temperature` and `topP` types**: The proposal should specify whether these are `float64` or stored as integers (scaled). Mendix BSON sometimes stores floats as scaled integers. If these come from reflection data, a note on the expected storage type would prevent a silent precision bug.

3. **`ToolChoiceToolName` — conditional field**: Should the MDL syntax include a guard showing when this is required vs. optional? E.g., is it only valid when `toolChoice = "Specific"`? If so, the proposal should note that, since the executor will need to validate the combination.

4. **`AuthenticationMicroflow` on ConsumedMCPService**: Is this a qualified name reference (`Module.MicroflowName`) stored as a string, or a GUID pointer? The distinction matters for the executor (string vs. ID lookup). Worth clarifying in the schema table.

5. **DESCRIBE roundtrip**: The proposal shows `CREATE MODEL` and `CREATE AGENT` syntax but doesn't show what `DESCRIBE MODEL MyModel` or `DESCRIBE AGENT MyAgent` should output. Since the PR checklist requires DESCRIBE roundtrip, it's worth adding example output — even rough — to the proposal so the implementor knows the target.

---

### Minor

- The sample JSON blocks are helpful — consider adding a note about which Mendix version introduced agent documents (for the `-- @version:` directive implementors will need).
- The `Excluded` field: is it persisted as `bool` or as a BSON `int32` (0/1)? Mendix sometimes uses int32 for booleans in older document types.

Overall this is a high-quality corrections PR. The schema additions for ConsumedMCPService in particular change the implementation scope — `endpoint` and `authenticationMicroflow` need to be in the executor's CREATE/DROP handlers, not just the proposal doc. Worth flagging that to the implementor.

ako:
All corrections from this PR have been applied directly to `docs/11-proposals/PROPOSAL_agent_document_support.md` on the `pull` branch (commit 269ef3d). Closing this PR as the changes are incorporated.

Issue #212, opened 2026-04-16
BUG: Some "list operations" in the microflow description are wrong


rbolt:
MxCLI version: 0.6.0
Mendix version: 11.8.0
MxCli diag: not necessary

When executing the "microflow describe" command I noticed that the list operation "Equals" got a wrong description.

The output looks like:
<img width="1039" height="155" alt="Image" src="https://github.com/user-attachments/assets/41b84e5f-c34e-4989-8344-a91438f55806" />

While the microflow part looks like: 

<img width="501" height="163" alt="Image" src="https://github.com/user-attachments/assets/118e0c25-25ce-463c-8b53-8cbab3514490" />

I expect something like:
-   $ContractListEqualsObject = EQUALS($ContractListInput_1, $ContractInput);
-   $ContractListEqualsList = EQUALS($ContractListInput_1, $ContractListInput_2);

Same issue for the following list operations:
- FILTER (remark: Filter by expression is ok)
- FIND (remark; find by expression is ok)
- RANGE

PR #217, opened 2026-04-16
fix: handle Microflows$Find, $Filter, $ListRange in list operation DESCRIBE

yscraft:
## Summary

- `DESCRIBE MICROFLOW` showed `$var = LIST OPERATION ...;` for `Microflows$Find`, `Microflows$Filter`, and `Microflows$ListRange` because `parseListOperation()` had no cases for these BSON types — they fell to `default: return nil`
- Added `FindByAttributeOperation`, `FilterByAttributeOperation`, `RangeOperation` structs to `sdk/microflows/microflows_actions.go`
- Added 3 parser cases in `sdk/mpr/parser_microflow.go`
- Added 3 formatter cases in `mdl/executor/cmd_microflows_format_action.go`
- `Microflows$ListRange` stores offset/limit inside a nested `CustomRange` map

## Test plan

- [ ] 4 new formatter tests in `mdl/executor/cmd_microflows_format_listop_test.go`
- [ ] 4 new parser tests in `sdk/mpr/parser_listoperation_test.go`
- [ ] `go test ./mdl/executor/... ./sdk/mpr/...` passes

### DIFF
--- a/mdl/executor/cmd_microflows_format_action.go
+++ b/mdl/executor/cmd_microflows_format_action.go
@@ -753,6 +753,24 @@ func (e *Executor) formatListOperation(op microflows.ListOperation, outputVar st
 		return fmt.Sprintf("$%s = CONTAINS($%s, $%s);", outputVar, o.ListVariable, o.ObjectVariable)
 	case *microflows.EqualsOperation:
 		return fmt.Sprintf("$%s = EQUALS($%s, $%s);", outputVar, o.ListVariable1, o.ListVariable2)
+	case *microflows.FindByAttributeOperation:
+		ref := o.Attribute
+		if ref == "" {
+			ref = o.Association
+		}
+		parts := strings.Split(ref, ".")
+		fieldName := parts[len(parts)-1]
+		return fmt.Sprintf("$%s = FIND($%s, $currentObject/%s = %s);", outputVar, o.ListVariable, fieldName, o.Expression)
+	case *microflows.FilterByAttributeOperation:
+		ref := o.Attribute
+		if ref == "" {
+			ref = o.Association
+		}
+		parts := strings.Split(ref, ".")
+		fieldName := parts[len(parts)-1]
+		return fmt.Sprintf("$%s = FILTER($%s, $currentObject/%s = %s);", outputVar, o.ListVariable, fieldName, o.Expression)
+	case *microflows.RangeOperation:
+		return fmt.Sprintf("$%s = RANGE($%s, %s, %s);", outputVar, o.ListVariable, o.OffsetExpression, o.LimitExpression)
 	default:
 		return fmt.Sprintf("$%s = LIST OPERATION %T;", outputVar, op)
 	}
--- a/mdl/executor/cmd_microflows_format_listop_test.go
+++ b/mdl/executor/cmd_microflows_format_listop_test.go
@@ -104,3 +104,51 @@ func TestFormatListOperation_Nil(t *testing.T) {
 		t.Errorf("got %q", got)
 	}
 }
+
+func TestFormatListOperation_FindByAttribute(t *testing.T) {
+	e := newTestExecutor()
+	got := e.formatListOperation(&microflows.FindByAttributeOperation{
+		ListVariable: "Orders",
+		Attribute:    "MyModule.Order.Status",
+		Expression:   "'Active'",
+	}, "Found")
+	if got != "$Found = FIND($Orders, $currentObject/Status = 'Active');" {
+		t.Errorf("got %q", got)
+	}
+}
+
+func TestFormatListOperation_FindByAssociation(t *testing.T) {
+	e := newTestExecutor()
+	got := e.formatListOperation(&microflows.FindByAttributeOperation{
+		ListVariable: "Orders",
+		Association:  "MyModule.Order_Customer",
+		Expression:   "$Customer",
+	}, "Found")
+	if got != "$Found = FIND($Orders, $currentObject/Order_Customer = $Customer);" {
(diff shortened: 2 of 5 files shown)


ako:
## Review

**Overall: good scope, clean code, but has a DESCRIBE roundtrip break that should block merge.** The BSON types being handled are real gaps (they fell to `default: return nil` before, which rendered as `LIST OPERATION <type>;` in DESCRIBE). Parser, model, and formatter are wired symmetrically with matching tests. One concrete problem: `RANGE` isn't in the MDL grammar.

### What I like
- **Symmetric surface**: each new `$Type` gets a new SDK struct, parser case, formatter case, and both parser + formatter tests. No half-finished slices.
- **Attribute/Association disambiguation**: the parser stores both fields, and the formatter falls back cleanly when one is empty. Matches how the BSON is actually shaped.
- **Test structure is right**: parser tests drive BSON maps through `parseListOperation` directly; formatter tests construct the SDK structs directly. Each layer is tested in isolation.
- **Small and focused**: 225 lines, single purpose, easy to review.

### Concerns

1. **`RANGE` is not valid MDL — DESCRIBE output won't round-trip.** The grammar at `MDLParser.g4:1689-1702` defines `listOperation` as `HEAD | TAIL | FIND | FILTER | SORT | UNION | INTERSECT | SUBTRACT | CONTAINS | EQUALS_OP`. There's no `RANGE`. Re-executing `$Page = RANGE($Orders, 0, 10);` will fail to parse. This violates the "DESCRIBE should output re-executable MDL" guideline. Options:
   - **(a)** Add `RANGE LPAREN VARIABLE COMMA expression COMMA expression RPAREN` to the grammar, wire visitor/AST/executor. Bigger scope but proper fix.
   - **(b)** Emit a placeholder until grammar is added: `-- ListRange(offset=0, limit=10) on $Orders — not yet expressible in MDL`. Safe regression; preserves DESCRIBE for diagnostics.
   - Current PR silently emits invalid MDL — worst of both worlds.

2. **`$currentObject/` prefix in FIND/FILTER output is unusual.** Typical Studio Pro-generated predicates read `Status = 'Active'`, not `$currentObject/Status = 'Active'`. The grammar's own comment on line 1659 shows `FIND($Products, Price > 100)` — bare attribute names. The MDL `expression` rule is probably permissive enough to accept the `$currentObject/...` form, but it's not idiomatic. Worth confirming that a simple `FIND($Orders, Status = 'Active')` wouldn't be cleaner (the list-item scope is implicit).

3. **Silent degradation on malformed `CustomRange`**. If `raw["CustomRange"]` is missing or wrong type, the parser returns `RangeOperation{OffsetExpression: "", LimitExpression: ""}` and the formatter emits `$Page = RANGE($Orders, , );` — broken both as MDL and as display. At minimum, default the empty strings to `"0"` for display, or emit a `-- malformed ListRange` line.

4. **Association path comparison semantics**. The test expects `$currentObject/Order_Customer = $Customer` — but the stored `Association` value is `"MyModule.Order_Customer"` (qualified), and only the last `.`-split part is displayed. That's fine for DESCRIBE readability but loses the module context. If someone re-executes this, does the parser resolve `Order_Customer` against the list element's entity? Worth verifying with a real Studio Pro example — might need to keep the module qualifier.

### Minor
- **No writer side**: this is read-only. `Microflows$FindByExpression` and `Microflows$Find` both map to identical MDL (`FIND($list, expr)`). The writer currently only emits the "expression" variant. That's fine — but means editing a DESCRIBEd microflow and re-creating it will collapse `$Find` → `$FindByExpression`, which Studio Pro may or may not accept as equivalent. Note as follow-up.
- Missing DESCRIBE roundtrip integration test — would have caught (1) immediately.

### Verdict
Blocker: (1) — don't emit invalid MDL from DESCRIBE. Easiest fix is option (b) (placeholder comment) unless you want to take the grammar addition in this PR. (2)–(3) are quick fixes worth doing at the same time. (4) needs a check against a real BSON sample.

hjotha:
Heads-up @yscraft — it looks like PR #212 (merged on 2026-04-18 as commit 2764747) already shipped the same fix for these three operations. Upstream `main` now has:

- `FindByAttributeOperation`, `FilterByAttributeOperation`, `ListRangeOperation` in `sdk/microflows/microflows_actions.go`
- `Microflows$Find` / `Microflows$Filter` / `Microflows$ListRange` cases in `sdk/mpr/parser_microflow.go`
- Matching formatter cases in `mdl/executor/cmd_microflows_format_action.go`

The only visible difference is naming: #212 went with `ListRangeOperation`, this PR with `RangeOperation`. The branch also currently shows `mergeable: dirty` against `main`.

Just flagging in case it's useful — might be worth checking whether the upstream version covers your case and closing this one if so. Thanks for the effort!

ako:
## Status: Superseded — can be closed

The functionality from this PR has already landed on `main` via other commits. All three types, parser cases, and formatter cases are present:

| PR #217 type | Current name | Parser case | Merged? |
|---|---|---|---|
| `FindByAttributeOperation` | `FindByAttributeOperation` | `Microflows$Find` @ `parser_microflow.go:947` | ✅ |
| `FilterByAttributeOperation` | `FilterByAttributeOperation` | `Microflows$Filter` @ `parser_microflow.go:961` | ✅ |
| `RangeOperation` | `ListRangeOperation` (renamed) | `Microflows$ListRange` @ `parser_microflow.go:1012` | ✅ |

`Microflows$FindByExpression` and `Microflows$FilterByExpression` are also handled (bonus coverage beyond this PR's scope).

This PR can be closed as superseded. The related aggregate bug is tracked separately in #241.

PR #210, opened 2026-04-16
feat: Support local file metadata for OData clients

dionesiusap:
# Support local file metadata for OData clients

Closes #206

## Summary

Adds support for local `file://` URLs and relative file paths in `CREATE ODATA CLIENT` `MetadataUrl` parameter, enabling offline development, reproducible testing, and version-pinned contracts.

### 1. Local File Support
- ✅ HTTP(S) URLs: `https://api.example.com/$metadata` (existing)
- ✅ Absolute file:// URLs: `file:///absolute/path/to/metadata.xml`
- ✅ Relative paths: `./metadata/service.xml` or `metadata/service.xml`

### 2. Path Normalization for Studio Pro Compatibility
Relative paths are automatically converted to absolute `file://` URLs in the Mendix model. This ensures Studio Pro's UI can properly detect local file vs HTTP metadata sources.

**Example:**
```sql
-- Input
CREATE ODATA CLIENT Module.Service (
  MetadataUrl: './metadata/service.xml'
);

-- Stored in model
MetadataUrl: 'file:///absolute/path/to/project/metadata/service.xml'
```

### 3. ServiceUrl Validation
`ServiceUrl` must now be a constant reference (prefixed with `@`) to enforce Mendix best practice of externalizing configuration.

**Example:**
```sql
-- ✅ Correct
CREATE CONSTANT Module.Location TYPE String DEFAULT 'https://api.example.com/';
CREATE ODATA CLIENT Module.Service (
  ServiceUrl: '@Module.Location'
);

-- ❌ Rejected with error
CREATE ODATA CLIENT Module.Service (
  ServiceUrl: 'https://api.example.com/'
);
```

### 4. Shared Utility Package
Extracted URL normalization logic into `internal/pathutil` for reuse across components:
- `pathutil.NormalizeURL()` - converts relative paths to absolute file:// URLs
- `pathutil.URIToPath()` - converts file:// URLs to filesystem paths
- `pathutil.PathFromURL()` - extracts path from file:// URLs

Benefits:
- Single source of truth
- Reusable for REST clients with OpenAPI contracts
- Consistent behavior across all components

## Use Cases

- **Offline development** — no network access required
- **Testing and CI/CD** — reproducible builds with metadata snapshots
- **Version control** — commit metadata files alongside code
- **Pre-production** — test against upcoming API changes before deployment
- **Firewall-friendly** — works in locked-down corporate environments

### Path Resolution
[...]

PR #222, opened 2026-04-17
feat: add typed error system for executor

retran:
## Summary

- Introduce `mdl/errors/` package with 6 structured error types (`NotConnectedError`, `NotFoundError`, `AlreadyExistsError`, `UnsupportedError`, `ValidationError`, `BackendError`) plus `ErrExit` sentinel
- Migrate ~1,084 `fmt.Errorf` calls across 79 executor files to typed errors, enabling programmatic error classification via `errors.Is`/`errors.As`
- 19 `fmt.Errorf` calls intentionally preserved (non-standard `%w` wrapping patterns)

## Motivation

The executor currently uses ad-hoc `fmt.Errorf` strings for all errors — no structure, no classification. This makes it impossible for consumers (LSP, TUI, API backends) to programmatically distinguish "not connected" from "entity not found" from "permission denied."

Typed errors enable:
- **Dispatch registry** (future MR): handlers return classified errors that the registry can route
- **LSP integration**: map error types to diagnostic severity and code actions
- **API backend support**: translate executor errors to structured API responses
- **`.mpr` v3 resilience**: `BackendError` wraps I/O failures with operation context, decoupling error semantics from storage format

## Error Types

| Type | Use case | Example |
|------|----------|---------|
| `NotConnectedError` | No project open, or not in write mode | `not connected to a project` |
| `NotFoundError` | Entity, microflow, page, etc. lookup failure | `entity not found: MyModule.MyEntity` |
| `AlreadyExistsError` | Conflict on create | `entity already exists: MyModule.MyEntity` |
| `UnsupportedError` | Unrecognized operation or feature | `unsupported attribute type: Binary` |
| `ValidationError` | Invalid input or state | `invalid entity name` |
| `BackendError` | I/O or storage failure wrapping a cause | `failed to write entity: disk full` |

## Testing

- Full `errors.Is`/`errors.As` contract tests in `mdl/errors/errors_test.go`
- `make build` ✅, `make test` ✅, `make lint` ✅

## Notes

- First commit is `go fmt` on 12 files with pre-existing formatting issues (unrelated to typed errors)
- `executor.ErrExit` is re-exported from the new package for backward compatibility
- No behavior changes — all existing consumers (CLI, TUI, LSP, diaglog) use `.Error()` only and remain fully compatible

### DIFF
--- a/cmd/mxcli/docker/detect.go
+++ b/cmd/mxcli/docker/detect.go
@@ -13,10 +13,10 @@
 // CDN downloads are the primary source for mxbuild and runtime.
 //
 // Resolution priority (all platforms):
-//   1. Explicit path (--mxbuild-path)
-//   2. PATH lookup
-//   3. OS-specific known locations (Studio Pro on Windows)
-//   4. Cached CDN downloads (~/.mxcli/mxbuild/)
+//  1. Explicit path (--mxbuild-path)
+//  2. PATH lookup
+//  3. OS-specific known locations (Studio Pro on Windows)
+//  4. Cached CDN downloads (~/.mxcli/mxbuild/)
 //
 // Path discovery on Windows must NOT hardcode drive letters. Use environment
 // variables (PROGRAMFILES, PROGRAMW6432, SystemDrive) to locate install dirs.
--- a/mdl/ast/ast_entity.go
+++ b/mdl/ast/ast_entity.go
@@ -33,15 +33,15 @@ func (k EntityKind) String() string {
 
 // CreateEntityStmt represents: CREATE [OR MODIFY] PERSISTENT|NON-PERSISTENT ENTITY Module.Name [EXTENDS Parent] (attributes) ...
 type CreateEntityStmt struct {
-	Name             QualifiedName
-	Kind             EntityKind
-	Generalization   *QualifiedName    // Parent entity for inheritance (e.g., System.Image)
-	Attributes       []Attribute
-	Indexes          []Index
-	EventHandlers    []EventHandlerDef // ON BEFORE/AFTER CREATE/COMMIT/DELETE/ROLLBACK CALL ...
-	Position         *Position
-	Documentation    string
-	Comment          string
+	Name           QualifiedName
+	Kind           EntityKind
+	Generalization *QualifiedName // Parent entity for inheritance (e.g., System.Image)
+	Attributes     []Attribute
+	Indexes        []Index
+	EventHandlers  []EventHandlerDef // ON BEFORE/AFTER CREATE/COMMIT/DELETE/ROLLBACK CALL ...
+	Position       *Position
+	Documentation  string
+	Comment        string
 	CreateOrModify bool // true for CREATE OR MODIFY
 }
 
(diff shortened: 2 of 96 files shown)


ako:
## Review

**Overall: well-structured error system with a mechanical but correct migration across 79 files.** CI is green, error messages are preserved, and the test coverage on the error types themselves is thorough. This is solid infrastructure that #224 builds on.

### What I like
- **Clean error taxonomy.** Six types that cover the executor's actual failure modes without over-categorizing. `NotConnectedError` with `WriteMode` flag, `NotFoundError` with `Kind`/`Name` fields, `BackendError` with `Op`+`Unwrap` — the right metadata for each class.
- **Only `BackendError` supports `Unwrap`.** The other five are leaf errors. This is a deliberate design choice that prevents confusing `errors.Is` chains and keeps the error tree flat. Good.
- **`ErrExit` re-exported for backward compat** (`executor.ErrExit = mdlerrors.ErrExit`). Callers that do `errors.Is(err, executor.ErrExit)` keep working. No breaking change.
- **Error message text preserved.** Spot-checked multiple migrations — the `.Error()` output is identical before and after. No consumer sees a change.
- **Classifications are correct.** `NewNotFound` is consistently used for element lookups, `NewBackend` for I/O wrapping, `NewValidation` for input constraints. No misclassifications found in spot checks.
- **Test quality.** `errors.As`/`errors.Is` contract tests, including double-wrapping and nil-cause edge cases on `BackendError`. Exactly the right tests for an error package.

### Concerns

1. **`go fmt` noise (same issue as #210 and #224).** `ast_entity.go`, `ast_microflow.go`, `ast_query.go`, `ast_rest.go`, `docker/detect.go` are whitespace realignment unrelated to typed errors. The PR description acknowledges this ("First commit is `go fmt` on 12 files"), which is better than bundling silently, but it should still be a separate PR or pre-landed commit. Three PRs in a row with the same `go fmt` changes will conflict with each other.

2. **925 → 1085 asymmetry.** 925 `fmt.Errorf` removed, 1085 `mdlerrors.Xxx` added. The delta suggests some error sites were split (one `fmt.Errorf` became multiple typed errors) or new error paths were added. Worth a brief note in the PR description on what accounts for the ~160-line difference — makes the migration auditable.

3. **19 intentionally-preserved `fmt.Errorf` calls.** PR description mentions these are "non-standard `%w` wrapping patterns." Consider adding a `// nolint:mdlerrors` or `// TODO: migrate` comment on each so they're discoverable later and don't look like oversights.

4. **`ValidationError` is the catch-all.** It has no structured fields — just a message string. Every other error type has metadata (`Kind`/`Name`, `WriteMode`, `Op`/`Err`). If a consumer needs to distinguish "invalid entity name" from "invalid XPath expression" from "invalid constant value," they'd need to string-match. Consider adding at least a `Kind` field for future use, even if it's empty for now.

5. **No `Is()` on typed errors.** `NotFoundError{Kind: "entity", Name: "Foo"}` doesn't equal another `NotFoundError{Kind: "entity", Name: "Foo"}` via `errors.Is`. Only `errors.As` works. This is fine for the current use case (classify, not compare) but worth documenting as a design decision since users familiar with sentinel patterns might expect `Is()` to work on instances.

### Minor
- `NewNotFoundMsg` and `NewAlreadyExistsMsg` constructors let callers override the message entirely — means `.Kind`/`.Name` might not match `.Error()`. Acceptable for flexibility but could lead to inconsistency. Trust callers.
- Import alias `mdlerrors` is consistent across all 79 files. Good discipline.

### Verdict
Approve. This should land before #224 (which depends on it). Concerns (1) and (2) are cleanup; (3)–(5) are suggestions for future iterations. No blockers.

PR #224, opened 2026-04-17
feat: dispatch registry, backend interfaces, and MockBackend

retran:
## Summary

Introduces the dispatch registry pattern, domain-specific backend interfaces, and supporting infrastructure for decoupling the executor from direct `.mpr` file access.

> **Stacked on #222** (typed errors). Review after #222 merges — diff will clean up automatically.

### Changes

**Backend interfaces (`mdl/backend/`):**
- 22 domain-specific sub-interfaces covering all Reader/Writer operations
- `FullBackend` composition interface with `Connect`/`Disconnect`/`Commit`
- `MprBackend` in `mdl/backend/mpr/` implementing `FullBackend` via Reader/Writer delegation
- `MockBackend` in `mdl/backend/mock/` with function-field pattern for testing

**Dispatch registry (`mdl/executor/`):**
- `Registry` with `Register()`/`Dispatch()`/`Validate()`/`HandlerCount()`
- `NewRegistry()` builder with explicit wiring (no `init()`)
- 115 handler registrations covering all AST statement types
- Replaced 308-line type-switch in `executor_dispatch.go` with `registry.Dispatch()`
- `ExecContext` struct for future handler migration

**Other improvements:**
- Replaced `goroutine + time.After` timeout with `context.WithTimeout`
- Fixed pre-existing bug: `ShowFeaturesStmt` was missing from dispatch
- Registry completeness test verifying all 115 AST statement types have handlers

### Testing

- `go build ./...` — clean
- `make test` — all pass (10 new registry tests + all existing)
- `make lint` — Go lint passes

### Design decisions

Key decisions (see commit history for ADR):
1. Backend interfaces in dedicated `mdl/backend/` package
2. SecurityBackend as composed sub-interfaces
3. Explicit builder wiring with `context.Context`

Depends on #222.


ako:
## Review

**Overall: the dispatch registry is solid and ready to merge; the backend layer is dead code that needs a proof-of-concept before I'd commit to the interface design.** CI is green, which is a good sign for a 5500-line PR. Stacked on #222 (typed errors, still open).

### What I like
- **Registry design is excellent** (`registry.go`, ~100 LOC). `reflect.Type` keying, panic on duplicate registration (catches wiring bugs at startup), typed `UnsupportedError` on missing handler. The completeness test against 115 known AST types is an impressive safety net — adding a new AST node without a handler will fail the test.
- **No `init()` magic.** `NewRegistry()` calls explicit `registerXxxHandlers()` functions. Easy to grep, easy to understand the wiring order.
- **Typed errors are well-designed** (`mdl/errors/`). Six leaf types with constructor ergonomics, proper `Unwrap` chains, tested via `errors.As`/`errors.Is`.
- **Bug fix caught.** `ShowFeaturesStmt` was missing from the old type-switch — now caught by the registry completeness test.
- **Timeout cleanup.** `goroutine + time.After` → `context.WithTimeout` is a clear improvement.

### Concerns

1. **Backend interfaces are dead code.** `MprBackend`, `MockBackend`, and the 22 sub-interfaces are defined but not wired into the executor. All `cmd_*.go` files still use `e.reader`/`e.writer` directly — zero references to `e.backend`. The changes in `cmd_*.go` are `fmt.Errorf` → `mdlerrors.NewNotFound`/`NewValidation` substitutions, not reader/writer→backend migrations.

   This means:
   - The interface design is untested in integration — only compile-checked via `var _ backend.FullBackend = (*MprBackend)(nil)`
   - The MockBackend can't be validated as useful until handlers actually consume it
   - Risk: the interface shapes may need revision once real migration starts

   **Acceptable as a phase** — but the PR description should be explicit that this is infrastructure-only with zero runtime behavioral change, and the next PR should wire at least a few handlers as proof-of-concept.

2. **22 sub-interfaces may be premature decomposition.** Several are very thin: `NavigationBackend` (3 methods), `WidgetBackend` (2 methods), `ScheduledEventBackend` (2 methods). The rationale (handlers receive only the sub-interface they need via `ExecContext`) is sound but speculative — if handlers commonly need 3-4 sub-interfaces simultaneously, the granularity is overhead without benefit. Until one handler migration is done end-to-end, this is unknowable. I'd be more comfortable with 8-10 coarser interfaces refined from actual usage.

3. **MockBackend is 200+ function fields of boilerplate.** `backend.go` (266 LOC) + 13 domain files, all following `if m.XxxFunc != nil { return m.XxxFunc(...) }; return zero, nil`. This should be code-generated — hand-maintaining it will bitrot as interfaces evolve. At minimum add a `//go:generate` directive or a comment noting the intent.

4. **`go fmt` noise again.** Same issue as PR #210: `ast_entity.go`, `ast_microflow.go`, `ast_query.go`, `ast_rest.go`, `docker/detect.go` are whitespace realignment of struct fields. These should be in a separate commit or dropped if main is already `go fmt`-clean. Inflates the diff significantly for a PR that's already 5500+ lines.

5. **Unchecked type assertions in handler stubs.** Every handler does:
   ```go
   r.Register(&ast.CreateEntityStmt{}, func(e *Executor, stmt ast.Statement) error {
       return e.execCreateEntity(stmt.(*ast.CreateEntityStmt))
   })
   ```
   The registry guarantees correct dispatch, so the assertion won't panic in practice. But a `stmt, ok := ...` defensive check costs nothing and prevents a confusing panic if someone ever bypasses the registry.

6. **`MprBackend.Connect()` always opens in write mode.** No read-only connection mode. If a future "read-only backend" or dry-run mode is planned, `ConnectionBackend` would need revision. Worth a comment on the intended evolution.

7. **`infrastructure.go` is a grab-bag.** Combines `RenameBackend`, `RawUnitBackend`, `MetadataBackend`, `WidgetBackend`, `AgentEditorBackend` in one file. Semantically unrelated — should match the per-interface file organization used elsewhere (`domainmodel.go`, `security.go`, etc.).

### Questions for the author

- Is there a concrete plan for when handlers start using `ExecContext` instead of `*Executor`? Without that migration, the backend interfaces are pure overhead.
- Have you prototyped one handler migration end-to-end (e.g., `execCreateEntity` receiving `ExecContext` with `DomainModelBackend`)? That would validate whether 22 interfaces is the right granularity.
- Re: dependency on #222 — if typed errors are revised, this PR's 79-file `mdlerrors` migration becomes a rebase nightmare. Is there appetite to land #222 first?

### Verdict

The **dispatch registry** and **typed errors** are independently valuable and merge-ready. The **backend interfaces** are reasonable architecture but unproven — I'd want one end-to-end handler migration as proof before committing to the 22-interface design. The `go fmt` noise should be split out.

Approve the registry + typed errors; request that the backend layer either (a) ships with one migrated handler as proof, or (b) is split into a separate follow-up PR that lands after #222 and includes at least one concrete migration.

retran:
> The backend interfaces are reasonable architecture but unproven — I'd want one end-to-end handler migration as proof before committing to the 22-interface design.

Just to clarify - there will be two more pull requests with further migration.

PR #225, opened 2026-04-17
refactor: migrate executor handlers to free functions with ExecContext

retran:
## Summary

Migrates ~450 Executor receiver methods to free functions receiving `*ExecContext`, completing the handler migration phase of the executor refactoring.

### Why free functions?

Receiver methods on `*Executor` have implicit access to all 14 struct fields (reader, writer, cache, etc.), making the dependency surface invisible and preventing backend substitution. Free functions can only access what `*ExecContext` provides — this makes dependencies explicit, enables `MockBackend`-based testing without a `.mpr` file, and is a prerequisite for the next MR where handlers switch from `e.reader.ListModules()` to `ctx.Backend.ListModules()` through domain-specific interfaces. The migration itself is mechanical (no behavioral changes) but necessary to decouple handlers from the concrete `Executor` struct.

### Changes

- Convert all handler methods from `(e *Executor)` receivers to `func name(ctx *ExecContext, ...)` free functions
- Update all 115 handler registrations in `register_stubs.go` to call free functions directly (zero `ctx.executor.` indirection)
- Remove 233 unused Executor wrapper methods; 62 remain (lifecycle API, exported wrappers for external callers, test-called wrappers)
- Add TODO for `ExecContext.Fork()` in `captureDescribe` (parallel describe operations)

## Stacked on

- #224 (dispatch registry + backend interfaces) — stacked on #222 (typed errors)

## Structure

15 incremental commits, each compiles independently:
1. Signature change (`StmtHandler` → `*ExecContext`)
2-10. Domain-by-domain handler migration (connection, module, enum/constant, entity, microflow, page, security, nav/settings/image/workflow, services, mapping/java)
11. Cross-cutting handlers (move/rename, styling, diff, structure, mermaid, contract, context, search, lint, SQL, import, fragments, misc)
12. Core utilities (helpers, format, hierarchy, autocomplete, validate, catalog, query)
13. Cleanup (remove 233 wrappers, fix remaining `e.method()` calls)
14. TODO comment for `ExecContext.Fork()`
15. Review feedback fixes

## Testing

- `make build` — clean
- `make test` — all pass
- `make lint` — Go lint passes
- No behavioral changes — pure mechanical refactoring

### DIFF
--- a/mdl/executor/autocomplete.go
+++ b/mdl/executor/autocomplete.go
@@ -4,8 +4,11 @@
 // Returns qualified names for modules, entities, microflows, pages, etc.
 package executor
 
-// GetModuleNames returns a list of all module names for autocomplete.
-func (e *Executor) GetModuleNames() []string {
+import "context"
+
+// getModuleNames returns a list of all module names for autocomplete.
+func getModuleNames(ctx *ExecContext) []string {
+	e := ctx.executor
 	if e.reader == nil {
 		return nil
 	}
@@ -20,12 +23,13 @@ func (e *Executor) GetModuleNames() []string {
 	return names
 }
 
-// GetMicroflowNames returns qualified microflow names, optionally filtered by module.
-func (e *Executor) GetMicroflowNames(moduleFilter string) []string {
+// getMicroflowNamesAC returns qualified microflow names, optionally filtered by module.
+func getMicroflowNamesAC(ctx *ExecContext, moduleFilter string) []string {
+	e := ctx.executor
 	if e.reader == nil {
 		return nil
 	}
-	h, err := e.getHierarchy()
+	h, err := getHierarchy(ctx)
 	if err != nil {
 		return nil
 	}
@@ -44,12 +48,13 @@ func (e *Executor) GetMicroflowNames(moduleFilter string) []string {
 	return names
 }
 
-// GetEntityNames returns qualified entity names, optionally filtered by module.
-func (e *Executor) GetEntityNames(moduleFilter string) []string {
+// getEntityNamesAC returns qualified entity names, optionally filtered by module.
+func getEntityNamesAC(ctx *ExecContext, moduleFilter string) []string {
+	e := ctx.executor
 	if e.reader == nil {
 		return nil
 	}
-	h, err := e.getHierarchy()
+	h, err := getHierarchy(ctx)
 	if err != nil {
 		return nil
 	}
@@ -70,12 +75,13 @@ func (e *Executor) GetEntityNames(moduleFilter string) []string {
 	return names
 }
 
-// GetPageNames returns qualified page names, optionally filtered by module.
-func (e *Executor) GetPageNames(moduleFilter string) []string {
+// getPageNamesAC returns qualified page names, optionally filtered by module.
+func getPageNamesAC(ctx *ExecContext, moduleFilter string) []string {
+	e := ctx.executor
 	if e.reader == nil {
 		return nil
 	}
-	h, err := e.getHierarchy()
+	h, err := getHierarchy(ctx)
 	if err != nil {
 		return nil
 	}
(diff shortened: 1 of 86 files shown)


ako:
## Review

**Overall: mechanically correct, CI green, but this is the third PR in a stack with zero proof that the end-state architecture works.** The migration itself is clean — 300 functions transformed with the same pattern, 15 commits that each compile independently. No behavioral changes. But no handler actually uses `ctx.Backend` yet.

### What I like
- **No `go fmt` noise.** First PR in this series without unrelated whitespace changes. Thank you.
- **ExecContext is well-designed.** Embeds `context.Context`, carries `Backend`, `Output`, `Format`, `Cache`, `Catalog`, `Fragments`, `Logger`. Clean dependency bag. The `executor` back-reference is honestly documented as temporary.
- **15 incremental commits that each compile.** Domain-by-domain migration makes bisection easy and review tractable despite the 3800-line diff.
- **Remaining 41 Executor methods are justified.** Public API for external callers (`GetModuleNames`, `Search`, `DiffProgram`), lifecycle (`executeInner`, `newExecContext`), formatters (microflow graph traversal). These genuinely need the Executor struct.
- **Handler registrations are direct calls** — `execCreateEntity(ctx, stmt.(*ast.CreateEntityStmt))` with no intermediate indirection through `ctx.executor.execCreateEntity()`. That's the right pattern.

### Concerns

1. **300× `e := ctx.executor` is the entire PR.** Every migrated handler does:
   ```go
   func execCreateEntity(ctx *ExecContext, s *ast.CreateEntityStmt) error {
       e := ctx.executor
       // ... body unchanged, still uses e.reader/e.writer ...
   }
   ```
   This is structurally correct as a phased migration, but means the PR delivers zero decoupling benefit on its own. The value materializes only when the NEXT PR replaces `e.reader.ListModules()` with `ctx.Backend.ListModules()`.

2. **Three PRs deep, zero `ctx.Backend` usage.** #222 (typed errors) → #224 (backend interfaces + registry) → #225 (ExecContext migration). Across all three, `ctx.Backend` is defined but never read by any handler. `MockBackend` exists but is never instantiated in a test. The full investment is ~10,000 lines of infrastructure with no proof-of-concept that the target architecture actually works for one real handler.

   **Strong request:** the next PR in this stack must wire `ctx.Backend` into at least 5-10 handlers from different domains (e.g., one entity handler, one microflow handler, one security handler) AND have a test that uses `MockBackend`. That would validate: (a) the 22 sub-interfaces are the right granularity, (b) the `e := ctx.executor` pattern can actually be removed, (c) `MockBackend` is usable. Without that, the stack remains faith-based.

3. **`StmtHandler` signature changed between PRs.** In #224: `func(e *Executor, stmt ast.Statement) error`. In #225: `func(ctx *ExecContext, stmt ast.Statement) error`. If someone builds on #224's API before #225 lands, their code breaks. This is fine for a stacked PR workflow, but worth noting that #224 and #225 should land together or #224's `StmtHandler` should use `*ExecContext` from the start.

4. **`captureDescribe` concurrency note.** The code has a comment "NOT safe for concurrent use — use captureDescribeParallel instead" and a `TODO: ExecContext.Fork()`. Worth tracking — if `Fork()` deep-copies the context, the `executor` back-reference becomes a shared-state problem. Once `executor` is removed (after Backend migration), this resolves naturally.

### Minor
- `Quiet` field on ExecContext is only used for connection messages. Could be folded into `Output` (use `io.Discard` when quiet). Not urgent.
- Some helper functions like `writeResult`, `writeDescribeJSON`, `getHierarchy` remain as Executor methods but are called by free-function handlers via `ctx.executor.writeResult(...)`. These should eventually migrate to free functions too, or become ExecContext methods.

### Verdict
**Approve.** The migration is mechanical and correct, CI is green, and the incremental commit structure is exemplary. But this PR + #224 + #222 form a three-PR stack of infrastructure prep. The next PR in the stack **must** include concrete Backend usage — without it, the architecture is untested and the 22-interface design is a guess. I won't approve a fourth infrastructure-only PR in this stack.

PR #229, opened 2026-04-17
Context & catalog isolation — ExecContext gains Backend, handlers decoupled from mpr

retran:
## Summary

Third PR in the executor refactoring series (follows #222 typed errors, #224 dispatch registry, #225 handler migration).

- **ExecContext gains domain-specific fields**: `Backend`, `MprPath`, `SqlMgr`, `ThemeRegistry`, `Settings`, `Connected()`/`ConnectedForWrite()` methods
- **Handler decoupling**: ~550 `e.reader`/`e.writer` calls replaced with `ctx.Backend` across 66 files; `e.quiet`/`e.format`/`e.fragments`/`e.mprPath`/`e.output` replaced with `ctx.*` equivalents
- **Free function conversions**: `getEntityNames`, `getMicroflowNames`, `getPageNames` converted from Executor methods to free functions using `ctx.Backend`/`ctx.Cache`; `getThemeRegistry` → `ctx.GetThemeRegistry()` lazy-load method
- **Catalog isolation**: `catalog.NewBuilder` accepts new `CatalogReader` interface instead of concrete `*mpr.Reader`
- **Mock-based tests**: `showEnumerations` handler tested with `MockBackend`, demonstrating the testability gains

~27 `e := ctx.executor` lines remain across ~18 files for legitimate reasons: Executor method calls (`trackModifiedDomainModel`, `Execute`, `ExecuteProgram`), shared state mutations (`catalog`, `cache`, `sqlMgr`), and concrete `*mpr.Reader`/`*mpr.Writer` type requirements (`pageBuilder`, lint, import/export mappings).

### Commits (5)
| # | Description |
|---|-------------|
| 1 | ExecContext gains Backend/MprPath/SqlMgr/ThemeRegistry/Settings |
| 2 | Replace 154 reader/writer nil checks with Connected()/ConnectedForWrite() |
| 3 | Bulk reader/writer → Backend + e.field → ctx.Field (66 files, -264 lines net) |
| 4 | Convert name lookups and theme registry to free functions |
| 5 | CatalogReader interface + mock-based handler tests |

### Verification
- `make build` — clean
- `make test` — all pass
- `make lint` — clean

Depends on: #225 (merged)

### DIFF
--- a/mdl/backend/java.go
+++ b/mdl/backend/java.go
@@ -11,6 +11,7 @@ import (
 // JavaBackend provides Java and JavaScript action operations.
 type JavaBackend interface {
 	ListJavaActions() ([]*mpr.JavaAction, error)
+	ListJavaActionsFull() ([]*javaactions.JavaAction, error)
 	ListJavaScriptActions() ([]*mpr.JavaScriptAction, error)
 	ReadJavaActionByName(qualifiedName string) (*javaactions.JavaAction, error)
 	ReadJavaScriptActionByName(qualifiedName string) (*mpr.JavaScriptAction, error)
--- a/mdl/backend/mock/backend.go
+++ b/mdl/backend/mock/backend.go
@@ -203,6 +203,7 @@ type MockBackend struct {
 
 	// JavaBackend
 	ListJavaActionsFunc            func() ([]*mpr.JavaAction, error)
+	ListJavaActionsFullFunc        func() ([]*javaactions.JavaAction, error)
 	ListJavaScriptActionsFunc      func() ([]*mpr.JavaScriptAction, error)
 	ReadJavaActionByNameFunc       func(qualifiedName string) (*javaactions.JavaAction, error)
 	ReadJavaScriptActionByNameFunc func(qualifiedName string) (*mpr.JavaScriptAction, error)
--- a/mdl/backend/mock/mock_java.go
+++ b/mdl/backend/mock/mock_java.go
@@ -15,6 +15,13 @@ func (m *MockBackend) ListJavaActions() ([]*mpr.JavaAction, error) {
 	return nil, nil
 }
 
+func (m *MockBackend) ListJavaActionsFull() ([]*javaactions.JavaAction, error) {
+	if m.ListJavaActionsFullFunc != nil {
+		return m.ListJavaActionsFullFunc()
+	}
+	return nil, nil
+}
+
 func (m *MockBackend) ListJavaScriptActions() ([]*mpr.JavaScriptAction, error) {
 	if m.ListJavaScriptActionsFunc != nil {
 		return m.ListJavaScriptActionsFunc()
--- a/mdl/backend/mpr/backend.go
+++ b/mdl/backend/mpr/backend.go
@@ -34,6 +34,18 @@ type MprBackend struct {
 	path   string
 }
 
+// Wrap creates an MprBackend that wraps an existing Writer (and its Reader).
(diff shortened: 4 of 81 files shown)


ako:
## Review

**This is the PR the stack needed.** 531 `ctx.Backend` calls across 66 files, mock-based tests proving the pattern works, `CatalogReader` interface decoupling the catalog — all the things I asked for in my #225 review. The architecture is no longer faith-based.

### What I like
- **Backend is now proven in practice.** 531 `ctx.Backend.XXX()` calls replace `e.reader`/`e.writer` across every domain — modules, entities, enumerations, microflows, pages, security, navigation, OData, REST, workflows, agent editor. This isn't a toy migration; it's the real surface area.
- **Mock-based test validates the pattern.** `TestShowEnumerations_Mock` constructs a `MockBackend` with function fields, wires it into an `ExecContext`, calls the handler, asserts the output. `TestShowEnumerations_Mock_FilterByModule` tests the filter path. This is exactly the proof that MockBackend is usable and that handlers can be tested without a `.mpr` file.
- **`Connected()`/`ConnectedForWrite()` replace ~154 nil checks.** Clean semantic methods instead of `if e.reader == nil`. When the backend changes (e.g., read-only mode, remote backend), these methods absorb the difference.
- **`CatalogReader` interface is well-scoped.** Both `*mpr.Reader` and `backend.FullBackend` satisfy it — the catalog builder doesn't know which one it's talking to. This is the right abstraction for the catalog's needs.
- **~27 `e := ctx.executor` remain, documented and justified.** Down from 300 in #225. The remaining sites are lifecycle (`Execute`, `ExecuteProgram`), shared state (`trackModifiedDomainModel`, `catalog`, `cache`), and concrete-type requirements (`pageBuilder`, lint). Honest about what's left.
- **No `go fmt` noise.** Two PRs in a row. Good discipline.
- **5 incremental commits** that map cleanly to the scope: ExecContext fields → connection checks → bulk migration → helper conversions → CatalogReader + mock tests.

### Concerns

1. **CI is still running.** Build-and-test not yet concluded. The diff is large enough that I'd want green before merge. (If it passes, this concern is moot.)

2. **Only one handler has a mock test.** `showEnumerations` proves the pattern, but the real value of MockBackend comes when more handlers get tested this way — especially write-path handlers (create/alter/drop) where testing without a `.mpr` is most valuable. Worth a follow-up ticket: "add mock tests for 5 write-path handlers."

3. **`InvalidateCache()` writes to both `ctx.Cache` and `ctx.executor.cache`.** Dual-state management that could desync if one is nil-checked before the other. Documented as temporary, and it resolves when `executor` is removed, but worth a comment on the invalidation order.

4. **`ConnectedForWrite()` is currently identical to `Connected()`.** The comment says "MprBackend always supports writes." If a read-only backend is introduced, callers of `ConnectedForWrite()` are already doing the right check. But there's no test that exercises the `false` case for write-specific checks. Low risk, just noting.

5. **`CatalogReader` includes `ListRawUnitsByType` and `GetRawUnit`** — these are `mpr`-specific methods (raw BSON units). If a future non-mpr backend doesn't have raw units, it would need to stub these. Consider whether the interface should be narrower (only typed list methods) with raw access as a separate optional interface. Not blocking — the current shape works for both implementations.

### Verdict
**Approve.** This delivers on the promise of #222/#224/#225: handlers are decoupled from `mpr`, MockBackend is validated, and the path to removing the remaining 27 `e := ctx.executor` sites is clear. Wait for CI green, then merge.

PR #232, opened 2026-04-19
feat: add comprehensive mock-based handler tests (189 tests)

retran:
## Summary

Add comprehensive mock-based handler tests for the MDL executor, covering 189 test cases across 33 test files.

**Depends on:** #229 (merged)

### Test infrastructure (`mock_test_helpers_test.go`)
- `newMockCtx` factory with functional options (`withBackend`, `withFormat`, `withHierarchy`, etc.)
- Model factory functions (`mkModule`, `mkEntity`, `mkMicroflow`, `mkPage`, etc.)
- Assertion helpers (`assertNoError`, `assertError`, `assertContainsStr`, `assertValidJSON`)
- `disconnectedBackend()` helper for not-connected guard tests

### Phase 1-8: Handler tests (85 tests)
- Read-path handlers (SHOW/DESCRIBE) across 20+ domains
- Write-path handlers (CREATE/DROP) for 10 operations
- Not-found error paths for describe handlers

### Phase 9: Error paths, not-connected guards, JSON format (104 tests)
- **Error propagation** (49 tests): Every handler group tested with backend returning errors
- **Not-connected guards** (29 tests): All handlers with `Connected()`/`ConnectedForWrite()` guards verified
- **JSON format** (26 tests): All `show`/`list` handlers that use `writeResult` verified with `FormatJSON`

### Production code changes
- Added `Func` fields to `MockBackend` for 8 agent-editor write methods (was hardcoded `return nil`)
- Fixed `ContainerID` parameter semantics in `withContainer` helper and all call sites
- Minor whitespace cleanup in `ast_agenteditor.go`

## Test count

| Category | Tests |
|----------|-------|
| Read-path (SHOW/DESCRIBE) | 73 |
| Write-path (CREATE/DROP) | 12 |
| Backend error propagation | 49 |
| Not-connected guards | 29 |
| JSON format output | 26 |
| **Total** | **189** |


ako:
## Code Review

### Overview

This PR adds a large mock-based test suite for the MDL executor package — 189 tests across 33 new test files. It also fixes the `MockBackend` for 8 agent-editor write methods (from hardcoded `return nil` to injectable `Func` fields), and includes minor alignment/whitespace cleanup in production code.

---

### Positive Aspects

- **Test infrastructure is well-designed.** The `newMockCtx`/functional-options pattern, shared `mkXxx` factories, and `nextID` with `atomic.Int64` are clean and reusable. No global mutable state leaks between tests.
- **Coverage is genuinely useful.** The not-connected guard tests (29) and backend error propagation tests (49) give confidence that every handler fails gracefully rather than panicking or silently succeeding.
- **`assertValidJSON` is well-guarded** — rejects scalars, not just invalid bytes. Good defensive choice.
- **MockBackend fix is correct.** The old hardcoded stubs would have silently hidden write-path failures. The new `Func`-delegating pattern is consistent with the rest of the mock.

---

### Issues & Suggestions

**1. `withContainer` is redundant for direct-module children (minor, low risk)**

The helper comment itself acknowledges: *"For elements whose ContainerID is already a module ID, this call is technically redundant."* Since ~90% of test setups do `withContainer(h, elem.ContainerID, mod.ID)` where `ContainerID == mod.ID`, this is just noise. The comment is adequate, but a helper like `withModuleChild(h, mod.ID)` that skips the call when unnecessary might reduce boilerplate in future tests.

**2. `disconnectedBackend()` is defined in `cmd_notconnected_mock_test.go` but is general-purpose**

If any future test file in this package wants a disconnected backend, it needs to either call this or duplicate it. Consider moving it into `mock_test_helpers_test.go` next to the other helpers.

**3. `idCounter` is package-global atomic — IDs are non-deterministic across test runs**

The comment notes this is "fine for string-contains assertions." That's true for now, but if a test ever asserts on a qualified name that includes the ID, it'll be fragile. Low risk currently, but worth keeping in mind.

**4. No write-path happy-path tests for agent-editor CREATE/DROP**

The PR adds error and not-connected tests for CREATE/DROP agent-editor operations, but there are no happy-path tests that verify a `CreateAgentEditorModelFunc` was actually invoked with the expected struct. Worth adding at least one per agent-editor write op to confirm the `Func` injection actually works end-to-end.

**5. Production code changes are cosmetic alignment only**

The struct literal alignment changes in `cmd_agenteditor_models.go` and `cmd_agenteditor_write.go` are whitespace-only. Fine, but slightly violates the checklist item *"Each commit does one thing."* A separate commit or squash would keep the diff cleaner.

**6. `internal/marketplace/types.go` alignment change is unrelated to tests**

Removing trailing whitespace on one line in `types.go` is unrelated to the PR's stated purpose. Minor, but could be its own fix commit.

---

### Checklist Against Project Standards

| Item | Status |
|---|---|
| New packages have test files | N/A (tests only) |
| MDL examples in `mdl-examples/doctype-tests/` | N/A (unit tests, not MDL scripts) |
| No `time.Sleep` synchronization | ✅ Pass |
| Security: no issues introduced | ✅ Pass |
| Each commit does one thing | ⚠️ Marginal — whitespace changes mixed with test additions |

---

### Verdict

**Approve with minor suggestions.** The infrastructure is solid, the 189 tests add meaningful coverage across all handler domains, and the MockBackend fix is a genuine bug fix. The main gap is missing happy-path write tests for the new agent-editor `Func` fields — those methods now have injectable behavior but no tests that exercise the injection. Everything else is polish.

Issue #257, opened 2026-04-22
Roundtrip and integration-harness stabilization — batch of small fixes


hjotha:
## Summary

While exercising the `describe` / `create` roundtrip on real `.mpr` files and running the integration test harness end-to-end, I hit ~10 distinct small issues spread across the parser, writer, describer, flow-graph builder, and test harness. Fixes are ready as a stack of focused PRs.

Opening this as an umbrella issue (rather than 10 separate ones) because the fixes are small, already implemented, and reviewing them together gives useful context. Happy to split into individual issues if preferred.

## Scope of fixes

Parser / writer (MPR storage layer):

- Parse legacy `NewCaseValue` sequence flows emitted by older `.mpr` files so they roundtrip cleanly.
- Persist `ContentsHash` on v1 unit writes — without this, v1 projects open corrupt after a roundtrip.
- Preserve the original `UnitID` across `DROP` + `CREATE OR REPLACE` so references from other documents don't dangle.
- Accept qualified function calls (e.g. `Module.rule(...)`) in the expression parser.
- Normalise built-in Mendix function case in the expression roundtrip (e.g. `toString` vs `ToString`).
- Gate Mendix 9-only microflow roundtrip keys by project version so Mx 10/11 projects don't get spurious fields.

Describer / flow-graph builder (MDL output + `create microflow`):

- Harden the microflow describe roundtrip (emit the subset of fields we understand, don't drop the rest).
- Support `log` node expressions in the microflow roundtrip.
- Preserve rule-based decision subtype (`RuleSplitCondition`) across microflow roundtrips — previously silently demoted to expression split.
- Emit the describer side of rule splits, free-floating annotations, and a `needMerge` guard so `if` branches that all return don't emit an orphan `SequenceFlow` with an empty `DestinationPointer`.
- Preserve decision/loop captions across nested control flow (`@caption` was lost on nested `if`/`loop`).
- Treat a terminal nested `if` (both branches return) as returning in the flow builder, so the parent doesn't try to wire a merge edge that has no source.

Integration / test harness:

- Prefer the installed `mx` binary in the integration harness (falls back to downloaded mxbuild) to avoid re-downloading on every run.
- Stabilize the MDL test runner and docker flow (timeouts, cleanup, error propagation).
- Stabilize MDL roundtrip and integration flows — specifically: `FindOrCreate` mapping roundtrip, `DROP` + `CREATE OR REPLACE` `UnitID` preservation, default document access roles.
[...]

PR #258, opened 2026-04-22
fix: preserve MPR v1 contents hash and UnitID across DROP+CREATE

hjotha:
## Summary

Two small v1 MPR storage fixes that together prevent Studio Pro from refusing to open projects after MDL roundtrips:

- **`persist contents hashes for v1 unit writes`** — v1 format stored units inline in the DB but didn't refresh the `ContentsHash` row when a write path bypassed the writer's hash recomputation. Studio Pro would see a hash mismatch and treat the file as corrupt.
- **`preserve microflow UnitID across DROP + CREATE OR MODIFY`** — running `DROP MICROFLOW X; CREATE OR MODIFY MICROFLOW X ...` in the same session used to delete the `Unit` row and re-insert with a freshly generated UUID. Studio Pro treats the rewritten `ContainerID`/`UnitID` pair as an unrelated document and refuses to open the resulting `.mpr` ("file does not look like a Mendix Studio Pro project"). Fix caches the dropped microflow's UnitID on the executor and reuses it when a subsequent CREATE OR REPLACE/MODIFY targets the same qualified name, so the delete+insert behaves like an in-place update.

Part of umbrella #257.

## Test plan

- [x] New regression test `TestDropThenCreatePreservesMicroflowUnitID` exercises the full DROP → CREATE OR MODIFY flow through `MockBackend` and asserts the UnitID is preserved and the cache entry consumed.
- [x] `go build ./... && go test ./...` pass on top of current `main`.
- [x] Manual validation: a project that previously fails to open in Studio Pro after the DROP + CREATE pattern now opens cleanly.

### DIFF
--- a/mdl-examples/bug-tests/258-drop-create-microflow-unitid-preservation.mdl
+++ b/mdl-examples/bug-tests/258-drop-create-microflow-unitid-preservation.mdl
@@ -1,0 +1,45 @@
+-- ============================================================================
+-- Bug #258: DROP + CREATE OR MODIFY microflow UnitID preservation
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   A DROP MICROFLOW X followed by CREATE OR MODIFY MICROFLOW X in the same
+--   session produced a *different* UnitID for the new row. Studio Pro treats
+--   Unit rows with a different UnitID as unrelated documents and refuses to
+--   open the project:
+--     ".mpr does not look like a Mendix Studio Pro project file"
+--
+-- After fix:
+--   The executor remembers the dropped UnitID (and ContainerID, AllowedRoles)
+--   and reuses them on the subsequent CREATE so Studio Pro sees an in-place
+--   update. Also exercises v1 ContentsHash persistence — MPR v1 projects need
+--   the hash populated on every write or Studio Pro rejects the file as
+--   corrupt.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/258-drop-create-microflow-unitid-preservation.mdl -p app.mpr
+--   Open the project in Studio Pro — it must load without corruption errors
+--   and the microflow MF_Replaced must be visible under BugTest258.
+-- ============================================================================
+
+create module BugTest258;
+
+create microflow BugTest258.MF_Original
+    returns string
+(
+    set $return = 'original'
+);
+
+-- Drop + recreate in the same session. Before the fix, this produced a
+-- dangling Unit row with a fresh UUID and the project became unopenable in
+-- Studio Pro. After the fix, the new microflow reuses the original UnitID.
+drop microflow BugTest258.MF_Original;
+
+create or modify microflow BugTest258.MF_Original
+    returns string
+(
+    set $return = 'replaced'
+);
+
+-- DESCRIBE must round-trip cleanly (no placeholder UUIDs leaking).
+describe microflow BugTest258.MF_Original;
--- a/mdl/executor/cmd_microflows_create.go
+++ b/mdl/executor/cmd_microflows_create.go
@@ -95,7 +95,9 @@ func execCreateMicroflow(ctx *ExecContext, s *ast.CreateMicroflowStmt) error {
 		if s.Folder == "" && dropped.ContainerID != "" {
 			containerID = dropped.ContainerID
 		}
-		existingAllowedRoles = cloneRoleIDs(dropped.AllowedRoles)
+		// consumeDroppedMicroflow removed the cache entry, so we own this
+		// slice — no need to clone it again.
+		existingAllowedRoles = dropped.AllowedRoles
 		preserveAllowedRoles = true
 	}
 
--- a/mdl/executor/cmd_microflows_create.go
+++ b/mdl/executor/cmd_microflows_create.go
@@ -57,6 +57,8 @@ func execCreateMicroflow(ctx *ExecContext, s *ast.CreateMicroflowStmt) error {
 	// Check if microflow with same name already exists in this module
 	var existingID model.ID
 	var existingContainerID model.ID
+	var existingAllowedRoles []model.ID
+	preserveAllowedRoles := false
 	existingMicroflows, err := ctx.Backend.ListMicroflows()
 	if err != nil {
 		return mdlerrors.NewBackend("check existing microflows", err)
@@ -68,11 +70,14 @@ func execCreateMicroflow(ctx *ExecContext, s *ast.CreateMicroflowStmt) error {
 			}
 			existingID = existing.ID
 			existingContainerID = existing.ContainerID
+			existingAllowedRoles = cloneRoleIDs(existing.AllowedModuleRoles)
+			preserveAllowedRoles = true
 			break
 		}
 	}
 
 	// For CREATE OR REPLACE/MODIFY, reuse the existing ID to preserve references
+	qualifiedName := s.Name.Module + "." + s.Name.Name
 	microflowID := model.ID(types.GenerateID())
 	if existingID != "" {
 		microflowID = existingID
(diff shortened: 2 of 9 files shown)


ako:
## Code Review

### Overview

Two independent v1 MPR correctness fixes:
1. **ContentsHash** — v1 `insertUnit`/`updateUnit`/`WriteUnit` now populate `ContentsHash`, preventing Studio Pro from treating the file as corrupt.
2. **UnitID preservation** — `DROP MICROFLOW X; CREATE OR MODIFY MICROFLOW X` in the same session now reuses the original `UnitID` and `ContainerID` instead of generating fresh UUIDs, so Studio Pro sees an in-place update rather than a delete+insert pair.

Both fixes are correctly motivated, well-tested, and the regression test for the UnitID case is especially thorough.

---

### Issues

**Silent error swallowing in the ContentsHash fallback (both `writer_units.go` and `writer_core.go`)**

```go
_, err := wt.tx.Exec(`UPDATE Unit SET Contents = ?, ContentsHash = ? WHERE UnitID = ?`, ...)
if err != nil {
    _, err = wt.tx.Exec(`UPDATE Unit SET Contents = ? WHERE UnitID = ?`, ...)
}
return err
```

The fallback fires on *any* error from the first `Exec`, not just `"table Unit has no column named ContentsHash"`. If the first update fails for a different reason — UnitID not found, transaction already rolled back, disk full — the fallback silently retries and returns the fallback's result. A write that should fail might silently succeed (without ContentsHash), or a misleading error is returned.

The same pattern appears in `updateUnit` (writer_units.go) and `WriteUnit` (writer_core.go). Minimal fix: check `strings.Contains(err.Error(), "ContentsHash")` before falling back, or detect schema capability once at open time and cache it. At minimum, add a comment acknowledging the broad catch and the risk.

**Performance: double SQL on every v1 write for old-schema projects**

With the current fallback pattern, every `updateUnit` and `WriteUnit` call on an old-schema v1 project executes two SQL statements (first attempt fails, second succeeds). For a script that writes many microflows, this doubles the write query count. Schema detection should ideally happen once at transaction start and the right query chosen statically. Not blocking for this fix, but worth a follow-up.

**Missing `mdl-examples/bug-tests/` script**

The CLAUDE.md checklist requires every bug fix to include an MDL bug-test script in `mdl-examples/bug-tests/` that reproduces the issue for Studio Pro verification. The `scripts/run-mdl-tests.sh` is useful for CI but targets the microflow spec test, not this specific bug. A two-line script like `DROP MICROFLOW X; CREATE OR MODIFY MICROFLOW X ...` in `mdl-examples/bug-tests/` would close the checklist item and serve as a permanent regression artifact.

---

### Minor notes

- `cloneRoleIDs` is called twice for the drop-then-create path: once in `rememberDroppedMicroflow` (when stashing) and once in `execCreateMicroflow` (`cloneRoleIDs(dropped.AllowedRoles)`). The second clone is unnecessary since `consumeDroppedMicroflow` removes the cache entry, so the caller owns the slice. Not a bug, just a spurious allocation.

- `ctx.executor.cache = ctx.Cache` in `rememberDroppedMicroflow` correctly syncs the cache pointer back to the executor so subsequent `newExecContext` calls see the same `droppedMicroflows` map. The test correctly exercises this with a single shared `ctx` rather than going through `executeInner`.

---

### Strengths

- **`TestDropThenCreatePreservesMicroflowUnitID`**: The test reproduces the exact bug scenario — `ListMicroflows` returns nil after `DeleteMicroflow` fires, so `CREATE OR MODIFY` sees no existing microflow and must fall through to the dropped-cache path. The white-box assertion on `ctx.Cache.droppedMicroflows` before and after CREATE precisely documents the invariants the fix relies on.

- **`TestUnitV1_OldSchemaWithoutContentsHashStillWorks`**: Tests the old-schema fallback end-to-end including both `insertUnit` and `updateUnit` in sequence on the same row. Good coverage of the compatibility case.

- **`contentHashBase64` extraction**: The helper eliminates four copies of the same SHA256+base64 block. Clean refactor, correctly shared across `writer_units.go` and `writer_core.go`.

- **`TestCreateOrModifyMicroflowPreservesAllowedRoles`**: Captures a second invariant — `AllowedModuleRoles` must survive a `CREATE OR MODIFY` on an existing microflow. Good addition alongside the drop-then-create test.

---

### Summary

The two fixes are correct and the regression tests are solid. The main concern is the broad-catch fallback pattern in the ContentsHash writes — it should at minimum be narrowed or commented. The missing bug-test script is a checklist item. Everything else is clean.

hjotha:
Thanks @ako — addressed in 9767368:

1. **ContentsHash fallback narrowed** — added `isContentsHashSchemaError` helper that checks the error message for "ContentsHash". Only that specific schema mismatch triggers the retry now; disk-full, invalid UnitID, rolled-back-tx errors propagate unchanged.
2. **Spurious cloneRoleIDs dropped** — `consumeDroppedMicroflow` removes the cache entry, so the caller owns the slice; no second clone needed. Added a comment pointing this out.
3. **Bug-test script added** — `mdl-examples/bug-tests/258-drop-create-microflow-unitid-preservation.mdl` reproduces the DROP+CREATE OR MODIFY scenario end-to-end so Studio Pro verification stays in-tree.

On the **double-SQL performance note**: not addressed in this fixup — the fallback only fires on truly old v1 schemas that are rare in practice, and caching the schema capability at transaction start adds state I'd rather not introduce in a bug-fix PR. Happy to do it as a follow-up if you want me to file an issue.

ako:
## Updated review — fix: preserve MPR v1 contents hash and UnitID across DROP+CREATE

All three concerns from the first review have been addressed. Here's a pass over the changes.

### What was fixed

**Narrow fallback ()** —  and  now only retry on the schema-absent error, not on any error. The  check correctly matches both SQLite message forms ("table Unit has no column named ContentsHash" / "no such column: ContentsHash"). Real errors (disk full, invalid UnitID, rolled-back tx) propagate. ✅

** call removed from CREATE path** — The comment makes the ownership transfer explicit: `consumeDroppedMicroflow` removes the entry from the map so the caller owns the slice. No double-clone. ✅

**Bug reproducer script** — `258-drop-create-microflow-unitid-preservation.mdl` plus `scripts/run-mdl-tests.sh` with explicit path validation. The shell script's `[[ -f "$VAR" ]] ||` guards are the right fix for the silent-empty-sandbox risk. ✅

---

### Minor observation — `insertUnit` fallback is still broad-catch

`updateUnit` now uses `isContentsHashSchemaError`, but `insertUnit` still retries on any error:
```go
_, err := w.reader.db.Exec(`INSERT ... ContentsHash ...`, ...)
if err != nil {
    // Try old schema with Type column
    _, err = w.reader.db.Exec(`INSERT ... Type ...`, ...)
}
```
This is intentional — the INSERT fallback handles two structurally different schemas (ContentsHash+ContentsConflicts vs Type column), not just one missing column, so the pre-existing broad-catch is appropriate. A failed INSERT leaves no data at all rather than stale data, so the risk profile is different from the UPDATE case. This is fine as-is; just noting the asymmetry is by design.

---

### Test coverage

`sdk/mpr/writer_units_test.go` covers all three cases cleanly:
- New v1 schema: hash populated on insert and correct after update
- Old Type-column schema: insert and update succeed without ContentsHash

`TestDropThenCreatePreservesMicroflowUnitID` is a solid regression test: it simulates `listedMicroflows = nil` after deletion (exactly the bug condition), asserts the cache stash, asserts ID reuse, asserts role preservation, and asserts the cache entry is consumed. `TestCreateOrModifyMicroflowPreservesAllowedRoles` covers the "existing via list" path separately.

---

**Verdict: LGTM.** The fixes are targeted, the fallback is correctly narrowed, the test coverage is thorough, and the CLAUDE.md checklist items (bug-test script, narrow error handling, no spurious clone) are all satisfied.

PR #259, opened 2026-04-22
fix: accept qualified function calls in expression parser

hjotha:
## Summary

The MDL expression parser rejected qualified function calls like `Module.rule(...)` even though microflow expressions legitimately reference module-scoped rules and Java actions. The grammar matched unqualified identifiers only, causing parse failures on expressions that Studio Pro accepts.

Fix extends the expression grammar to accept dot-qualified callee names and passes the qualified name through to the visitor, which builds a `QualifiedCall` AST node that the rest of the pipeline already understands.

Part of umbrella #257.

## Test plan

- [x] New visitor test `TestQualifiedCall` covers `Mod.rule(...)` and `Mod.SubMod.rule(...)` forms.
- [x] Grammar regenerated with ANTLR4 (`make grammar`); parser + listener bundle committed.


ako:
## Code Review

### Overview

Minimal, correctly-targeted grammar fix. The one-line grammar change — adding `qualifiedName` as an alternative in `functionCall` — is exactly the right scope. The visitor change is the matching four-line if-else. The tests demonstrate the specific failure mode (describe → check roundtrip breaking on `Module.Rule(Named = $Arg)` IF conditions).

---

### Grammar correctness

ANTLR4's ordered-choice prediction handles the new alternative cleanly:

- `length($s)` → `functionName` wins (listed first; `LENGTH` is in `functionName`)
- `Module.Rule($arg)` → `functionName` fails prediction (`IDENTIFIER DOT` ≠ `IDENTIFIER LPAREN`), `qualifiedName` matches ✓
- `Module.Entity` without `(...)` → `functionCall` prediction fails (no LPAREN following), falls through to `atomicExpression` → `qualifiedName` ✓

The ambiguity between `functionName` and `qualifiedName` for single-token calls (e.g. `find(...)`) is resolved correctly because `functionName` is the first alternative.

---

### Issues

**Missing test for three-segment qualified name**

The PR description mentions `Module.SubMod.rule(...)` is now supported, but neither test exercises three-dot-separated segments. `qualifiedName` allows `identifierOrKeyword (DOT identifierOrKeyword)*`, so `A.B.C($arg)` should work — but a test case confirming it would close the gap claimed by the PR description.

**Missing `mdl-examples/bug-tests/` script**

Per the CLAUDE.md checklist, every bug fix should include an MDL bug-test script that reproduces the issue. A one-statement script like the IF condition from `TestQualifiedCallInIfCondition` would serve as a permanent regression artifact and allow manual Studio Pro validation.

---

### Minor notes

- **Named arguments parse as `BinaryExpr(=, ...)`**: the test comment explains this correctly — `expressionToString` serializes `BinaryExpr(=, IdentifierExpr("String"), VariableExpr("S"))` back to `String = $S`, so the roundtrip is preserved. This is working-as-intended behavior but documenting it as a known semantic overlap (not a true named-argument mechanism) somewhere would prevent future confusion.

- The `buildQualifiedName(qn).String()` call in the visitor is correct and reuses the existing helper. The result (`"ControlCenterCommons.IsNotEmptyString"`) flows cleanly through `expressionToString`'s `FunctionCallExpr` case as `e.Name + "(" + args + ")"`.

---

### Strengths

- The grammar doc comment explains the named-argument parsing semantics and the describe→check symmetry motivation — this is useful context that will survive future grammar edits.
- Tests are structured as end-to-end parse + AST inspection, verifying both that parsing succeeds and that the AST structure is semantically correct (name, argument count, argument type, operator, operand names).
- The interp file is committed alongside the grammar change as expected.

---

### Summary

No blocking issues. Two checklist gaps (three-segment test, bug-test script). The grammar fix itself is correct and conservative.

PR #260, opened 2026-04-22
fix: normalise built-in Mendix function case in expression roundtrip

hjotha:
## Summary

Round-tripping a microflow through `describe` → `create` lower-cased built-in Mendix functions (`toString`, `parseInteger`, etc.). Studio Pro's expression parser is case-sensitive for built-ins, so the round-tripped microflow surfaced CE0117 "Error(s) in expression" on reopen.

Fix normalises the case of known Mendix built-ins when emitting expressions from the describer, so the round-tripped output matches what Studio Pro expects.

Part of umbrella #257.

## Test plan

- [x] Regression test added for `toString`, `parseInteger`, `formatDateTime`, `length`, `substring` round-trip.
- [x] Studio Pro opens the round-tripped project without CE0117.

### DIFF
--- a/mdl/executor/cmd_microflows_helpers.go
+++ b/mdl/executor/cmd_microflows_helpers.go
@@ -68,6 +68,103 @@ func convertASTToMicroflowDataType(dt ast.DataType, entityResolver func(ast.Qual
 	}
 }
 
+// mendixBuiltinFunctions is the canonical spelling of every built-in Mendix
+// expression function. The expression runtime is case-sensitive: it only
+// recognises these names as spelt here (lower-case with camelCase for
+// compound words). Emitting an alternative spelling causes CE0117
+// ("Error(s) in expression.") on Studio Pro validation.
+//
+// Source: https://docs.mendix.com/refguide/expressions/ and the linked
+// function-specific pages (string, math, date arithmetic, parse/format,
+// trim-to-date, list operations, aggregates, type conversions).
+//
+// The map key is the upper-case spelling for case-insensitive lookup; the
+// value is the runtime-accepted canonical spelling. Custom user-defined
+// java actions, sub-microflows, and unknown function names pass through
+// unchanged so user case is preserved.
+var mendixBuiltinFunctions = func() map[string]string {
+	canonical := []string{
+		// List operations
+		"head", "tail", "find", "filter", "sort", "union",
+		"intersect", "subtract", "contains", "equals", "range",
+		// List aggregates
+		"count", "sum", "average", "minimum", "maximum",
+		"allTrue", "anyTrue",
+		// String functions (docs.mendix.com/refguide/string-function-calls)
+		"toUpperCase", "toLowerCase", "trim", "length", "substring",
+		"findLast", "replaceAll", "replaceFirst", "startsWith", "endsWith",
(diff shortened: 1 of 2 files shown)


ako:
## Code Review — fix: normalise built-in Mendix function case in expression roundtrip

### Overview

Two-line fix in the right place ( and ), backed by a comprehensive lookup table and solid test coverage. The root cause is clear: the visitor upper-cases list/aggregate operation names for dispatch (e.g. `FIND`, `COUNT`), and prior to this fix those names were emitted verbatim, producing CE0117.

---

### What works well

**Correct layer** — Fixing in the serializer rather than in the visitor is pragmatic: it avoids touching dispatch logic and passthrough-unknown preserves user case for custom Java actions and sub-microflows.

**Both serialization paths covered** — `mendixFunctionName` applied to `FunctionCallExpr` in both `expressionToString` and `expressionToXPath`. The `flowBuilder.exprToString` method delegates to `expressionToString` (line 84 of `cmd_microflows_builder.go`), so the fix propagates through all callers automatically.

**Map construction** — IIFE initializer is idiomatic Go. Upper-casing the key and storing canonical-case as the value is the right design for a case-insensitive lookup. No allocations at call time.

**Test coverage** — `TestMendixFunctionName_CanonicalSpelling` pins the full documented set; `TestExpressionToString_FunctionCallCaseFixed` verifies the exact `FIND` → `find` path; the nested `parseInteger(substring(..., find(...))))` test mirrors a real Studio Pro pattern and is the most convincing regression guard.

---

### Minor observations

**A few map entries look like documentation errors — harmless but noisy**

- `"rand"` — Mendix docs list `random()` only; `rand` doesn't appear. An unknown caller writing `RAND(...)` would get `rand(...)` emitted, which Studio Pro would still reject. No worse than before, but the entry is misleading.
- `"indexOf"` — Mendix string functions use `find()`/`findLast()`. If `indexOf` isn't valid, the map entry does nothing useful.
- `"toFloat"` — Not in the Mendix expression docs I can find (only `toDecimal` / implicit numeric coercions exist). Same reasoning — harmless if wrong.
- `"reverse"` — Not listed in the Mendix string function reference.

None of these break anything: false positives pass through with the canonical-case spelling they'd have used anyway. But it's worth a quick cross-check against the Mendix 10 refguide before merge so the table serves as accurate documentation, not just a lookup table.

**`"if"` and `"empty"` as function entries** — In Mendix expressions, `if` is `if ... then ... else ...` syntax (handled by `ast.IfThenElseExpr`) and `empty` is a literal keyword (`ast.LiteralEmpty`). Neither would ever arrive as a `FunctionCallExpr.Name`, so these entries are dead code. Harmless, but removing them would make the table more precise.

**XPath `contains` collision** — The map normalizes `CONTAINS` → `contains` (Mendix list operation). In XPath, `contains(string, substring)` is an XPath built-in with the same spelling, so the normalization is correct there too. Just worth noting it was considered.

---

### Verdict

**LGTM with nits.** The core fix is correct and well-tested. The minor items (a few potentially non-existent entries, the dead `if`/`empty` entries) are documentation-quality issues only — they do not affect runtime behavior. A quick refguide pass before merge would make the table trustworthy as a reference, which seems worth it since the comment points to the docs as its source.

PR #263, opened 2026-04-22
fix: preserve decision/loop captions across nested control flow

hjotha:
## Summary

`@caption '...'` annotations on nested decisions (`if`) and loops were being dropped when the describer emitted MDL. The annotation state on the flow builder was overwritten by each child statement as it was processed, so when the outer loop in `buildFlowGraph` consumed the pending annotations for a parent, it picked up the wrong ones.

Fix snapshots each IF's / loop's own pending annotations into local scope before recursing into children, then re-applies them to the outer split/loop after the body is processed. Captions on nested control flow now round-trip.

Part of umbrella #257.

## Test plan

- [x] Regression test exercises `@caption` on nested `if` inside `loop` inside `if`.

## Stack

This PR is the **base** of a 3-PR stack. The other two PRs build on it:

1. **This PR** — nested caption preservation (no dependencies).
2. #265 — rule-based decision subtype preservation (depends on this PR).
3. #266 — describer side of rule splits + free-floating annotations (depends on PR 2).

The stacking is needed because all three touch shared structs in the flow-graph builder; duplicating fields would produce nasty conflicts.

### DIFF
--- a/mdl-examples/bug-tests/263-nested-caption-preservation.mdl
+++ b/mdl-examples/bug-tests/263-nested-caption-preservation.mdl
@@ -1,0 +1,45 @@
+-- ============================================================================
+-- Bug #263: Preserve decision/loop captions across nested control flow
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   `@caption` on an outer IF/LOOP/WHILE was being overwritten by the inner
+--   IF/LOOP's caption because `pendingAnnotations` was shared mutable state
+--   across recursive addStatement calls. Annotations attached to the outer
+--   split ended up bound to the inner split, and the outer split inherited
+--   whatever caption the inner statement carried.
+--
+-- After fix:
+--   addIfStatement / addLoopStatement / addWhileStatement snapshot + clear
+--   `pendingAnnotations` before recursing, then re-apply to their own activity
+--   after it's created. The WHILE case also gained explicit handling in
+--   mergeStatementAnnotations (previously fell through to `default: nil`).
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/263-nested-caption-preservation.mdl -p app.mpr
+--   Open in Studio Pro — each split/loop displays its own caption, not
+--   inherited from a nested statement.
+-- ============================================================================
+
+create module BugTest263;
+
+create microflow BugTest263.MF_NestedCaptions (
+  $S: string
+)
+returns boolean as $ok
+begin
+  declare $ok boolean = false;
+
+  @caption 'String not empty?'
+  if $S != empty then
+    @caption 'Right format?'
+    if isMatch($S, 'x') then
+      return true;
+    else
+      return false;
+    end if;
+  else
+    return false;
+  end if;
+end;
+/
--- a/mdl/executor/cmd_microflows_builder_annotations.go
+++ b/mdl/executor/cmd_microflows_builder_annotations.go
@@ -37,6 +37,8 @@ func getStatementAnnotations(stmt ast.MicroflowStatement) *ast.ActivityAnnotatio
 		return s.Annotations
 	case *ast.LoopStmt:
 		return s.Annotations
+	case *ast.WhileStmt:
+		return s.Annotations
 	case *ast.LogStmt:
 		return s.Annotations
 	case *ast.CallMicroflowStmt:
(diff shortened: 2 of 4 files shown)


ako:
## Code Review — fix: preserve decision/loop captions across nested control flow

### Overview

Clean fix for shared-state contamination in the flow builder. The root cause is clear from reading `addStatement` + `buildFlowGraph`:

1. `addStatement` calls `mergeStatementAnnotations(stmt)` which writes into `fb.pendingAnnotations`.
2. For an IF/LOOP/WHILE, `addStatement` then dispatches into `addIfStatement` / `addLoopStatement` / `addWhileStatement`.
3. Those methods recurse into child statements, which call `addStatement` → `mergeStatementAnnotations` again, overwriting `fb.pendingAnnotations`.
4. When `buildFlowGraph` applies `fb.pendingAnnotations` to the returned `activityID` (lines 62–64 of `cmd_microflows_builder_graph.go`), it reads a child's annotations, not the outer statement's.

The fix is correct: snapshot + clear before recursion, re-apply after the activity is created. `buildFlowGraph`'s post-call check then finds `pendingAnnotations == nil` and no-ops cleanly — no double-apply.

---

### Correctness

**`addIfStatement` fix** — Applies annotations immediately after the `ExclusiveSplit` is appended to `fb.objects` and before any body recursion. `applyAnnotations` searches `fb.objects` for the ID, so the split must already be appended — it is. Correct.

**`addLoopStatement` / `addWhileStatement` fix** — Snapshot before body processing, re-apply after `loop` is created. Identical pattern, same reasoning. Correct.

**No double-apply** — `addIfStatement` clears `fb.pendingAnnotations`; `buildFlowGraph`'s post-call check (`fb.pendingAnnotations != nil`) fails and skips. Clean.

**`ExclusiveSplit` / `InheritanceSplit` in `applyAnnotations`** — Previously only `ActionActivity` was handled; a `@caption` on an IF fell through the type switch silently. Now both split types set `Caption` when the annotation provides one. Correct.

---

### Tests

`TestNestedIfPreservesCaptions` directly reproduces the bug: outer IF wraps inner IF, each with a different `@caption`, verifies both `ExclusiveSplit.Caption` values are correct. This is the right regression test.

`TestIfCaptionWithoutNesting` is a useful sanity baseline.

`TestIfAnnotationStaysWithCorrectSplit` goes further — tests `@annotation` text routing through `annotationFlows` to the correct split. This is the most thorough of the three.

---

### Gaps

**Loop caption fix is untested** — `addLoopStatement` and `addWhileStatement` get the same snapshot/restore treatment but no test covers `@caption` on a `loop` statement. Given the pattern is identical to IF, the correctness is clear from inspection, but a test would close the gap.

**`InheritanceSplit` caption fix is untested** — The new `case *microflows.InheritanceSplit` branch in `applyAnnotations` has no corresponding test case.

**No MDL bug-test script** — CLAUDE.md checklist requires `mdl-examples/bug-tests/` for every bug fix. The unit tests here are thorough, but an MDL-level script (the `@caption 'X' if ... else if ...` pattern) would also serve as documentation of the fixed syntax.

---

### Minor

The test comment "splits are appended in creation order: outer first" is an implementation assumption rather than a contract. It's stable given the current builder logic, but worth flagging as a test fragility if the builder is ever reordered.

---

### Verdict

LGTM with minor gaps. The core fix is correct and well-tested for the primary case. The loop/InheritanceSplit coverage gaps and missing MDL script are worth a follow-up but are not blockers given the identical pattern is tested on the IF side.

hjotha:
Thanks @ako — addressed the LGTM gaps in 94bded8:

1. **Loop caption test added** — `TestLoopCaptionPreserved` and `TestWhileLoopCaptionPreserved` pin the `@caption` behaviour on `LoopStmt` and `WhileStmt`.
2. **InheritanceSplit caption test added** — `TestInheritanceSplitCaptionApplied` exercises the new `applyAnnotations` case directly.
3. **Bug-test MDL script added** at `mdl-examples/bug-tests/263-nested-caption-preservation.mdl`.

While wiring the loop caption test I also caught two related gaps the original fix didn't close:

- `mergeStatementAnnotations` was missing a `*ast.WhileStmt` case, so `@caption` on a `while` silently fell through to `default: nil`. Added.
- `applyAnnotations` had no `*microflows.LoopedActivity` case, so the loop's own caption couldn't land on the activity even when the snapshot/restore delivered it correctly. Added.

Both gaps are covered by the new tests.

Re: the minor note about "splits appended in creation order: outer first" — fair flag. Left the comment as-is for now since reordering the builder would be a separate refactor and the assumption is documented in the test itself rather than hidden in the production code.

PR #264, opened 2026-04-22
fix: support log node expressions and harden microflow describe roundtrip

hjotha:
## Summary

Three microflow round-trip fixes that surfaced together while exercising real `.mpr` files:

- **`parse legacy NewCaseValue sequence flows`** — older `.mpr` files (Mendix 9.x) emit `NewCaseValue` for the true/false labels on exclusive split outgoing sequence flows. The parser dropped them, so the round-tripped describer lost the case value and Studio Pro couldn't reconstruct the split.
- **`harden microflow describe roundtrip`** — the describer bailed or emitted malformed MDL when it hit certain activity shapes it only partially understood. Fix makes the describer skip-and-preserve unknown subfields so the BSON survives write-back unchanged.
- **`support log node expressions in microflow roundtrip`** — the MDL grammar didn't accept the `log` activity's expression/category shape, so round-trips that included a log node failed to re-parse. Grammar + visitor + describer updated end-to-end.

Part of umbrella #257.

## Test plan

- [x] Regression tests cover each of the three cases (legacy case value parse, describer survival, log node round-trip).
- [x] Grammar regenerated; `go build ./... && go test ./...` pass.
- [x] Studio Pro opens the round-tripped project for each fixture.


ako:
## Code Review — fix: support log node expressions and harden microflow describe roundtrip

### Overview

Three fixes bundled together: (1) legacy `NewCaseValue` parser support for Mx 9, (2) `mdlQuote` extended with escape-sequence round-trip, and (3) `LogStmt.Node` promoted from a bare string to a full expression. The PR is larger than the other #257 fixes but everything is coherent and well-tested.

---

### Scope concern

CLAUDE.md requires each PR to cover a single concern. This is three separate fixes that happen to touch overlapping files. The description acknowledges this ("Three microflow round-trip fixes that surfaced together"). Practically speaking the overlap in `cmd_microflows_helpers.go` and the grammar is real, and splitting would require careful ordering. This is a judgment call but worth noting for future PRs in the umbrella.

---

### Fix 1: `NewCaseValue` legacy parser

**Correct.** `parseSequenceFlow` now falls back to `parseCaseValue(raw["NewCaseValue"])` when `CaseValues` is absent. The two new tests (`TestParseSequenceFlow_NewCaseValueEnumerationCase`, `TestParseSequenceFlow_NewCaseValueNoCase`) cover both branch types. The fix is in exactly the right place.

---

### Fix 2: `mdlQuote` escape-sequence round-trip

**Correct direction; one subtle risk.**

The `mdlQuote` upgrade (adding `\n`, `\r`, `\t`, `\\` escaping) and the matching `unquoteString` update create a symmetric encode/decode pair. The test `TestUnquoteString_DecodesEscapes` pins the round-trip. Good.

**`unquoteString` default case** — the final `default` branch correctly writes both the backslash and the following character (`b.WriteByte('\\'); b.WriteByte(s[i])`), preserving unknown sequences. This was the right call.

**Breaking change in `unquoteString`**: existing MDL scripts that contain a `\n`, `\r`, or `\t` sequence would previously have treated these as literal characters (backslash + letter). They now decode as control characters. This is likely intentional and the correct long-term behaviour, but it is a subtle compatibility break worth a note in the CHANGELOG.

**`mdlQuote` applied to `expressionToString`** — `expressionToString` is used both for MDL output (correct to use `mdlQuote`) and for writing Mendix expression strings into BSON (where backslash is not an escape character). For the common case (no backslash or control chars in expression literals) this is identical. The edge case is a Mendix expression string literal containing an actual `\n` or `\\` sourced from BSON directly: the old code emitted a raw newline in the MDL literal (unparseable), while the new code emits `\n` (parseable, but Mendix expression engine sees `\n` as two chars not a newline). Net result is better round-trippability via MDL even if the semantic roundtrip through the expression engine is not bit-identical. Worth a comment in `expressionToString` noting the tension.

---

### Fix 3: `LogStmt.Node` as expression

**Architecture is correct.** Changing `Node string` → `Node Expression` means constant refs (`@Module.Name`), variable refs (`$nodeVar`), or string literals (`'App'`) all work uniformly.

**Grammar change** — `LOG logLevel? (NODE expression)? expression logTemplateParams?` — NODE is now optional and the node is a full expression. Backwards-compatible since `'App'` is a valid expression.

**Visitor ambiguity** — collecting all `IExpressionContext` children and using list position to distinguish node from message is fragile. If the grammar ever adds another optional expression, the index-based dispatch breaks silently. Labelled ANTLR alternatives (`# logNoNode` / `# logWithNode`) or explicit child-accessor methods would be more robust, but this approach is consistent with how other multi-expression statements are handled in the visitor.

**`referencedVars` updated** — `refs = append(refs, exprVarRefs(s.Node)...)` is correct; variables used in the node expression now participate in scope validation.

**`formatAction` log output** — emits `node` as-is from `a.LogNodeName`, which is already a stored expression string. No change needed there; the builder stores the serialised expression. Correct.

---

### Test coverage

Good across all three fixes:
- `sdk/mpr/parser_microflow_test.go`: two `NewCaseValue` tests
- `mdl/visitor/visitor_test.go`: log with string literal node + constant ref node
- `mdl/executor/bugfix_regression_test.go`: builder tests for loop position + log node expression + template literal quoting
- `mdl/executor/roundtrip_microflow_test.go`: full create → describe → re-execute roundtrip for node expression
- `mdl/executor/cmd_microflows_show_helpers_test.go`: multiline escape in `@caption`/`@annotation`
- `mdl/executor/cmd_microflows_format_action_test.go`: multiline message escaping, node expression passthrough

---

### Minor

- `cmd_diff_mdl.go` uses `nodeStr := "'Application'"` as the nil-node default rather than a constant. Should mirror `addLogMessageAction` default for consistency.
- No MDL bug-test script in `mdl-examples/bug-tests/` — CLAUDE.md checklist item.

---

### Verdict

LGTM with the notes above. The three fixes are correct and well-tested. The main items before merge: note the `unquoteString` compatibility break in the CHANGELOG, add a comment in `expressionToString` about the Mendix-expression vs MDL-string dual use of `mdlQuote`, and add an MDL bug-test script.

hjotha:
Thanks @ako — addressed in 494900a:

1. **CHANGELOG entry** — Added to `[Unreleased]`'s Changed section documenting the `mdlQuote`/`unquoteString` escape-sequence behaviour change as a compatibility break for MDL scripts that intentionally embedded raw `\n` / `\t` / `\\` pairs.
2. **Dual-use comment in `expressionToString`** — Added a note explaining the MDL source vs Mendix expression-engine trade-off and why the current behaviour (MDL-parseable output, not bit-identical through the expression engine) is the right call for describe→re-execute.
3. **Default constant unified** — Extracted `defaultLogNodeExpression = "'Application'"` and used it from `addLogMessageAction` (builder), the formatter, and `cmd_diff_mdl`, so the three code paths share a single source of truth instead of duplicating the literal.
4. **Bug-test MDL script** — Added `mdl-examples/bug-tests/264-log-node-expression-roundtrip.mdl` covering literal-node, variable-ref node, and multi-line escape cases.

Re: the scope concern — fair flag. The three fixes did genuinely surface from the same end-to-end exercise and splitting them after the fact would have required grammar regeneration in a specific order; agree on the principle for future PRs in this umbrella.

hjotha:
CI was red on 6ac56fa because `TestRoundtripMicroflow_LogWithNodeExpression` still checked `containsProperty(output, "LOG INFO NODE @...")` in uppercase. The describer now emits lowercase keywords (per the `make lowercase keywords the standard convention in MDL output` change that landed after this branch was written), so the integration test was asserting against the old shape.

Fixed in the same commit — changed the assertion to the lowercase form, matching the other `log info node` check on line 94 of the same file. Pushed the amended commit; CI should be green now.

PR #265, opened 2026-04-22
fix: preserve rule-based decision subtype across microflow roundtrips

hjotha:
## Summary

A decision (`ExclusiveSplit`) whose condition was a rule call was being silently demoted to an expression split on round-trip. The BSON stores a `Microflows$RuleSplitCondition` (not `ExpressionSplitCondition`), and emitting the wrong subtype causes Studio Pro to raise CE0117 "Error(s) in expression".

Fix adds a `RuleQualifiedName` + `ParameterName` field to the split condition struct and wires the parser / writer to round-trip the subtype faithfully. The describer side of this fix is in PR #266 (stacked).

Part of umbrella #257.

## Test plan

- [x] Regression test exercises `if Mod.Rule(param=$value) then ...` round-trip; the resulting BSON contains `Microflows$RuleSplitCondition`, not the expression variant.

## Stack

This PR is the **second** of a 3-PR stack. It depends on #263 (nested caption preservation) because both modify the same flow-builder structs. The describer-side counterpart is in #266, which depends on this PR.

### DIFF
--- a/mdl-examples/bug-tests/263-nested-caption-preservation.mdl
+++ b/mdl-examples/bug-tests/263-nested-caption-preservation.mdl
@@ -1,0 +1,45 @@
+-- ============================================================================
+-- Bug #263: Preserve decision/loop captions across nested control flow
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   `@caption` on an outer IF/LOOP/WHILE was being overwritten by the inner
+--   IF/LOOP's caption because `pendingAnnotations` was shared mutable state
+--   across recursive addStatement calls. Annotations attached to the outer
+--   split ended up bound to the inner split, and the outer split inherited
+--   whatever caption the inner statement carried.
+--
+-- After fix:
+--   addIfStatement / addLoopStatement / addWhileStatement snapshot + clear
+--   `pendingAnnotations` before recursing, then re-apply to their own activity
+--   after it's created. The WHILE case also gained explicit handling in
+--   mergeStatementAnnotations (previously fell through to `default: nil`).
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/263-nested-caption-preservation.mdl -p app.mpr
+--   Open in Studio Pro — each split/loop displays its own caption, not
+--   inherited from a nested statement.
+-- ============================================================================
+
+create module BugTest263;
+
+create microflow BugTest263.MF_NestedCaptions (
+  $S: string
+)
+returns boolean as $ok
+begin
+  declare $ok boolean = false;
+
+  @caption 'String not empty?'
+  if $S != empty then
+    @caption 'Right format?'
+    if isMatch($S, 'x') then
+      return true;
+    else
+      return false;
+    end if;
+  else
+    return false;
+  end if;
+end;
+/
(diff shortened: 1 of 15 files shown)


hjotha:
Re: the AI reviewer's minor note about the unused `RuleID` field — addressed in 1589b68. The field was only populated by the parser and never read anywhere downstream; the rule call is identified by `RuleQualifiedName` in the writer/describer path, so `RuleID` was pure dead weight. Removed from both the struct and the parser assignment.

The second note (performance/caching characteristics of `IsRule`) is a judgement call — `IsRule` does a linear scan over the microflow list in `MprBackend`, which matches other lookup methods in the same interface (`GetMicroflowByQualifiedName`, etc.). Adding caching just for rule-split validation would introduce inconsistency; if callers show up in hot paths later, the whole name-resolution layer can be memoised at once.

ako:
## PR #265 Review — fix: preserve decision/loop subtype across nested control flow

### Overview
Full-stack fix for `RuleSplitCondition` preservation: new `IsRule` backend method → mock stub → MPR reader → builder dispatch → BSON writer. Also retroactively fills the test gaps flagged in the PR #263 review (loop/WHILE/InheritanceSplit caption tests + MDL bug script).

### What's Good
- Clean backend abstraction compliance: `IsRule` is on the interface, mock has a `Func`-field stub, MPR implementation is in `backend/mpr/`, compile-time check via `var _ backend.FullBackend = (*MprBackend)(nil)` (which embeds `MicroflowBackend`) automatically covers `IsRule`
- `tryBuildRuleSplitCondition` / `unwrapParenCall` / `extractNamedArg` are clean, single-purpose helpers
- BSON `Parameter` field extracted as string (`extractString`) rather than binary ID — consistent with how `RuleCall.Microflow` is stored as a qualified name; the prior binary extraction was likely wrong
- Retroactive MDL bug script for #263 and the loop/WHILE/InheritanceSplit caption tests fill the gaps flagged in that review

### Notes / Non-blockers

**`IsRule` performance (3 reads per call)**
`reader_documents.go:IsRule` lists all units by type, builds a module map, then resolves the qualified name — 3 SQLite reads on every invocation. For a microflow with many `IF` branches calling rules this adds up. Consider caching the `Microflows$Rule` set on `MprReader` (lazily, invalidated on write), similar to how entity metadata is cached elsewhere.

**Positional arguments to rule calls silently dropped**
`extractNamedArg` only looks for `NamedArgExpr` nodes. If the caller passes positional args (`myRule(arg1, arg2)`), `paramMappings` stays empty and the rule call is emitted with no parameter mappings — no warning, no error. Consider logging a warning or returning an error when positional args are encountered:
```go
if _, ok := arg.(*ast.NamedArgExpr); !ok {
    return nil, fmt.Errorf("rule call %q: positional arguments are not supported; use named arguments", qualifiedName)
}
```

**Mock default for `IsRuleFunc` returns `(false, nil)`**
This is a safe default (falls back to `ExpressionSplitCondition`) but differs from other stubs that return a `"MockBackend.X not configured"` error when the func is nil. Document the intentional fallback in a comment so future readers don't "fix" it to match the error pattern.

### Verdict
LGTM — the correctness fix and test coverage are solid. The `IsRule` caching and positional-arg warning are worth addressing before the next performance-sensitive workload hits this path, but neither is a blocker for merging.

PR #266, opened 2026-04-22
fix: preserve rule split describer output and free annotations

hjotha:
## Summary

Describer-side counterpart to the rule-split subtype fix, plus two related describer bugs that surfaced from the same end-to-end exercise:

- **Rule split describer output** — with the parser now preserving `RuleSplitCondition`, the describer also needs to emit the qualified rule call (`Mod.Rule(param=$value)`) in the `if ...` header rather than treating every split condition as an expression. Introduces `formatSplitCondition` to dispatch on the condition subtype.
- **`@caption` for splits** — `ExclusiveSplit` / `InheritanceSplit` carry a caption that the describer was dropping. Now emitted as `@caption '...'` before the `if` / `inherit` keyword.
- **Free-floating annotations** — `Annotation` objects with no `AnnotationFlow` (the "sticky note" style) were ignored by the describer. Now collected and emitted as `@annotation '...'` at the top of the microflow body.
- **`needMerge` guard in flow builder** — when both branches of an IF terminate in `return`, the builder must not emit a dangling `SequenceFlow` pointing at an empty `DestinationPointer`. Added guard rails in three emit sites.
- **`writer_core` warning comment** — added a tight warning on `idToBsonBinary` explaining that an empty id here is almost always a bug (produces a random UUID and surfaces as `KeyNotFoundException` in Studio Pro); the fix belongs in the caller.

Part of umbrella #257.

## Test plan

- [x] Regression tests cover rule-split describer output, split captions, free annotations, and needMerge guard.
- [x] Studio Pro opens the round-tripped project without `KeyNotFoundException` or CE0117.

## Stack

This PR is the **top** of a 3-PR stack. It depends on #265 (rule subtype preservation), which in turn depends on #263 (nested caption preservation). Once the two base PRs merge, this one can be rebased directly onto `main`.

### DIFF
--- a/mdl-examples/bug-tests/263-nested-caption-preservation.mdl
+++ b/mdl-examples/bug-tests/263-nested-caption-preservation.mdl
@@ -1,0 +1,45 @@
+-- ============================================================================
+-- Bug #263: Preserve decision/loop captions across nested control flow
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   `@caption` on an outer IF/LOOP/WHILE was being overwritten by the inner
+--   IF/LOOP's caption because `pendingAnnotations` was shared mutable state
+--   across recursive addStatement calls. Annotations attached to the outer
+--   split ended up bound to the inner split, and the outer split inherited
+--   whatever caption the inner statement carried.
+--
+-- After fix:
+--   addIfStatement / addLoopStatement / addWhileStatement snapshot + clear
+--   `pendingAnnotations` before recursing, then re-apply to their own activity
+--   after it's created. The WHILE case also gained explicit handling in
+--   mergeStatementAnnotations (previously fell through to `default: nil`).
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/263-nested-caption-preservation.mdl -p app.mpr
+--   Open in Studio Pro — each split/loop displays its own caption, not
+--   inherited from a nested statement.
+-- ============================================================================
+
+create module BugTest263;
+
+create microflow BugTest263.MF_NestedCaptions (
+  $S: string
+)
+returns boolean as $ok
+begin
+  declare $ok boolean = false;
+
+  @caption 'String not empty?'
+  if $S != empty then
+    @caption 'Right format?'
+    if isMatch($S, 'x') then
+      return true;
+    else
+      return false;
+    end if;
+  else
+    return false;
+  end if;
+end;
+/
--- a/mdl/backend/microflow.go
+++ b/mdl/backend/microflow.go
@@ -26,4 +26,10 @@ type MicroflowBackend interface {
 	UpdateNanoflow(nf *microflows.Nanoflow) error
 	DeleteNanoflow(id model.ID) error
 	MoveNanoflow(nf *microflows.Nanoflow) error
+
+	// IsRule reports whether the given qualified name refers to a rule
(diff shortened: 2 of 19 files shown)


ako:
## PR #266 Review — fix: preserve rule split describer output and free annotations

### Overview
Describer-side counterpart to PRs #263 and #265. Adds `formatSplitCondition` to dispatch rule vs expression split conditions, emits `@caption` for `ExclusiveSplit`/`InheritanceSplit`, collects and emits free-floating annotations, and guards three `SequenceFlow` emit sites against dangling pointers when both branches of an IF terminate.

### What's Good
- `formatSplitCondition` is clean: the `LastIndex` strip for parameter names is correct, the `"true"` fallback is safe and produces valid MDL
- `needMerge` guard (`!thenReturns && needMerge`) at all three emit sites is correct — without it an empty `mergeID` produces a `KeyNotFoundException` in Studio Pro
- `collectFreeAnnotations` correctly identifies unreferenced `Annotation` objects and the `@annotation` prefix approach is simple and doesn't affect activity ordering
- `idToBsonBinary` warning comment is well-placed and actionable

### Issues

**Loop caption describer gap (non-blocker, but inconsistent with builder)**
`applyAnnotations` in this PR gains a `*microflows.LoopedActivity` case so the builder can set loop captions. However `emitObjectAnnotations` in `cmd_microflows_show_helpers.go` only handles `ActionActivity`, `ExclusiveSplit`, and `InheritanceSplit` — it does not emit `@caption` for `LoopedActivity`. The new tests (`TestLoopCaptionPreserved`, `TestWhileLoopCaptionPreserved`) verify the builder side but not the describe → exec roundtrip. A caption applied via `@caption 'foo'` on a LOOP will be silently dropped on the next DESCRIBE.

Suggested addition to `emitObjectAnnotations`:
```go
if loop, ok := obj.(*microflows.LoopedActivity); ok && loop.Caption != "" {
    escapedCaption := strings.ReplaceAll(loop.Caption, "'", "''")
    *lines = append(*lines, indentStr+fmt.Sprintf("@caption '%s'", escapedCaption))
}
```

**Duplicated free-annotation prefix block**
The 5-line free annotation prefix block appears verbatim in both `describeMicroflow` and `renderMicroflowMDL` in `cmd_microflows_show.go`. These two callers share identical logic — worth extracting into a one-liner helper:
```go
func prependFreeAnnotations(lines []string, oc *microflows.MicroflowObjectCollection) []string {
    for _, text := range collectFreeAnnotations(oc) {
        lines = append([]string{fmt.Sprintf("@annotation '%s'", strings.ReplaceAll(text, "'", "''"))}, lines...)
    }
    return lines
}
```
(Or just keep it but it's the kind of duplication that drifts.)

### Minor Notes
- Free annotations are always prepended at the start of the body regardless of canvas position. Since annotations have no execution semantics this is acceptable, but describe output will reorder them relative to their spatial neighbours on re-exec.
- `ParameterName` / `ParameterID` dual extraction for old BSON docs (binary UUID vs string) remains a known pre-existing limitation — not a regression from this PR.

### Verdict
LGTM pending the loop caption describer gap. The `needMerge` guard and `formatSplitCondition` are the correctness-critical pieces and both look right. The loop caption issue is a consistency gap (builder and describer disagree on what `@caption` on a LOOP means) but doesn't break existing functionality.

PR #267, opened 2026-04-22
fix: stabilize MDL roundtrip and integration test harness

hjotha:
## Summary

Four integration-harness and roundtrip stabilization fixes that surfaced from running the full test matrix end-to-end:

- **`test: prefer installed mx in integration harness`** — the harness was re-downloading mxbuild on every run even when a working `mx` was already installed locally. Check for an installed `mx` first, fall back to download. Dramatically reduces test turnaround.
- **`stabilize MDL test runner and docker flow`** — timeouts, cleanup, and error propagation in `scripts/run-mdl-tests.sh` and the docker check flow. Previously a single failing container could leave the harness stuck or swallow the underlying error.
- **`stabilize MDL roundtrip and integration flows`** — three independent roundtrip fixes bundled because they're all exercised by the same integration test fixtures:
  - `FindOrCreate` import-mapping roundtrip (split into `ObjectHandling=Find` + `ObjectHandlingBackup=Create`).
  - Default document access roles (new `autoDocumentRoleDescription` sentinel + `cmd_security_defaults.go` helpers).
  - `DROP` + `CREATE OR REPLACE` microflow UnitID preservation honored across `move` as well.
- **`align role casing and unstick integration roundtrip tests`** — case-insensitive role lookup + fixture updates so the roundtrip tests unblock on projects that use mixed-case role names.

Part of umbrella #257.

## Test plan

- [x] Regression tests added in `bugfix_regression_test.go`, `roundtrip_doctype_test.go`, `parser_import_mapping_test.go`, `writer_export_mapping_test.go`, `writer_import_mapping_test.go`.
- [x] `go build ./... && go test ./...` pass on top of current `main`.
- [x] Integration harness runs end-to-end on the new path.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,7 +151,7 @@ test-integration:
 # Usage: make test-mdl MPR=path/to/app.mpr
 MPR ?= app.mpr
 test-mdl: build
-	$(BUILD_DIR)/$(BINARY_NAME) test mdl-examples/doctype-tests/microflow-spec.test.mdl -p $(MPR)
+	./scripts/run-mdl-tests.sh "$(abspath $(MPR))" "$(abspath $(BUILD_DIR)/$(BINARY_NAME))"
 
 # Lint all code (Go + TypeScript)
 lint: lint-go lint-ts
--- a/cmd/mxcli/docker/build.go
+++ b/cmd/mxcli/docker/build.go
@@ -65,7 +65,7 @@ func Build(opts BuildOptions) error {
 
 	// Step 2: Resolve MxBuild
 	fmt.Fprintln(w, "Resolving MxBuild...")
-	mxbuildPath, err := resolveMxBuild(opts.MxBuildPath)
+	mxbuildPath, err := resolveMxBuild(opts.MxBuildPath, pv.ProductVersion)
 	if err != nil {
 		// Auto-download fallback
 		fmt.Fprintln(w, "  MxBuild not found locally, downloading from CDN...")
@@ -96,7 +96,7 @@ func Build(opts BuildOptions) error {
 	// Step 4: Pre-build check
 	if !opts.SkipCheck {
 		fmt.Fprintln(w, "Checking project for errors...")
-		mxPath, err := ResolveMx(opts.MxBuildPath)
+		mxPath, err := ResolveMxForVersion(opts.MxBuildPath, pv.ProductVersion)
 		if err != nil {
 			fmt.Fprintf(w, "  Skipping check: %v\n", err)
 		} else {
@@ -383,8 +383,10 @@ func Run(opts RunOptions) error {
 }
 
 // findPADDir searches for a PAD output in the output directory tree.
-// It looks for a Dockerfile first, then for docker_compose/Default.yaml as a fallback
-// (MxBuild 11.6.3+ generates docker_compose instead of a Dockerfile).
+// It recognizes:
+//   - classic PAD output with a Dockerfile
+//   - PAD output with docker_compose/Default.yaml
+//   - newer PAD output already extracted at the build root (app/bin/etc/lib)
 func findPADDir(outputDir string) (string, error) {
 	// Check output dir itself
 	if isPADDir(outputDir) {
@@ -405,11 +407,12 @@ func findPADDir(outputDir string) (string, error) {
 		}
 	}
 
-	return "", fmt.Errorf("no PAD output found in %s (looked for Dockerfile or docker_compose/Default.yaml)", outputDir)
+	return "", fmt.Errorf("no PAD output found in %s (looked for Dockerfile, docker_compose/Default.yaml, or extracted app/bin/etc/lib layout)", outputDir)
 }
 
 // isPADDir checks if a directory contains PAD output.
-// Recognizes both Dockerfile-based PAD and docker_compose-based PAD.
+// Recognizes Dockerfile-based PAD, docker_compose-based PAD, and newer
+// already-extracted PAD layout at the build root.
 func isPADDir(dir string) bool {
 	if _, err := os.Stat(filepath.Join(dir, "Dockerfile")); err == nil {
 		return true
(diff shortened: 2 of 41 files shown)


hjotha:
Re: the AI reviewer comment about closing the mpr reader on all error paths in `cmd/mxcli/docker/check.go`:

```go
if reader, err := mpr.Open(opts.ProjectPath); err == nil {
    projectVersion = reader.ProjectVersion().ProductVersion
    reader.Close()
}
```

The reader is only materialized inside the `err == nil` branch, and `reader.Close()` runs unconditionally before the block exits. There's no error path after `mpr.Open` where the reader could leak — if `Open` fails, `reader` is unused. No change needed.

ako:
## PR #267 Review — fix: stabilize MDL roundtrip and integration test harness

### Overview
Four independent stabilization areas: version-aware mx/mxbuild resolution, test harness cleanup, three roundtrip fixes (FindOrCreate, default document roles, cross-module move), and case-insensitive role matching. Solid individual fixes, but the bundling is a concern.

### Blockers / Notable Issues

**PR scope (checklist item: "Each PR is scoped to a single feature or concern")**
This PR bundles four unrelated concerns: mx binary resolution, test harness shell scripts, roundtrip serialization fixes, and security role casing. Each surfaced from the same test run, but they're independently testable and independently reviewable. The CLAUDE.md checklist is explicit here. The roundtrip fixes and the version detection work especially deserve separate attribution and revert surfaces.

**`sdk/mpr` imported directly in `cmd_microflows_builder.go` (backend abstraction violation)**
`lookupMicroflowReturnType` calls `mpr.ParseMicroflowBSON` directly:
```go
import "github.com/mendixlabs/mxcli/sdk/mpr"
...
if mf, err := mpr.ParseMicroflowBSON(rawUnit.Contents, ...); err == nil {
    return mf.ReturnType
}
```
The executor must go through `ctx.Backend.*` — this is explicitly in the checklist. The fix belongs in a new backend method, e.g. `GetMicroflowReturnType(qualifiedName string) (microflows.DataType, error)`, implemented in `mdl/backend/mpr/backend.go`. The existing `GetRawUnitByName` + BSON parsing pattern in the builder is the wrong layer for this.

**`newestVersionedPath` / `versionFromPath` / `parseVersionParts` / `compareVersionParts` duplicated**
These four functions are defined identically in both `cmd/mxcli/docker/download.go` and `mdl/executor/roundtrip_helpers_test.go`. The test file could import the production helpers directly (they're in the same Go module), or both could be extracted to a small internal version-comparison package. Duplicated version comparison logic is exactly the kind of thing that drifts.

### Minor Issues

**`pruneInvalidUserRoles` silent fallback on stale state**
```go
if latest, err := ctx.Backend.GetProjectSecurity(); err == nil {
    ps = latest
} else if ps == nil {
    return err
}
```
If `GetProjectSecurity()` fails AND `ps` is non-nil, the function continues on stale data, silently pruning roles based on outdated state. The safer approach is to return the error unconditionally if the refresh fails.

**`remapDocumentAccessRoles` silently drops unmatched roles on cross-module move**
When `SourceModule.Admin` has no counterpart in the target module, the role is dropped without any output. The user gets no feedback that access was lost. At minimum, log a warning: `"Role %s not found in target module — access removed"`.

### What's Good
- **FindOrCreate roundtrip fix** is clean: writer splits `FindOrCreate → ObjectHandling=Find + ObjectHandlingBackup=Create`; parser reconstructs `FindOrCreate` when it sees `Find + Create` pair. Both directions tested. CE6702 correctly removed from `scriptKnownCEErrors`.
- **`autoDocumentRoleDescription` sentinel** is an elegant way to distinguish auto-provisioned roles from user roles without a separate tracking mechanism.
- **`hasExtractedPADLayout`** heuristic is tight (requires specific files in `bin/start` and `lib/runtime/launcher/`), preventing false positives on arbitrary directories.
- **`assignPattern` multiline fix** (`(?m)^`) is correct — the test `TestGenerateTestRunner_RenamesAllAssignmentsInTestBlock` validates multi-assignment renaming that was broken before.
- **`AddModuleRole` case-insensitive merge** in `writer_security.go` is well-structured; preserving the existing BSON ID when casing changes is the right approach.
- Version detection tests are thorough (exact match, newest fallback, PATH priority, Windows binary names).

### Verdict
The individual fixes are sound and well-tested. The two structural issues — backend abstraction violation in `lookupMicroflowReturnType` and PR scope — need addressing before merge. The duplication in version helpers is a maintenance risk worth fixing in the same pass.

hjotha:
Thanks @ako — addressed the two structural items in 4a1ce9a:

1. **sdk/mpr import in `cmd_microflows_builder.go`** — `lookupMicroflowReturnType` now goes through `fb.backend.ParseMicroflowBSON`. Added `ParseMicroflowBSON(contents, unitID, containerID)` to `MicroflowBackend`; implemented on `MprBackend` (delegates to `mpr.ParseMicroflowBSON`) and `Mock` (`Func`-field stub with a descriptive `MockBackend.ParseMicroflowBSON not configured` error default). The `sdk/mpr` import is out of the builder.

2. **`newestVersionedPath` duplication** — exported the production helper as `docker.NewestVersionedPath` with a full doc comment explaining the ranking. Internal `cmd/mxcli/docker` callers updated. Integration-test harness now imports `docker` and calls the exported helper; all four copies (`newestVersionedPath`, `versionFromPath`, `parseVersionParts`, `compareVersionParts`) removed from `mdl/executor/roundtrip_helpers_test.go`.

On the scope concern — fair call. I'd prefer not to split this one retroactively since the four areas did surface from the same test run and the roundtrip fixes (FindOrCreate, default document roles, cross-module move) are small enough that reverting them piecewise would mean untangling them from the test harness changes that validate them. Open to doing it if you consider it a blocker, though — in that case I'd cut three separate PRs (version resolver + harness; roundtrip fixes; role casing) from this branch.

The two minor issues (`pruneInvalidUserRoles` silent fallback; `remapDocumentAccessRoles` unmatched-role warning) — both fair, will fix in a follow-up on a narrower branch so they don't drag this one further.

`go test ./...` stays green (29 packages, 0 failures).

PR #270, opened 2026-04-23
fix: correct microflow body syntax in bug-test 258

hjotha:
## Summary

The bug-test script that shipped with #258 (`mdl-examples/bug-tests/258-drop-create-microflow-unitid-preservation.mdl`) used the legacy paren-body microflow form:

```mdl
create microflow BugTest258.MF_Original
    returns string
(
    set $return = 'original'
);
```

The current MDL parser only accepts the `begin ... end;` form — the script fails `mxcli check` before it can ever be executed. Script is defanged as a regression artifact until the body is fixed.

This PR rewrites the two microflows with the `begin/end` form so the script parses. The functional payload — `DROP MICROFLOW`, then `CREATE OR MODIFY MICROFLOW` with the same qualified name to exercise `UnitID` preservation, then `DESCRIBE` to verify the roundtrip — is unchanged.

## Test plan

- [x] `mxcli check mdl-examples/bug-tests/258-drop-create-microflow-unitid-preservation.mdl` reports `Syntax OK (5 statements)`.

### DIFF
--- a/mdl-examples/bug-tests/258-drop-create-microflow-unitid-preservation.mdl
+++ b/mdl-examples/bug-tests/258-drop-create-microflow-unitid-preservation.mdl
@@ -24,22 +24,26 @@
 
 create module BugTest258;
 
-create microflow BugTest258.MF_Original
-    returns string
-(
-    set $return = 'original'
-);
+create microflow BugTest258.MF_Original ()
+returns string as $result
+begin
+  declare $result string = 'original';
+  return $result;
+end;
+/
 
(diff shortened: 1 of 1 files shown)


ako:
## PR #270 Review — fix: correct microflow body syntax in bug-test 258

Straightforward fix: two microflows in the #258 bug-test script used the legacy paren-body form that the current parser no longer accepts. Both are rewritten to `begin/end` form. The test payload (DROP + CREATE OR MODIFY to exercise UnitID preservation, then DESCRIBE) is functionally unchanged.

No concerns. LGTM.

PR #271, opened 2026-04-23
fix: DESCRIBE PAGE recurses into ScrollContainer / TabControl children

hjotha:
## Summary

`DESCRIBE PAGE` walks the widget tree by recursing into `Widgets[]` on every container, but two widget types store their children elsewhere:

- **ScrollContainer** holds children under `CenterRegion.Widgets` (Mendix 9+). A fallback to the top-level `Widgets` is kept for older BSON shapes.
- **TabControl** holds children under `TabPages[].Widgets`. The parser now emits a synthetic `Pages$TabPage` intermediate widget per tab, with the tab name and caption preserved, so the DESCRIBE output distinguishes which tab each nested widget belongs to.

Before this fix every widget nested inside those two containers was invisible in DESCRIBE PAGE output, creating silent gaps for any downstream tooling that relied on DESCRIBE to build a widget-ID-to-page map.

## Relationship to #219

This PR is a reimplementation of #219 against the current `main`. Since #219 was opened the `parseRawWidget` / `outputWidgetMDLV3` signatures were refactored (`(e *Executor)` method → free function taking `*ExecContext`), and `tabcontainer` / `tabpage` keywords landed in the MDL grammar. Changes vs #219:

- Updated to the new free-function + `*ExecContext` API.
- Lowercase `tabcontainer` / `tabpage` / `scrollcontainer` output to match the current MDL convention.
- Preserves `Pages$TabPage` as an intermediate node with tab name and caption (#219 collapsed tab structure, losing the grouping — flagged by @ako in the #219 review).
- `tabcontainer` / `tabpage` already exist in the grammar, so the DESCRIBE output round-trips through `mxcli exec`. `scrollcontainer` has no grammar keyword yet; this PR outputs it for DESCRIBE but does not make it re-executable. Grammar addition is a clean follow-up.
- Added unit tests and an `mdl-examples/bug-tests/` reproducer.

Credit to @NikolaSimsic for identifying the original bug in #219 and for the `CenterRegion.Widgets` / `TabPages[].Widgets` field insight. Once this merges, #219 can be closed.

## Test plan

- [x] Unit tests in `mdl/executor/cmd_pages_describe_container_test.go`:
  - `parseRawWidget`: ScrollContainer via CenterRegion, ScrollContainer legacy-Widgets fallback, TabControl with multiple TabPages (names + child widgets preserved per tab).
  - `outputWidgetMDLV3`: scroll container header, tab container + tab page + caption emission.
- [x] Bug-test MDL script at `mdl-examples/bug-tests/219-scrollcontainer-tabcontrol-describe.mdl` round-trips through `mxcli check`.

### DIFF
--- a/mdl-examples/bug-tests/219-scrollcontainer-tabcontrol-describe.mdl
+++ b/mdl-examples/bug-tests/219-scrollcontainer-tabcontrol-describe.mdl
@@ -1,0 +1,44 @@
+-- ============================================================================
+-- Bug #219: DESCRIBE PAGE misses widgets inside ScrollContainer / TabControl
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   `DESCRIBE PAGE` walks the widget tree via Widgets[] on every container.
+--   ScrollContainer stores children under CenterRegion.Widgets and TabControl
+--   under TabPages[].Widgets — neither path was traversed, so every widget
+--   nested inside those two container types was invisible in the describe
+--   output. Any external tool that relied on DESCRIBE output to map widget
+--   IDs to pages (e.g. test impact analysis linkage maps) had silent gaps.
+--
+-- After fix:
+--   parseRawWidget now recurses into CenterRegion.Widgets for ScrollContainer
+--   (with a fallback to Widgets for older BSON) and iterates TabPages[] for
+--   TabControl, emitting a synthetic Pages$TabPage intermediate so DESCRIBE
+--   output shows which tab each widget belongs to.
+--
+-- Usage:
+--   mxcli describe page BugTest219.PageWithTabs -p app.mpr
+--   The describe output must include the inner widget names on each tab;
+--   before the fix these were dropped.
+-- ============================================================================
+
+create module BugTest219;
+
+create page BugTest219.PageWithTabs
+(
+  title: 'Page with tabs',
+  layout: Atlas_Core.Atlas_Default,
+  url: 'bugtest219/tabs',
+  folder: 'Pages'
+)
+{
+  tabcontainer tabs {
+    tabpage tab1 (caption: 'General') {
+      dynamictext generalField (content: 'Hi from tab 1', rendermode: paragraph)
+    }
+    tabpage tab2 (caption: 'Details') {
+      dynamictext detailsField (content: 'Hi from tab 2', rendermode: paragraph)
+    }
+  }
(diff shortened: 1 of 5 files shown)


ako:
## PR #271 Review — fix: DESCRIBE PAGE recurses into ScrollContainer / TabControl children

### Overview
Clean, focused fix for the silent widget-tree gap in DESCRIBE PAGE. `CenterRegion.Widgets` path for ScrollContainer and `TabPages[].Widgets` iteration for TabControl are both correctly added, with the legacy `Widgets` fallback preserved. Synthetic `Pages$TabPage` wrappers preserve tab grouping in the output.

### What's Good
- `inheritedCtx` is correctly defined before the new early-return blocks (lines 27–29 in `parseRawWidget`) — no scoping issue
- `CaptionTemplate` → `extractTextCaption` fallback chain for tab captions is thorough
- `tabpage Name (Caption: '...')` output round-trips correctly through the existing grammar: `TABPAGE` is in `widgetTypeV3`, `CAPTION COLON stringExprV3` is in `widgetPropertyV3`, children go in `widgetBodyV3`
- `scrollcontainer` non-roundtrippability is clearly acknowledged — the grammar keyword omission is the right follow-up, not a blocker here
- 5 focused unit tests cover both parse paths (CenterRegion + legacy fallback, TabControl multi-page) and both output paths (scroll header, tab container structure)

### Minor Notes

**Appearance extraction duplicated between ScrollContainer and TabControl handlers**
Both handlers copy the same `Appearance.Class` / `Appearance.Style` / `DesignProperties` / `extractConditionalSettings` block. Worth extracting into a small helper like `populateWidgetAppearance(w *rawWidget, raw map[string]any)` so the pattern stays in one place. Not a blocker.

**`outputWidgetMDLV3` ScrollContainer and TabControl cases are identical**
Both emit `header`, `appendAppearanceProps`, recurse children — same code, different case labels. Could share a helper. Again minor.

**Tab page appearance properties not extracted**
The synthetic `Pages$TabPage` captures `Name`, `CaptionTemplate`, and children but doesn't extract `Appearance.Class/Style` or `DesignProperties` from the tab page BSON. Real tab pages do carry these. Not visible in the bug case but worth a follow-up.

### Verdict
LGTM. The core fix is correct and well-tested. The minor duplication items are polish, not correctness issues.

PR #289, opened 2026-04-24
refactor: introduce LintReader interface, decouple linter from sdk/mpr

retran:
## Why

The linter and executor packages import `sdk/mpr` directly for `*mpr.Reader`, coupling them to the concrete SQLite implementation. This blocks the WASM build (PR #7 series) where `sdk/mpr` is unavailable, and prevents testing with mock backends. Introducing a `LintReader` interface breaks this dependency using Go's implicit interface satisfaction — zero changes needed in the 10+ linter rule files.

## Summary

- Define `LintReader` interface in `mdl/linter/context.go` with 7 methods matching both `*mpr.Reader` and `MprBackend` signatures
- Replace concrete `*mpr.Reader` dependency with `LintReader` interface throughout linter and executor packages
- Remove `Reader()` methods and `readerProvider` interface from executor — callers now pass `Backend()` directly
- Zero `sdk/mpr` imports remain in `mdl/linter/` and `mdl/executor/` production code

## Test

`make build && make test && make lint-go` — all pass

### DIFF
--- a/cmd/mxcli/cmd_lint.go
+++ b/cmd/mxcli/cmd_lint.go
@@ -120,13 +120,8 @@ Examples:
 		}
 
 		// Create lint context
-		ctx := linter.NewLintContext(cat)
+		ctx := linter.NewLintContext(cat, exec.Backend())
 		ctx.SetExcludedModules(excludeModules)
-
-		// Set reader so rules that inspect raw BSON (MPR004, MPR005) work
-		if reader := exec.Reader(); reader != nil {
-			ctx.SetReader(reader)
-		}
 
 		// Create linter and register rules
 		lint := linter.New(ctx)
--- a/cmd/mxcli/cmd_report.go
+++ b/cmd/mxcli/cmd_report.go
@@ -78,13 +78,8 @@ Examples:
 		}
 
 		// Create lint context
-		ctx := linter.NewLintContext(cat)
+		ctx := linter.NewLintContext(cat, exec.Backend())
 		ctx.SetExcludedModules(excludeModules)
-
-		// Set reader so rules that inspect raw BSON work
-		if reader := exec.Reader(); reader != nil {
-			ctx.SetReader(reader)
-		}
 
 		// Create linter and register all rules
 		lint := linter.New(ctx)
--- a/mdl/backend/mpr/backend.go
+++ b/mdl/backend/mpr/backend.go
@@ -7,6 +7,7 @@ package mprbackend
 
 import (
 	"github.com/mendixlabs/mxcli/mdl/backend"
+	"github.com/mendixlabs/mxcli/mdl/linter"
 	"github.com/mendixlabs/mxcli/mdl/types"
 	"github.com/mendixlabs/mxcli/model"
 	"github.com/mendixlabs/mxcli/sdk/agenteditor"
@@ -20,6 +21,7 @@ import (
 )
 
 var _ backend.FullBackend = (*MprBackend)(nil)
+var _ linter.LintReader = (*MprBackend)(nil)
 
 // MprBackend implements backend.FullBackend by delegating to mpr.Reader
 // and mpr.Writer.
@@ -87,7 +89,7 @@ func (b *MprBackend) MprReader() *mpr.Reader { return b.reader }
 
 func (b *MprBackend) Version() types.MPRVersion { return convertMPRVersion(b.reader.Version()) }
 func (b *MprBackend) ProjectVersion() *types.ProjectVersion {
-	return convertProjectVersion(b.reader.ProjectVersion())
+	return b.reader.ProjectVersion()
 }
 func (b *MprBackend) GetMendixVersion() (string, error) { return b.reader.GetMendixVersion() }
 
--- a/mdl/backend/mpr/convert.go
+++ b/mdl/backend/mpr/convert.go
@@ -5,7 +5,6 @@ package mprbackend
 import (
 	"github.com/mendixlabs/mxcli/mdl/types"
 	"github.com/mendixlabs/mxcli/sdk/mpr"
-	"github.com/mendixlabs/mxcli/sdk/mpr/version"
 )
 
 // ---------------------------------------------------------------------------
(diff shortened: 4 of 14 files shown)


retran:
Thanks for the review — no issues found, agreed on all points.

The `Reader()` backward-compat shim and the consistent interface approach are working as intended.

retran:
Good catch on the `Reader()` return type question. Clarification:

The `Reader()` shim intentionally returns `backend.FullBackend` rather than `*mpr.Reader`. This is safe because:

1. **No external callers use `executor.Reader()` with a concrete `*mpr.Reader` type assertion.** All callers in the codebase access `Reader()` through the linter context interface, which only requires `LintReader` (7 methods). Both `*mpr.Reader` and `backend.FullBackend` satisfy this interface.

2. **The only direct `executor.Reader()` callers** were `cmd_lint.go` and `cmd_report.go` — both updated in this PR to use `Backend()` instead.

3. **`Reader()` on `*mpr.Writer`** (used in `docker/build.go`, examples, and `backend/mpr/backend.go`) is a different method on a different type — unaffected by this change.

The shim exists solely for any out-of-tree callers that might reference `executor.Reader()`. Returning `backend.FullBackend` is a strict superset, so it's backward-compatible at the interface level. If we find a caller doing a concrete type assertion to `*mpr.Reader`, that would already be broken by the backend abstraction and should be migrated.

Re: `types.ProjectVersion.IsAtLeast` — correct, the method was added in this PR to `mdl/types/infrastructure.go` to match the old `sdk/mpr/version` API.

ako (commented):
## Code Review — PR #289

**Overview**: Introduces `LintReader` interface in `mdl/linter/context.go` to decouple linter and executor from `*mpr.Reader`. Clean, minimal diff — the 10+ rule files need no changes because Go's implicit interface satisfaction does the work. The direction is correct.

---

### Good

- Zero `sdk/mpr` imports remain in `mdl/linter/` and `mdl/executor/` production code — stated goal achieved
- `MockBackend` already satisfies `LintReader` (all 7 methods present), so linter rules are immediately testable with mocks
- Moving `IsAtLeast` to `mdl/types.ProjectVersion` is the right direction — `mdl/` code should use `types`, not `sdk/mpr/version`
- Constructor injection (`NewLintContext(cat, reader)`) is cleaner than the old two-step `New` + `SetReader`

---

### Issues

**1. Missing compile-time assertion for `LintReader`**

There's no `var _ linter.LintReader = (*mpr.MprBackend)(nil)` check anywhere. If a future PR removes or renames one of the 7 interface methods in the backend, the breakage is silent until a linter rule calls it at runtime. Add to `mdl/backend/mpr/backend.go`:
```go
var _ linter.LintReader = (*MprBackend)(nil)
```

**2. `Reader()` shim is a silent API break**

`Executor.Reader()` changes return type from `*mpr.Reader` to `backend.FullBackend`. The comment says "backward compatibility" but the signature changed — callers that stored the result as `*mpr.Reader` will fail to compile. This isn't backward compatible; it's a type replacement. The shim gives false confidence. Either drop it (it was already `Deprecated`) or keep the old signature and panic/log if someone calls it (clearer failure mode).

**3. Implicit `nil` reader passed to `NewLintContext`**

Previously, `cmd_lint.go` and `cmd_report.go` had an explicit nil guard before calling `SetReader`:
```go
if reader := exec.Reader(); reader != nil {
    ctx.SetReader(reader)
}
```
Now `exec.Backend()` (which returns `nil` when not connected) is passed unconditionally to `NewLintContext`. Rules already need to nil-check `ctx.Reader()`, so this is safe in practice — but worth a comment on `NewLintContext` that `reader` may be nil and rules must guard accordingly.

**4. `IsAtLeast` now exists on two distinct types**

`mdl/types.ProjectVersion.IsAtLeast` (new) and `sdk/mpr/version.ProjectVersion.IsAtLeast` (existing) are separate implementations on separate types with the same logic. There's no deduplication plan. At minimum, add a `// TODO: remove once all callers use types.ProjectVersion` note to `sdk/mpr/version` to track the migration.

---

### Minor

- `roundtrip_helpers_test.go:599` calls `e.executor.Reader().ProjectVersion()` — this now calls through `backend.FullBackend.ProjectVersion()` returning `*types.ProjectVersion`, which has `IsAtLeast` from this PR. Works, but the indirection via the deprecated `Reader()` shim should be updated to `e.executor.Backend().ProjectVersion()` for clarity.
- `LintReader` is defined in `context.go` but semantically belongs closer to an interface file. Minor — fine as-is given the package size.

---

**Overall**: Approve with the compile-time assertion added (#1) — it's a one-liner and guards against silent future regressions. The `Reader()` shim issue (#2) is worth fixing before merge to avoid confusing external callers.

retran (commented):
Thanks for the review! All items addressed in ab43ea1:

1. **Compile-time assertion** — added `var _ linter.LintReader = (*MprBackend)(nil)` alongside existing `FullBackend` assertion.
2. **Reader() shim** — removed the deprecated method entirely. Only 2 test callers existed; updated both to use `Backend()`.
3. **Nil reader comment** — added doc on `NewLintContext`: "reader may be nil; rules that require backend access must check Reader() != nil."
4. **IsAtLeast dedup** — went with the shared type approach: `sdk/mpr/version.ProjectVersion` is now a type alias for `types.ProjectVersion`. Moved `IsAtLeastFull`, `String`, `IsMPRv2` to `types.ProjectVersion`. Converted `IsSupported`/`SupportsFeature` to package-level functions (can't add methods to type alias). Removed `convertProjectVersion` and its tests since the types are now identical.
5. **Minor** — updated test callers to use `Backend().ProjectVersion()`.

ako (commented):
## Code Review — PR #289 (updated review)

**Overview**: Decouples linter and executor from `*mpr.Reader` by introducing `LintReader` interface. This is a substantially stronger version than the original draft — all three main issues from the earlier review have been addressed.

---

### What's improved since the first review

- **Compile-time assertion added** — `var _ linter.LintReader = (*MprBackend)(nil)` in `mdl/backend/mpr/backend.go` ✓
- **`Reader()` shim removed entirely** — `Executor.Reader()` is gone; callers migrated to `Backend()` directly ✓
- **Nil reader documented** — `NewLintContext` comment notes `reader` may be nil ✓
- **`IsAtLeast` duplication eliminated** — `version.ProjectVersion` is now a type alias for `types.ProjectVersion`; methods live in `mdl/types` once, used everywhere ✓
- **`convertProjectVersion` eliminated** — no longer needed since it's the same type ✓

---

### One remaining issue

**`IsSupported` / `SupportsFeature` method-to-function conversion is an undocumented public API break**

The PR converts:
```go
// before (methods)
pv.IsSupported()
pv.SupportsFeature(feature)

// after (package functions)
version.IsSupported(pv)
version.SupportsFeature(pv, feature)
```

A grep shows no callers outside `version.go` itself, so this is safe for internal code. But these are exported symbols — if anyone imports `sdk/mpr/version` externally, their code breaks silently. Worth a CHANGELOG entry under **Changed** (analogous to the `mdl/types` extract entry in v0.7.0).

---

### Minor observations

- `MockBackend` doesn't have `var _ linter.LintReader = (*MockBackend)(nil)` — not a safety gap since `MockBackend` already has `var _ backend.FullBackend` and `FullBackend` is a superset of `LintReader`. But adding it would make the mock's linting capability explicit and keep the two assertions in sync.
- `mdl/linter` now imports `sdk/microflows`, `sdk/pages`, `sdk/security` for the interface method signatures. This is unavoidable and correct — no new cycles introduced.

---

**Overall**: Approve. The type-alias approach for `ProjectVersion` is particularly clean. The `IsSupported`/`SupportsFeature` change is the only thing worth noting before merge, and only as a changelog entry since there are no internal callers to update.

PR #297, opened 2026-04-24
refactor: add type-safe BSON helpers in writer_security

retran:
## Summary

- Replace ~20 raw `interface{}` type assertions in `sdk/mpr/writer_security.go` with type-safe `bsonString()`/`bsonBool()` helper functions
- Helpers log warnings with field name context instead of silently returning zero values
- Add comprehensive tests for unexpected types and nil handling

> **Depends on #289** (Reader Migration). Retarget after #289 merges — diff currently shows combined changes.

## Why?

**Panic risk in production.** `writer_security.go` uses ~20 bare type assertions (`val.(string)`, `val.(bool)`) against BSON data read from `.mpr` files. If a Mendix version introduces an unexpected type in a security field — or if a corrupted project is opened — any of these assertions will panic, crashing the CLI with no diagnostic context.

**Silent failures are worse than crashes.** The alternative (comma-ok assertions scattered across every call site) would silently return zero values, making security merge bugs nearly impossible to diagnose. The `bsonString`/`bsonBool` helpers strike the right balance: return a safe zero value *and* log a warning with the field name, so unexpected data is surfaced without crashing.

**Dependency coupling.** The linter and executor packages previously imported `sdk/mpr.Reader` directly, creating a concrete dependency on the MPR implementation. The `LintReader` interface (from the #289 base) breaks this coupling, which is prerequisite work for the WASM extraction milestone where `sdk/mpr` won't be available at all.

## Details

`writer_security.go` had ~20 bare `val.(string)` / `val.(bool)` assertions that would panic on unexpected BSON data. This PR:

1. Introduces `bsonString(val, field)` and `bsonBool(val, field)` helpers using `any` type parameter
2. Each helper returns a zero value + logs a `warning:` message with `%q` field name on type mismatch or nil
3. Replaces all raw assertions in `mergeAccessRule()` and related functions
4. Adds `TestBsonHelpers_UnexpectedTypes_NoPanic` — verifies no panic on wrong types, validates return values
5. Adds `TestMergeAccessRule_UnexpectedTypes_NoPanic` — integration test with malformed BSON input
6. Tests capture and assert log output for all 6 warning paths (string×3 + bool×3: wrong type, nil, correct)

## Stack

- **Depends on #289** (PR 2: Reader Migration) — please merge #289 first, then retarget this PR
- Part of the mxcli extraction/cleanup PR sequence

### DIFF
--- a/mdl/bsonutil/bsonutil.go
+++ b/mdl/bsonutil/bsonutil.go
@@ -1,12 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
-// Package bsonutil provides BSON-aware ID conversion utilities for model elements.
+// Package bsonutil provides BSON-aware ID conversion and type-extraction utilities
+// for model elements.
 // It depends on mdl/types (WASM-safe) and the BSON driver (also WASM-safe),
 // but does NOT depend on sdk/mpr (which pulls in SQLite/CGO).
 package bsonutil
 
 import (
 	"fmt"
+	"log"
 
 	"github.com/mendixlabs/mxcli/mdl/types"
 	"go.mongodb.org/mongo-driver/bson/primitive"
@@ -55,4 +57,32 @@ func NewIDBsonBinaryErr() (primitive.Binary, error) {
 		return primitive.Binary{}, fmt.Errorf("generating ID: %w", err)
 	}
 	return IDToBsonBinaryErr(id)
+}
+
+// String extracts a string from an interface value.
+// Returns the zero value and logs a diagnostic warning on type mismatch.
+func String(v any, field string) string {
+	if s, ok := v.(string); ok {
+		return s
+	}
+	if v == nil {
+		log.Printf("warning: bsonutil: expected string for %q, got <nil>", field)
+	} else {
+		log.Printf("warning: bsonutil: expected string for %q, got %T", field, v)
+	}
+	return ""
+}
+
+// Bool extracts a bool from an interface value.
+// Returns the zero value and logs a diagnostic warning on type mismatch.
+func Bool(v any, field string) bool {
+	if b, ok := v.(bool); ok {
+		return b
+	}
+	if v == nil {
+		log.Printf("warning: bsonutil: expected bool for %q, got <nil>", field)
+	} else {
+		log.Printf("warning: bsonutil: expected bool for %q, got %T", field, v)
+	}
+	return false
 }
--- a/mdl/bsonutil/bsonutil_test.go
+++ b/mdl/bsonutil/bsonutil_test.go
@@ -3,6 +3,9 @@
 package bsonutil
 
 import (
+	"bytes"
+	"log"
+	"strings"
 	"testing"
 
 	"github.com/mendixlabs/mxcli/mdl/types"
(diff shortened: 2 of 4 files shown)


ako (commented):
## Code Review — PR #297

**Overview**: Two independent changes stacked together — the `LintReader` work from #289 (base) and the new `bsonString`/`bsonBool` helpers in `writer_security.go`. The note to retarget after #289 merges is correct; reviewing only the net-new work here.

---

### The actual new work: `bsonString` / `bsonBool` helpers

**Good:**
- Correctly eliminates all ~20 panicking bare assertions in `writer_security.go` — the only writer file that had them (the remaining assertions in `writer_domainmodel.go` are already comma-ok form, which is safe)
- Helper pattern is clean and consistent: zero value + diagnostic log with field name
- `TestBsonHelpers_UnexpectedTypes_NoPanic` verifies both the no-panic contract AND the 6 warning messages — thorough
- `TestMergeAccessRule_UnexpectedTypes_NoPanic` gives an integration-level smoke test with malformed BSON

**One issue:**

**`bsonString`/`bsonBool` are private to `sdk/mpr` — but `mdl/bsonutil` already exists**

`sdk/mpr` already imports `mdl/bsonutil` (for `IDToBsonBinary`). The new helpers are a natural fit there — they're BSON utilities that don't depend on SQLite/CGO, which is exactly `bsonutil`'s design goal. If `writer_domainmodel.go` and other writers get the same treatment in future PRs, every writer will redeclare private copies.

Not blocking, but worth considering now while the helpers are small:
```go
// mdl/bsonutil/bsonutil.go
func String(v any, field string) string { ... }
func Bool(v any, field string) bool { ... }
```

**Minor observations:**

- **`log.Printf` is correct here** — `sdk/mpr` has no `diaglog` usage; stdlib logger is the right choice for a low-level package.
- **Test captures global `log` output** via `log.SetOutput` — this is not parallel-safe. The test implicitly relies on sequential execution (Go's default). A `// not parallel-safe: redirects global log output` comment would make the intent clear.
- **`TestMergeAccessRule_UnexpectedTypes_NoPanic`** verifies the result is non-nil but doesn't assert the actual field values (e.g. `AllowCreate` should still be `true` after merging with a wrong-typed existing value). Could be more thorough, but "no panic" is the stated goal, so fine as-is.

---

### Noise from #289 base

The diff includes all of #289's changes (LintReader interface, ProjectVersion alias, Reader() removal). After #289 merges and this is retargeted, the actual diff will be ~40 lines: two helper functions + ~20 call-site replacements + two test functions. Clean and reviewable on its own.

---

**Overall**: Approve once retargeted onto main after #289 merges. The helpers solve a real panic risk cleanly. Consider moving them to `mdl/bsonutil` if other writers follow the same treatment.

retran:
Thanks for the thorough review! All points addressed in 62b5a16:

**Moved `bsonString`/`bsonBool` to `mdl/bsonutil`** — exported as `bsonutil.String()` / `bsonutil.Bool()`. Log prefix changed from `writer_security:` to `bsonutil:` since they're now generic. Unit tests moved to `bsonutil_test.go`; integration test (`TestMergeAccessRule_UnexpectedTypes_NoPanic`) stays in `writer_security_test.go`.

**Added `// Not parallel-safe: redirects global log output.` comments** on both tests that use `log.SetOutput`.

Re: `TestMergeAccessRule` field value assertions — agreed, "no panic" is the stated goal, keeping as-is.

Rebased on latest main (which now includes merged #289), so the diff is clean — just the 2 net-new commits.

PR #298, opened 2026-04-24
Fix DESCRIBE PAGE / ALTER PAGE column name mismatch for DataGrid2 columns (#116)

### LINKED ISSUE #116: DESCRIBE PAGE widget names don't match ALTER PAGE identifiers
## Description

\`DESCRIBE PAGE\` outputs column names like \`COLUMN Description (Attribute: Description)\`, but these names can't be used in \`ALTER PAGE\` operations to target specific widgets. The DESCRIBE output suggests these are valid widget identifiers, but ALTER PAGE can't find them.

## Expected Behavior

DESCRIBE output should show names that can be used directly in ALTER PAGE, or show the internal widget ID alongside the display name.

## Workaround

Use \`CREATE OR MODIFY PAGE\` to recreate the page instead of \`ALTER PAGE\` for targeted changes.

## Impact

Low — misleading output, but workaround exists.

yscraft:
## Summary

Fixes #116.

`DESCRIBE PAGE` and `ALTER PAGE` derived DataGrid2 column names independently with two bugs that caused mismatches:

- **Wrong TextTemplate path (H1):** `deriveColumnNameBson` read `TextTemplate → Items` directly, skipping the intermediate `Template` document. `DESCRIBE` correctly traverses `TextTemplate → Template → Items`. Result: caption-only columns always fell back to `col1`, `col2`, … in ALTER regardless of what DESCRIBE showed.
- **Sanitization divergence (H2):** `sanitizeColumnName` didn't trim leading/trailing underscores and didn't fall through to `col{N}` for all-special-char captions, unlike `deriveColumnName` on the DESCRIBE side.

## Changes

**`mdl/executor/cmd_alter_page.go`**
- `deriveColumnNameBson` — traverse `TextTemplate → Template → Items` (nil-fallback for edge cases); gate `sanitizeColumnName` result so empty string falls through to `col{N}`
- `sanitizeColumnName` — rewritten to match `deriveColumnName` exactly: `strings.Map` + `strings.TrimFunc`, returns `""` when result is all underscores

**`mdl/executor/alter_page_test.go`**
- 5 regression tests: attribute binding, caption fallback (the broken path), all-special-char caption → `col{N}`, index fallback, sanitization trimming

## Test plan

- [x] `go test ./mdl/executor/... -run TestDeriveColumnNameBson|TestSanitizeColumnName` — all 5 pass
- [x] `go test ./mdl/executor/...` — full suite passes, no regressions

### DIFF
--- a/mdl-examples/bug-tests/116-datagrid2-column-name-mismatch.mdl
+++ b/mdl-examples/bug-tests/116-datagrid2-column-name-mismatch.mdl
@@ -1,0 +1,72 @@
+-- ============================================================================
+-- Bug test: issue #116 — DESCRIBE PAGE / ALTER PAGE DataGrid2 column name mismatch
+-- ============================================================================
+--
+-- Verifies that ALTER PAGE can address DataGrid2 columns using the names that
+-- DESCRIBE PAGE produces — for both attribute-bound columns and caption-only
+-- columns (the previously broken case).
+--
+-- Affected functions (fixed in PR #298):
+--   deriveColumnNameBson  — now traverses TextTemplate → Template → Items
+--   sanitizeColumnName    — now matches deriveColumnName exactly
+--
+-- To verify in Studio Pro:
+--   1. Run this script against a project.
+--   2. DESCRIBE PAGE Bug116.OrderList — column names in output should match
+--      what ALTER PAGE uses below (Name, Order_Status, col3).
+--   3. Confirm no errors.
+-- ============================================================================
+
+-- ############################################################################
+-- PREREQUISITES
+-- ############################################################################
+
+CREATE MODULE Bug116;
+
+CREATE PERSISTENT ENTITY Bug116.Order (
+  Name: String(200),
+  Status: String(50)
+);
+
+-- ############################################################################
+-- PAGE WITH THREE COLUMN TYPES
+--
+-- colAttr     — attribute-bound: DESCRIBE yields "Name"
+-- colCaption  — caption-only (the previously broken case): DESCRIBE yields "Order_Status"
+-- colNoName   — no attribute, special-char caption only: DESCRIBE yields "col3"
+-- ############################################################################
+
+CREATE PAGE Bug116.OrderList (
+  Title: 'Order List',
+  Layout: Atlas_Core.Atlas_Default
+) {
+  DATAGRID dgOrders (DataSource: DATABASE Bug116.Order) {
+    COLUMN colAttr    (Attribute: Name,   Caption: 'Name')
+    COLUMN colCaption (                   Caption: 'Order Status')
+    COLUMN colNoName  (                   Caption: '---')
+  }
+};
+
+-- ############################################################################
+-- ALTER PAGE USING DESCRIBE-DERIVED COLUMN NAMES
+--
+-- H1 fix: colCaption uses TextTemplate → Template → Items path;
+--         previously always fell back to col2 because Template was skipped.
+-- H2 fix: sanitizeColumnName now trims underscores and falls through to col{N}
+--         for all-special-char captions like '---'.
+-- ############################################################################
+
+-- Attribute-bound column — always worked; regression guard
+ALTER PAGE Bug116.OrderList {
+  SET WIDGET dgOrders.Name (Visible: true)
+};
+
+-- Caption-only column — previously broken (derived col2 instead of Order_Status)
+ALTER PAGE Bug116.OrderList {
+  SET WIDGET dgOrders.Order_Status (Visible: true)
+};
+
+-- All-special-char caption → falls back to col3 on both sides
+ALTER PAGE Bug116.OrderList {
+  SET WIDGET dgOrders.col3 (Visible: true)
+};
--- a/mdl/executor/alter_page_test.go
+++ b/mdl/executor/alter_page_test.go
@@ -700,3 +700,138 @@ func TestExtractWidgetScopeFromBSON_Nil(t *testing.T) {
 		t.Error("Expected empty scope for nil input")
 	}
 }
+
+// ============================================================================
+// deriveColumnNameBson regression tests (issue #116)
+// ============================================================================
+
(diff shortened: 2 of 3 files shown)


yscraft:
Hey @ako, this PR fixes issue #116 — DataGrid2 column name mismatch between `DESCRIBE PAGE` and `ALTER PAGE`. Could you take a look when you get a chance?

ako (commented):
## Code Review — PR #298

**Overview**: Fixes a DESCRIBE/ALTER mismatch for DataGrid2 column names. Two independent bugs correctly identified and fixed. The unit tests are solid. One design concern before merging.

---

### What's good

- Root cause analysis is correct — the PR description accurately explains both bugs (H1: wrong BSON traversal path, H2: sanitization divergence)
- `sanitizeColumnName` rewrite is a direct copy of the DESCRIBE-side logic — guaranteed to match
- The empty-string gate (`if name := sanitizeColumnName(caption); name != ""`) correctly enables the `col{N}` fallback for all-special-char captions
- 5 unit tests cover attribute binding, caption path, all-special-char fallback, index fallback, and sanitization trimming

---

### One concern: the `template = tmpl` fallback may reintroduce a divergence

The H1 fix adds a defensive fallback:
```go
template := dGetDoc(tmpl, "Template")
if template == nil {
    template = tmpl // fallback: some versions store Items directly
}
```

The DESCRIBE side (`extractTextContent` in `cmd_pages_describe_output.go`) does **not** have this fallback:
```go
template, ok := content["Template"].(map[string]any)
if !ok {
    return ""  // no Template → no caption
}
```

If `Template` is nil, DESCRIBE returns `""` and falls through to `col{N}`. ALTER with the fallback would instead read `Items` from `tmpl` directly and derive a name — producing a different result than DESCRIBE.

The stated goal is for ALTER and DESCRIBE to agree. The safest fix is to match DESCRIBE exactly and drop the fallback:
```go
template := dGetDoc(tmpl, "Template")
if template == nil {
    break // no Template → no caption, match DESCRIBE behaviour
}
```

There's also no test for the fallback path, which suggests there's no known Mendix version that actually uses it.

---

### Missing: bug-test MDL script

Per the PR checklist in `CLAUDE.md`:
> Every bug fix should include an MDL test script in `mdl-examples/bug-tests/` that reproduces the issue, so the fix can be verified in Studio Pro if applicable.

`mdl-examples/bug-tests/116-datagrid2-column-name-mismatch.mdl` is missing. A script that creates a DataGrid2 with a caption-only column and verifies `DESCRIBE PAGE` roundtrips the column name correctly would cover this.

---

### Minor

- `makePropTypeID` / `propKey` helpers in the test file are clear and well-named
- `TestDeriveColumnNameBson_CaptionFallback` directly exercises the previously broken path — good regression anchor

---

**Summary**: Approve after dropping the `template = tmpl` fallback (match DESCRIBE exactly) and adding a `bug-tests/116-*.mdl` script. The core fix is correct.

yscraft:
Thanks for the thorough review, @ako. Both points addressed:

**1. Dropped the `template = tmpl` fallback**
If `Template` is nil, `deriveColumnNameBson` now matches `extractTextContent` on the DESCRIBE side exactly — no caption, falls through to `col{N}`. There was no test covering the fallback path and no known Mendix version that needs it.

**2. Added `mdl-examples/bug-tests/116-datagrid2-column-name-mismatch.mdl`**
Covers three column scenarios: attribute-bound (regression guard), caption-only (the previously broken H1 path), and all-special-char caption → `col{N}` (H2). Instructions included for manual Studio Pro verification.

All 5 unit tests still pass.

ako:
## Review: Fix DESCRIBE PAGE / ALTER PAGE column name mismatch for DataGrid2 columns

**Verdict: Needs work — wrong target file (blocker)**

### Blocker: functions have been moved since the PR was opened

The PR patches `mdl/executor/cmd_alter_page.go`, but `deriveColumnNameBson` and `sanitizeColumnName` no longer live there. They were relocated to `mdl/backend/mpr/page_mutator.go` as part of a backend abstraction refactoring (the functions are at lines 876 and 925 respectively). The current `page_mutator.go` still has both bugs:

```go
// H1 — still broken in page_mutator.go:
case "header":
    if tmpl := dGetDoc(valDoc, "TextTemplate"); tmpl != nil {
        items := dGetArrayElements(dGet(tmpl, "Items"))  // skips Template doc
```
```go
// H2 — still broken in page_mutator.go:
func sanitizeColumnName(caption string) string {
    var result []rune
    for _, r := range caption {
        if ... || r == '_' {
            result = append(result, r)
        } else {
            result = append(result, '_')
        }
    }
    return string(result)  // no TrimFunc — "---" becomes "___" not ""
}
```

The production code is untouched by this PR as it stands.

### Blocker: tests target the wrong package

`mdl/executor/alter_page_test.go` calls `deriveColumnNameBson` and `sanitizeColumnName` directly. Since those functions are in the `backend/mpr` package (not `executor`), the test file won't compile against the current codebase. Running `go test ./mdl/executor/... -run TestDeriveColumnNameBson` returns "no tests to run" confirming the tests don't exist in the compiled binary.

### What needs to change

1. Apply the `TextTemplate → Template → Items` traversal fix and the `sanitizeColumnName` rewrite to **`mdl/backend/mpr/page_mutator.go`** (not `cmd_alter_page.go`)
2. Move the 5 regression tests to **`mdl/backend/mpr/`** (e.g., `page_mutator_test.go`)
3. Drop the changes to `mdl/executor/cmd_alter_page.go` entirely — those functions are gone from that file

### The fix logic is correct

The diagnosis and the fixes themselves are right:
- H1: the intermediate `Template` document must be traversed — `TextTemplate → Template → Items` matches the DESCRIBE side path
- H2: `strings.Map` + `strings.TrimFunc` exactly replicates `deriveColumnName` in `cmd_pages_describe_output.go`
- The `if name := sanitizeColumnName(caption); name != ""` caller guard correctly gates the `col{N}` fallback
- MDL bug test and the 5 unit tests are well-structured

This just needs a rebase targeting the right file.

PR #299, opened 2026-04-24
Document DataGrid2 column name derivation rules and TextTemplate read path

yscraft:
## Summary

Follow-up to #298. Documents two lessons surfaced by issue #116 so future contributors don't repeat the same mistakes.

- **`alter-page.md`** — adds a troubleshooting section under DataGrid Column Operations explaining the silent no-op symptom when column names don't match ALTER PAGE expectations, and documents all three derivation rules (attribute short name → `Description`; sanitized caption → `Order_Status`; all-special-char caption → `col{N}`)
- **`debug-bson.md`** — adds a "Reading TextTemplate in executor code" subsection to Part 10 documenting the correct `TextTemplate → Template → Items` traversal path, with wrong/correct Go examples. The missing `Template` level was the root cause of #116.

## Test plan

- [x] Docs-only change — no code modified
- [x] Spot-checked both files for accuracy against the fixed code in #298

### DIFF
--- a/.claude/skills/mendix/alter-page.md
+++ b/.claude/skills/mendix/alter-page.md
@@ -150,6 +150,14 @@ REPLACE dgProducts.Description WITH {
 
 To discover column names, run `DESCRIBE PAGE Module.PageName` and look at the COLUMN names inside the DATAGRID.
 
+**Troubleshooting: column operation succeeds but does nothing**
+If an ALTER targeting a DataGrid column completes without error but makes no change, the column name used in the statement didn't match any column. The most common cause is a mismatch between what DESCRIBE shows and what ALTER resolves internally. Derivation rules:
+- Attribute-bound column → short attribute name (last segment after `.`): `Module.Entity.Description` → `Description`
+- Caption-only column → sanitized caption (non-alphanumeric replaced with `_`, leading/trailing `_` trimmed): `"Order Status"` → `Order_Status`
+- Caption with only special chars (e.g. `"---"`) → falls back to `col1`, `col2`, … (1-based index)
+
+If the column name you copied from DESCRIBE still doesn't work, check whether the column has an attribute binding — attribute names take priority over captions.
+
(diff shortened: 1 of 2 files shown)


yscraft:
Hey @ako, this is a docs follow-up to #298 — captures the DataGrid2 column name derivation rules and the correct TextTemplate read path as lessons in the skills files.

ako (commented):
## Code Review — PR #299

**Overview**: Docs-only follow-up to #298. Two additions to skill files documenting lessons from issue #116. Short, accurate, and useful.

---

### `alter-page.md` addition

Good troubleshooting section. The three derivation rules are correct and match the fixed code in #298:
- Attribute-bound → short attribute name ✓
- Caption-only → sanitized caption ✓
- All-special-char caption → `col{N}` ✓

The final tip ("attribute names take priority over captions") is accurate and pre-empts a likely follow-up confusion.

One small wording note: "completes without error but makes no change" implies this is always silent. Worth clarifying *why* — the column traversal in `dSet` is intentionally silent for absent keys (documented in `page_mutator.go`). So this is by design, not a missing error. Optional — the current wording is still accurate.

---

### `debug-bson.md` addition

The WRONG/CORRECT Go snippet directly shows the mistake and the fix. `dGet(template, "Items")` on a nil `template` is safe (iterating a nil `bson.D` returns nil → empty slice → loop doesn't run), so the example works correctly even without an explicit nil guard.

Breadth note: "This applies to any executor function that reads column headers, button captions, or any other translatable text stored as `Forms$ClientTemplate`" — good, this generalises the lesson beyond DataGrid2.

---

### Dependency

Must merge after #298 (and after #298's requested changes are addressed — specifically the `template = tmpl` fallback). If the fallback is dropped from #298, the docs here remain accurate: the CORRECT example already doesn't rely on it.

---

**Overall**: Approve — clean, accurate, and useful. Merge after #298.

PR #305, opened 2026-04-26
fix: drop empty else branch from describer output

### LINKED ISSUE #304: DESCRIBE MICROFLOW emits an empty else branch for if-without-else
Problem:

`DESCRIBE MICROFLOW` can print an `else` keyword even when the false branch of an exclusive split jumps directly to the merge and has no body. Re-executing the described MDL normalizes the empty else away, so describe/exec/describe produces textual drift.

Expected:

For `if Condition then ... end if;`, the describer should not emit an `else` line unless the model has an actual false-branch body.

Impact:

This breaks byte-exact roundtrip validation for else-less IF shapes and adds noise to generated MDL diffs.

Proposed fix:

Detect when the false branch destination is the split merge point and skip emitting the `else` keyword in that case.

hjotha:
Part of #332.

Fixes #304.

## Summary

`DESCRIBE MICROFLOW` emitted an `else` line for IF statements whose false flow jumps directly to the merge point. That produced a spurious empty branch in MDL output and broke describe/exec/describe equality.

## Root cause

The split traversal printed `else` whenever a false flow existed, without checking whether the false flow was only the direct split-to-merge edge that represents "no else body".

## Fix

Skip the `else` keyword when the false branch destination is the matching merge point. Explicit false-branch bodies are still emitted normally.

## Tests

Added focused traversal coverage for an IF without an ELSE body.

## Validation

- [x] `make build`
- [x] `make test`
- [x] `make lint-go`
- [x] `make test-integration`

## Agentic Code Testing

- [x] Synthetic regression coverage exercises the describe output shape agents consume.

## Test plan

- [x] Else-less IF does not emit a blank `else`.
- [x] Existing IF/ELSE traversal remains covered by the existing test suite.

### DIFF
--- a/mdl-examples/bug-tests/304-describer-empty-else-branch.mdl
+++ b/mdl-examples/bug-tests/304-describer-empty-else-branch.mdl
@@ -1,0 +1,57 @@
+-- ============================================================================
+-- Bug #304: Empty `else` branch emitted by DESCRIBE MICROFLOW
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   `describe microflow` printed an `else` line for every IF whose false
+--   branch jumped straight to the merge point (the shape produced by an
+--   `if X then ... end if;` with no else body and a non-terminal true branch).
+--   Output looked like
+--     if $X > 0 then
+--       ...
+--     else
+--     end if;
+--   The empty else block re-parsed as a real branch, breaking
+--   describe → exec → describe fixpoint and creating cosmetic BSON drift.
+--
+-- After fix:
+--   Split traversal in cmd_microflows_show_helpers.go skips the `else`
+--   keyword when the false branch destination is the matching merge point.
+--   Explicit else bodies (with at least one activity) are still emitted.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/304-describer-empty-else-branch.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest304.MF_IfWithoutElse"
+--   The output must NOT contain an `else` keyword and MUST be a fixpoint
+--   under describe → exec → describe.
+-- ============================================================================
+
+create module BugTest304;
+
+create microflow BugTest304.MF_IfWithoutElse (
+  $value: integer
+)
+begin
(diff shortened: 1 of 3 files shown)


ako (commented):
## Code Review — PR #305

**Overview**: One-line guard in `traverseFlow` to suppress the spurious `else` keyword when the false branch connects directly to the split's matching merge node. Clean, focused, correct.

---

### Fix correctness

```go
falseHasBody := falseFlow != nil && falseFlow.DestinationID != mergeID
if falseHasBody {
    *lines = append(*lines, indentStr+"else")
```

The invariant is sound: if the false flow's destination **is** the merge node, there are no activities on the false branch — the edge is purely structural. If it points to anything else, there are activities between the split and merge on the false side. The nil guard is preserved from the original condition.

The only theoretical edge case is if `mergeID == ""` (zero value) and `falseFlow.DestinationID == ""` — both would be empty strings and false branch would be suppressed. In practice this can't happen since a split always has a resolvable merge from `splitMergeMap`, but a defensive comment wouldn't hurt.

---

### Tests

`TestTraverseFlow_IfWithoutElse` is well-structured:
- Uses the existing `mkBranchFlow("split", "merge", ...)` to represent the direct false-to-merge edge
- Asserts no `else` line is emitted
- Asserts `end if;` is still present (the IF itself is not suppressed)
- The helper wrapper `e.traverseFlow` correctly passes `nil` for `flowsByDest` — this suppresses `@anchor` emission, which is correct for this test

Bug test script `304-describer-empty-else-branch.mdl` is thorough:
- Positive case: IF without ELSE — verifies no `else` keyword in output
- Negative control: IF with ELSE — verifies explicit else body is still preserved
- Comment explains the symptom, the fix, and how to verify it in Studio Pro

---

### Minor

- The comment block in the fix (`// Emit the ELSE branch only if it has statements...`) accurately explains the logic and is worth keeping
- `flowsByOrigin` wires up `"split"` → two branch flows (true to `true_act`, false to `merge`) — exactly the shape described in the root cause. The test directly targets the regression.

---

**Overall**: Approve. Minimal, correct, well-tested fix for a describe/exec/describe fixpoint failure.

PR #307, opened 2026-04-26
fix: do not emit annotations before unsupported-action comments

### LINKED ISSUE #306: DESCRIBE MICROFLOW emits orphan annotations before unsupported-action comments
Problem:

When the describer cannot render an activity and falls back to a comment, it can still emit `@position`, `@anchor`, or similar activity annotations before that comment. The grammar accepts annotations only before real microflow statements, so the output fails to parse on `mxcli exec`.

Expected:

Unsupported activity comments should remain visible in describe output, but they should not be preceded by statement annotations that cannot attach to a real statement.

Impact:

Roundtrip execution fails for microflows containing unsupported activity types, even though the unsupported activity itself is represented as a comment.

Proposed fix:

Skip object annotations when the formatted statement is a comment placeholder.

hjotha:
Part of #332.

Fixes #306.

## Summary

Unsupported microflow activities are represented as comments in `DESCRIBE MICROFLOW` output. This PR prevents activity annotations from being emitted before those comments, so the generated MDL remains parseable.

## Root cause

The describer emitted object annotations before every formatted activity result. For unsupported activity placeholders the formatted result starts with `--`, leaving annotations orphaned because comments are not valid `microflowStatement` targets.

## Fix

When a formatted statement begins with a comment marker, emit only the comment and skip `emitObjectAnnotations`.

## Tests

Added traversal coverage for unsupported/comment activity output with preceding activity metadata.

## Validation

- [x] `make build`
- [x] `make test`
- [x] `make lint-go`
- [x] `make test-integration`

## Agentic Code Testing

- [x] Synthetic regression coverage verifies generated MDL stays parseable.

## Test plan

- [x] Unsupported activity comments remain visible.
- [x] No orphan annotations are emitted before comment placeholders.

### DIFF
--- a/mdl-examples/bug-tests/306-describer-annotations-before-unsupported-comment.mdl
+++ b/mdl-examples/bug-tests/306-describer-annotations-before-unsupported-comment.mdl
@@ -1,0 +1,61 @@
+-- ============================================================================
+-- Bug #306: Annotations emitted before unsupported-activity comments
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   Activities whose BSON type isn't yet implemented in mxcli's describer
+--   render as `-- unsupported: <Type>` comments. Before the fix, the
+--   describer still prefixed these comment lines with `@position`,
+--   `@anchor`, etc. The grammar only accepts annotations as prefixes of
+--   real microflow statements (not comments), so the resulting MDL failed
+--   to re-parse with errors like:
+--     line N:0 mismatched input '@' expecting {...}
+--
+-- After fix:
+--   When a formatted statement begins with `--`, `emitObjectAnnotations`
+--   is skipped — only the comment is emitted. Annotations on SUPPORTED
+--   activities continue to be emitted as before.
+--
+-- Reproducibility note:
+--   The unsupported-activity branch can only be triggered by an .mpr
+--   containing a BSON activity type that mxcli does not yet describe
+--   (typical examples are very new or rarely-used activity types). Pure
+--   MDL cannot create such an activity. The Go-side regression
+--   `TestTraverseFlow_UnsupportedActivitySkipsAnnotations` covers that
+--   path directly.
+--
+--   This script provides a POSITIVE CONTROL: a microflow whose every
+--   activity carries position / caption / anchor annotations. Its describe
+--   output must (a) parse with `mxcli check`, and (b) round-trip cleanly,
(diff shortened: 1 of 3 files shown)


ako (commented):
## Code Review — PR #307

**Overview**: One-guard fix in `emitActivityStatement` to suppress annotations before unsupported-activity comments. Clean and well-motivated.

---

### Fix correctness

```go
if strings.HasPrefix(strings.TrimSpace(stmt), "--") {
    *lines = append(*lines, indentStr+stmt)
    return
}
```

- Correctly intercepts before `emitObjectAnnotations`, which unconditionally emits `@position(x, y)` for every object (no nil check, no condition) — so without this guard, every unsupported activity would produce orphaned annotations
- `--` is exclusively a comment marker in MDL; no valid statement starts with `--`, so the prefix check is unambiguous
- `strings.TrimSpace` is defensive — `stmt` at this point doesn't carry leading whitespace (indent is prepended separately) — harmless but could be dropped

---

### Test correctness

`TestTraverseFlow_UnsupportedActivitySkipsAnnotations` is valid and would fail without the fix. `mkObj("soap")` produces a zero-valued `BaseMicroflowObject`, and `emitObjectAnnotations` calls:

```go
pos := obj.GetPosition()
*lines = append(*lines, indentStr+fmt.Sprintf("@position(%d, %d)", pos.X, pos.Y))
```

unconditionally — so `@position(0, 0)` would be emitted before the comment without the guard. The test correctly catches this.

The check for `@position` specifically is sufficient: `emitObjectAnnotations` is a single call that emits all annotation types atomically — if `@position` is absent, the rest are absent too.

---

### Bug test script

The honest note that pure MDL can't reproduce the unsupported-activity path (requires an .mpr with an unknown action type) is appreciated. The positive control (annotated microflow that must still parse correctly after `DESCRIBE`) is the right substitute — it confirms the annotation-emission path for supported activities is unaffected.

---

### Minor

- The fix is general: any new unsupported action type that returns a `-- ...` statement automatically benefits without any additional changes
- Adding `@anchor` and `@caption` to the negative assertions alongside `@position` would make the test stricter, but since all come from the same `emitObjectAnnotations` call, the current single check is adequate

---

**Overall**: Approve. Minimal, correct fix with a meaningful test that would fail without it.

ako:
@hjotha can you rebase to fix the conflict?

PR #309, opened 2026-04-26
fix: preserve commit error handling through describe/exec/describe

### LINKED ISSUE #308: Commit error handling can be dropped during MPR rewrite
Problem:

Commit object activities with rollback error handling can roundtrip from `commit $X on error rollback;` to `commit $X;` because the writer omits rollback as a default and the parser does not restore it when the BSON field is absent.

Expected:

Rollback error handling should survive describe/exec/describe for commit activities.

Impact:

The textual roundtrip loses an explicit behavior setting and creates unnecessary drift for commit actions.

Proposed fix:

Make parser and writer handling symmetric: parse absent commit error handling as rollback, and always write the field with rollback as the default value.

hjotha:
Part of #332.

Fixes #308.

## Summary

`commit $X on error rollback;` could roundtrip as plain `commit $X;` after rewriting an MPR.

## Root cause

The commit writer omitted `ErrorHandlingType` when it was rollback, treating it as implicit. The parser then left absent `ErrorHandlingType` empty, so the describer had no rollback setting to print on the next describe.

## Fix

Make commit parser/writer behavior symmetric:

- Parser defaults absent commit `ErrorHandlingType` to rollback.
- Writer always emits commit `ErrorHandlingType`, defaulting to rollback.

## Tests

Added parser/writer regression coverage for commit rollback preservation.

## Validation

- [x] `make build`
- [x] `make test`
- [x] `make lint-go`
- [x] `make test-integration`

## Agentic Code Testing

- [x] Synthetic parser/writer coverage verifies the roundtrip behavior agents rely on.

## Test plan

- [x] Commit rollback survives BSON parse/write.
- [x] Existing action writer behavior remains covered by the full test suite.

### DIFF
--- a/mdl-examples/bug-tests/308-commit-on-error-rollback-roundtrip.mdl
+++ b/mdl-examples/bug-tests/308-commit-on-error-rollback-roundtrip.mdl
@@ -1,0 +1,53 @@
+-- ============================================================================
+-- Bug #308: `commit ... on error rollback` lost during describe → exec → describe
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   Authoring `commit $X on error rollback;` and then re-executing the
+--   describe output silently dropped the `on error rollback` clause —
+--   subsequent describes printed plain `commit $X;`. In Studio Pro the
+--   "Action call action" still showed Rollback, but mxcli's view of the
+--   model diverged from the BSON, so any further mutation could mis-write
+--   the field.
+--
+-- Root cause:
+--   The commit writer omitted `ErrorHandlingType` when its value was
+--   Rollback (treating Rollback as implicit). The parser then left the
+--   absent field as the empty string, so the describer printed `commit $X`
+--   without a clause.
+--
+-- After fix:
+--   Parser defaults absent commit `ErrorHandlingType` to Rollback; the
+--   writer always emits the field, defaulting to Rollback. The behaviour
+--   is symmetric and round-trips cleanly.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/308-commit-on-error-rollback-roundtrip.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest308.MF_CommitRollback"
+--   The output must contain `on error rollback` and must be a fixpoint
+--   under describe → exec → describe.
+-- ============================================================================
+
+create module BugTest308;
+
+create entity BugTest308.Item (
+  Name : string(100)
+);
+/
+
+create microflow BugTest308.MF_CommitRollback (
+  $Item: BugTest308.Item
+)
+begin
+  commit $Item on error rollback;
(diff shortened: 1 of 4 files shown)


ako (commented):
## Code Review — PR #309

**Overview**: Fixes a roundtrip loss where `commit $X on error rollback;` silently became `commit $X;` after describe → exec → describe. Root cause and fix are clearly identified.

---

### Fix correctness

**Parser** (`parser_microflow.go`):
```go
// before
if errType, ok := raw["ErrorHandlingType"].(string); ok {
    action.ErrorHandlingType = microflows.ErrorHandlingType(errType)
}

// after
action.ErrorHandlingType = microflows.ErrorHandlingType(extractString(raw["ErrorHandlingType"]))
if action.ErrorHandlingType == "" {
    action.ErrorHandlingType = microflows.ErrorHandlingTypeRollback
}
```
`extractString` returns `""` for absent/nil keys — the explicit default to `Rollback` is correct.

**Writer** (`writer_microflow_actions.go`):
```go
{Key: "ErrorHandlingType", Value: stringOrDefault(string(a.ErrorHandlingType), "Rollback")},
```
This directly matches the pattern used by every other action that has `ErrorHandlingType` (e.g., `CreateObjectAction` at line 38, `DeleteObjectAction` at line 51, `RollbackObjectAction` at line 175, etc.). The `CommitObjectsAction` case was the only outlier conditionally omitting it.

Emitting `"Rollback"` explicitly does not cause Studio Pro issues — other actions already do this unconditionally.

---

### Tests

`TestParseCommitAction_ErrorHandlingTypeExplicit` — verifies `"Continue"` is preserved correctly.  
`TestParseCommitAction_ErrorHandlingTypeDefaultsToRollback` — directly targets the regression: absent `ErrorHandlingType` must default to `Rollback`.

Both tests are in `parser_microflow_test.go`, which is the right layer for BSON parsing regressions.

**Minor gap**: There's no corresponding writer test verifying `ErrorHandlingType` is always present in the serialized BSON. A test like:
```go
func TestSerializeCommitAction_AlwaysEmitsErrorHandlingType(t *testing.T) {
    doc := serializeMicroflowAction(&microflows.CommitObjectsAction{...})
    // assert ErrorHandlingType key is present in doc
}
```
would guard against the condition being re-introduced. Not blocking given the pattern is consistent with other actions, but worth adding to `writer_microflow_actions_test.go` if that file exists.

---

### Bug test script

`308-commit-on-error-rollback-roundtrip.mdl` is correct:
- Positive case: `on error rollback` must survive describe → exec → describe
- Negative control: `on error continue` must also survive (confirms the writer doesn't flatten all to Rollback)

---

**Overall**: Approve. The fix correctly aligns `CommitObjectsAction` with the universal pattern in the writer. Parser + writer are now symmetric and the bug test covers the full roundtrip.

PR #311, opened 2026-04-26
fix: honor branch anchor destination on split-to-merge flows

### LINKED ISSUE #310: Split-to-merge flows lose branch destination anchors
Problem:

For IF branch flows that go directly from a split to a merge, the builder applies the branch `from` anchor but loses the branch `to` anchor. The next describe changes branch anchor output, for example replacing an intended destination side with the default destination.

Expected:

Branch-level `@anchor(... true: (...), false: (...))` metadata should preserve both origin and destination sides for split-to-merge shortcut flows.

Impact:

Roundtrip output drifts for visually meaningful branch anchors.

Proposed fix:

Pass the branch anchor as both origin and destination anchor when constructing split-to-merge shortcut flows.

hjotha:
Part of #332.

Fixes #310.

## Summary

IF branch anchors on direct split-to-merge flows preserved the branch origin side but dropped the branch destination side.

## Root cause

Several split-to-merge shortcuts called `applyUserAnchors` with the branch anchor as the origin override and `nil` as the destination override. That let the default destination side replace the user-provided branch `to` side.

## Fix

Use the branch anchor for both origin and destination overrides in the split-to-merge shortcut paths, matching the existing behavior for branches with a first statement.

## Tests

Added focused builder coverage for branch destination anchors on split-to-merge flows.

## Validation

- [x] `make build`
- [x] `make test`
- [x] `make lint-go`
- [x] `make test-integration`

## Agentic Code Testing

- [x] Synthetic builder regression coverage verifies anchor metadata survives generated MDL roundtrips.

## Test plan

- [x] False/true branch destination sides survive a builder roundtrip.
- [x] Existing anchor defaults remain covered by the full test suite.

### DIFF
--- a/mdl-examples/bug-tests/310-anchor-split-to-merge-destination.mdl
+++ b/mdl-examples/bug-tests/310-anchor-split-to-merge-destination.mdl
@@ -1,0 +1,63 @@
+-- ============================================================================
+-- Bug #310: Branch anchor `to:` side dropped on direct split-to-merge flows
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   When an IF had a branch anchor whose `to:` side wasn't the default Left
+--   AND that branch body was empty (so the split connected straight to the
+--   merge), the writer applied only the branch FROM side and let the
+--   default destination (Left) overwrite the user-provided `to:` side.
+--   Re-describing turned
+--     @anchor(to: left, false: (from: bottom, to: top))
+--   into
+--     @anchor(to: left, false: (from: bottom, to: left))
+--   silently changing the visual layout in Studio Pro on roundtrip.
+--
+-- Root cause:
+--   Several split-to-merge shortcuts called `applyUserAnchors` with the
+--   branch anchor only as the origin override and `nil` as the destination
+--   override.
+--
+-- After fix:
+--   `applyUserAnchors` now receives the branch anchor for both origin and
+--   destination overrides on split-to-merge shortcuts, matching the
+--   behavior used when a branch has a first statement.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/310-anchor-split-to-merge-destination.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest310.MF_AnchorFalseBranch"
+--   The output must preserve `false: (from: bottom, to: top)` on the
+--   IF anchor and must be a fixpoint under describe → exec → describe.
+-- ============================================================================
+
+create module BugTest310;
+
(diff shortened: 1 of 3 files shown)


ako (commented):
## Code Review — PR #311

**Overview**: Three symmetric one-character fixes in `addIfStatement` where split-to-merge shortcut flows were applying only the origin side of the branch anchor, leaving the destination side at its default. Minimal, correct, well-tested.

---

### Fix correctness

`applyUserAnchors` signature:
```go
func applyUserAnchors(flow *SequenceFlow, origin *ast.FlowAnchors, destination *ast.FlowAnchors)
// origin.From  → flow.OriginConnectionIndex
// destination.To → flow.DestinationConnectionIndex
```

Before:
```go
applyUserAnchors(flow, falseBranchAnchor, nil)   // destination.To never applied
```

After:
```go
applyUserAnchors(flow, falseBranchAnchor, falseBranchAnchor)  // both sides applied
```

The function guards each side independently with `!= AnchorSideUnset`, so passing `branchAnchor` as destination is safe for partial anchors (e.g. `@anchor(false: (from: bottom))` without a `to:`) — `DestinationConnectionIndex` is left at the default when `To` is unset. No regression risk.

All three fix locations are identical in shape:
1. Empty THEN body (within-else path) — covered by `TestBuilder_AnchorTrueBranchTo_EmptyThenIfWithElse`
2. FALSE direct-to-merge path (IF-without-ELSE) — covered by `TestBuilder_AnchorFalseBranchTo_IfWithoutElse`
3. Empty THEN body (no-else variant) — same pattern, consistent with the other two

---

### Tests

Both new tests follow the established `buildWithAnchors` + `hasFlow` pattern from the existing anchor test file. The assertions are precise:
- `hasFlow(oc.Flows, AnchorBottom, AnchorTop)` — verifies the specific `OriginConnectionIndex=2, DestinationConnectionIndex=0` pair exists (matching `ast.AnchorSideBottom=2`, `ast.AnchorSideTop=0`)
- The test godoc correctly names the pre-fix symptom (`to: left` overwriting `to: top`)

---

### Bug test script

`310-anchor-split-to-merge-destination.mdl` provides two concrete cases:
- `MF_AnchorFalseBranch` — IF-without-ELSE with `false: (from: bottom, to: top)`
- `MF_AnchorEmptyThen` — empty THEN with `true: (from: right, to: top)`

Both reproduce the lost `to:` side. The usage comment correctly describes the fixpoint test (`describe → exec → describe` must preserve the anchor).

---

**Overall**: Approve. Three-character change, correct reasoning, solid regression tests.

PR #313, opened 2026-04-26
fix: skip reference checks for excluded microflows

### LINKED ISSUE #312: Reference validation reports stale references inside excluded microflows
Problem:

`mxcli check --references` reports missing references inside excluded microflows, even though Studio Pro allows excluded documents to keep stale references because they are not part of the runnable application.

Expected:

Reference validation should skip the body of excluded microflows while still validating included microflows.

Impact:

Large project audits can fail on intentionally excluded legacy scaffolding instead of actionable model errors.

Proposed fix:

Short-circuit microflow reference collection when a `CreateMicroflowStmt` is marked excluded.

hjotha:
Part of #332.

Fixes #312.

## Summary

Reference validation now skips excluded microflows. Included microflows are still validated normally.

## Root cause

The validator walked every microflow body and collected missing microflow/page/java-action references without considering the microflow's excluded flag.

## Fix

Skip reference collection for excluded `CreateMicroflowStmt` values. This aligns validation with writer behavior and Studio Pro's treatment of excluded documents.

## Tests

Added paired regression tests:

- Excluded microflow with a missing call passes reference validation.
- Included microflow with the same missing call still reports an error.

## Validation

- [x] `make build`
- [x] `make test`
- [x] `make lint-go`
- [x] `make test-integration`

## Agentic Code Testing

- [x] Synthetic validation coverage verifies excluded microflows do not create false positives.

## Test plan

- [x] Excluded stale references are ignored.
- [x] Included stale references remain errors.

### DIFF
--- a/mdl-examples/bug-tests/312-validate-skip-excluded-microflows.mdl
+++ b/mdl-examples/bug-tests/312-validate-skip-excluded-microflows.mdl
@@ -1,0 +1,63 @@
+-- ============================================================================
+-- Bug #312: Reference validation flagged broken calls inside excluded microflows
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   `mxcli check --references` walked every microflow body — including ones
+--   marked `@excluded` — and reported missing microflow / page / java-action
+--   references in them. Excluded documents are not part of the build, and
+--   Studio Pro tolerates dangling references in them, so this produced
+--   false positives during agentic workflows that legitimately stash
+--   broken intermediate state inside excluded scaffolding.
+--
+-- Root cause:
+--   The validator collected references from every CreateMicroflowStmt
+--   without checking the Excluded flag.
+--
+-- After fix:
+--   `validate.go` skips reference collection for excluded microflows.
+--   Included microflows are still validated normally, so the negative
+--   case (typo in a real call) still fails the check.
+--
+-- Usage:
+--   mxcli check --references mdl-examples/bug-tests/312-validate-skip-excluded-microflows.mdl -p app.mpr
+--   The check must succeed: the excluded microflow's broken call to
+--   BugTest312.NoSuchTarget is ignored.
+--
+--   To verify the negative case still triggers, remove `@excluded` from
+--   the second microflow and re-run — `mxcli check --references` should
+--   then report the missing reference.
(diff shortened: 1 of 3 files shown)


ako (commented):
## Review: fix: skip reference checks for excluded microflows

**Summary**: Clean, focused fix. One early-return guard in `validateMicroflowReferences` skips reference validation for excluded microflows, matching Studio Pro's behavior of tolerating stale references in documents excluded from the runnable app.

### What the fix does

```go
if s.Excluded {
    return nil
}
```

Added to `validateMicroflowReferences` after the existing nil/empty guards. When a microflow carries the `@excluded` flag, reference checks are skipped entirely — Studio Pro allows these documents to reference deleted or renamed targets without raising an error.

### Code quality

- **Correct placement**: The guard is positioned with the other early-return preconditions, before any lookups happen.
- **Comment is accurate**: Explains *why* (Studio Pro behavior for excluded documents) not just *what*.
- **No over-engineering**: A single `if` is exactly the right scope for this fix.

### Test coverage

Two regression tests in `bugfix_regression_test.go`:
- `TestValidateMicroflowReferencesSkipsExcludedMicroflow`: excluded microflow with a missing callee — expects no error ✓
- `TestValidateMicroflowReferencesReportsIncludedMissingMicroflow`: same setup without `Excluded` — still reports error ✓

Both cases covered, using `MockBackend` cleanly. This is the right way to test this: the positive guard (excluded passes) and the negative guard (included still fails) are both exercised.

### Bug test script

`mdl-examples/bug-tests/312-validate-skip-excluded-microflows.mdl` tests the end-to-end MDL path: an `@excluded` microflow calling `BugTest312.NoSuchTarget` (would fail without the fix), plus an included microflow calling a valid target (regression guard).

### Minor note

The bug test script creates module `BugTest312` and a microflow `ValidTarget` — but `ValidTarget` is called from the included microflow. Ensure `BugTest312.ValidTarget` is actually defined in the test script so the included case passes; if it is missing the positive control would also fail. From the diff it looks like it is defined, just worth a double-check when running the test.

**Verdict**: Approve. The fix is correct, minimal, well-tested, and matches Studio Pro semantics exactly.

PR #315, opened 2026-04-26
feat(executor): allow MXCLI_EXEC_TIMEOUT to override the per-statement timeout

### LINKED ISSUE #314: Executor timeout is hard-coded for long-running project operations
Problem:

The executor has a fixed per-statement timeout. Large projects and long-running audit operations can legitimately exceed the default, forcing users to fork mxcli or avoid the operation.

Expected:

Operators should be able to increase the executor timeout without changing code.

Impact:

Large project audit workflows are fragile and can time out before producing useful diagnostics.

Proposed fix:

Add an environment variable override for the executor timeout while preserving the existing default behavior.

hjotha:
Part of #332.

Closes #314.

## Summary

Adds `MXCLI_EXEC_TIMEOUT` so long-running executor operations can use a larger per-statement timeout without rebuilding mxcli.

## Configuration

Accepted values:

- Go duration strings, for example `12m`, `2h30m`, or `45s`.
- Bare seconds, for example `900`.

Unset, empty, or invalid values fall back to the existing default.

## Root cause

The executor used a hard-coded timeout constant, which is too rigid for large-project audit and describe workflows.

## Fix

Replace direct use of the constant with a `configuredExecuteTimeout()` helper read on each `Execute()` call.

## Tests

Added helper tests for Go duration values, bare-second values, and invalid-value fallback.

## Validation

- [x] `make build`
- [x] `make test`
- [x] `make lint-go`
- [x] `make test-integration`

## Agentic Code Testing

- [x] No new MDL syntax.
- [x] Environment-variable behavior is documented in the PR and covered by unit tests.

## Test plan

- [x] Duration override works.
- [x] Bare-second override works.
- [x] Invalid override falls back to the default.

### DIFF
--- a/mdl/executor/executor.go
+++ b/mdl/executor/executor.go
@@ -7,6 +7,8 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"os"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ako (commented):
## Review: feat(executor): allow MXCLI_EXEC_TIMEOUT to override the per-statement timeout

**Summary**: Clean, minimal feature. Replaces the hard-coded `executeTimeout` constant with a `configuredExecuteTimeout()` helper that reads `MXCLI_EXEC_TIMEOUT` from the environment, with fallback to the existing default.

### Code quality

- **Re-read on every call**: The env var is re-read inside `Execute()` on each invocation rather than cached at startup. This is fine for this use case — `os.Getenv` is cheap and allows the value to be changed between runs in a REPL session without restarting the process. Worth keeping as-is.
- **Dual format support**: Both Go duration strings and bare seconds are accepted — the two most natural ways a user would specify a timeout. The parse order (duration first, then seconds) is correct since bare integers would also pass `time.ParseDuration` for values like `"5s"` — wait, actually bare `"900"` does *not* parse as a duration (it requires a unit suffix), so the order is fine.
- **Positive-only guard**: `d > 0` and `seconds > 0` both reject zero/negative values — correct, since a zero timeout would fire immediately.
- **Shadow variable name**: `executeTimeout := configuredExecuteTimeout()` shadows the old constant name. Since the constant was renamed to `defaultExecuteTimeout`, there's no actual shadow issue, but the local variable name `executeTimeout` is a bit confusing now that the constant has a different name. A clearer name like `timeout` or `stmtTimeout` would avoid the ghost of the old constant — minor style point, not a blocker.

### Test coverage

Three unit tests using `t.Setenv` (parallel-safe, automatically restored):
- Duration string (`12m`) → correct
- Bare seconds (`900`) → 15 minutes ✓
- Invalid value (`"invalid"`) → falls back to default ✓

Missing: test for unset variable (though this is implicitly covered by the fallback test if the env var happens to not be set in CI, it's more robust to add `t.Setenv("MXCLI_EXEC_TIMEOUT", "")` explicitly). Low priority.

### Documentation gap

The env var is documented in the PR description but not in any user-facing location. Worth adding to:
- CLI help text (e.g. a note in the executor or `mxcli exec` command's `Long` description)
- Or at minimum `docs/01-project/MDL_QUICK_REFERENCE.md`

Not a blocker, but `MXCLI_EXEC_TIMEOUT` is invisible until documented somewhere discoverable.

### Verdict

Approve. The fix is correct, safe, and minimal. The local variable naming nit and documentation gap are worth a follow-up but don't block merging.

PR #319, opened 2026-04-26
feat: support free microflow annotations

### LINKED ISSUE #318: MDL cannot distinguish free microflow notes from activity-bound annotations
Problem:

Standalone annotation notes that describe emits near an activity can be parsed as if they are bound to that activity. On exec, the note is attached to the activity metadata and no longer roundtrips as a free note.

Expected:

MDL should preserve a free annotation note separately from activity-bound annotation metadata when order makes the distinction visible.

Impact:

Describe/exec/describe can move or reclassify user-visible notes in microflow output.

Proposed fix:

Add explicit AST support for free annotations and document the order-based discrimination rule.

hjotha:
Part of #332.

Closes #318.

## Summary

Adds first-class support for free microflow annotations so standalone notes can roundtrip without being attached to the following activity.

## Semantics

An `@annotation 'text'` is treated as free when it appears before activity-binding metadata such as `@position`, `@caption`, `@color`, `@excluded`, or `@anchor`, and a later annotation binds the activity. Activity-bound annotations remain supported as before.

## Root cause

The MDL AST had only one annotation text field, so free notes and activity-bound notes collapsed into the same property during parsing and graph building.

## Fix

- Add `ActivityAnnotations.FreeAnnotation`.
- Classify free annotations in the visitor using annotation order.
- Thread free annotations through the microflow builder.
- Document the syntax and open design questions in a draft proposal.

## Tests

Added parser and builder coverage for free-versus-attached annotation order plus a doctype fixture.

## Documentation

- `docs/11-proposals/PROPOSAL_microflow_free_annotation.md`
- `docs/11-proposals/README.md`
- `docs/01-project/MDL_QUICK_REFERENCE.md`
- `.claude/skills/mendix/write-microflows.md`
- `mdl-examples/doctype-tests/free_annotation.test.mdl`

## Validation

- [x] `make build`
- [x] `make test`
- [x] `make lint-go`
- [x] `make test-integration`
- [x] `./bin/mxcli check mdl-examples/doctype-tests/free_annotation.test.mdl`

## Agentic Code Testing

- [x] Skills and language reference updated for the new annotation distinction.
- [x] Example fixture passes `mxcli check`.

## Test plan

- [x] Free annotation order parses into `FreeAnnotation`.
- [x] Activity-bound annotations still attach normally.
- [x] Example fixture checks cleanly.


hjotha:
Addressed the review concern about the model-to-DESCRIBE side of free annotations in follow-up commit `98729a4`.

Clarification: this path does not use `ActivityAnnotations`; existing MPR/BSON parsing produces `Microflows$Annotation` objects plus `AnnotationFlow` links. DESCRIBE separates standalone notes from attached notes via `collectFreeAnnotations` and `buildAnnotationsByTarget`.

I extracted the free-annotation prefixing into `prependFreeAnnotationLines` and added `TestPrependFreeAnnotationLines_ModelAnnotationsStayFree`, which starts from a model object collection containing both a standalone annotation and an attached annotation. The test verifies that only the standalone note is emitted as a free `@annotation` before activity metadata, while the attached note remains attached.

Validation rerun:
- `go test ./mdl/executor -run 'TestPrependFreeAnnotationLines_ModelAnnotationsStayFree|TestFreeAnnotationBeforePositionStaysUnattached'`
- `make build`
- `make test`
- `make lint-go`
- `make test-integration`

ako (commented):
## Review: feat: support free microflow annotations

**Summary**: Good feature with a clean visitor-level disambiguation rule (annotation before activity-binding metadata = free, after = attached). The core logic is correct. A few structural concerns worth discussing.

### Dual-field redundancy in `ActivityAnnotations`

The AST now carries both:
```go
FreeAnnotation  string   // first free annotation (convenience)
FreeAnnotations []string // all free annotations
```

These two fields are always kept in sync — the visitor, the builder, and `mergeStatementAnnotations` all set both together. This creates an implicit invariant that's easy to violate: test helpers like `TestFreeAnnotationBeforePositionStaysUnattached` set only `FreeAnnotation` without `FreeAnnotations`, which exercises a defensive code path in `freeAnnotationTexts()` that deduplicates by value. If two different free annotations happened to have the same text, they'd silently collapse there.

A simpler design: drop `FreeAnnotation string` from the AST and use `FreeAnnotations[0]` everywhere a "first annotation" is needed. The deduplication loop in `freeAnnotationTexts()` goes away entirely, and test fixtures that need a single free annotation just set `FreeAnnotations: []string{"note"}`.

### `freeAnnotationTexts()` dedup is load-bearing only because of the dual field

The function's deduplication loop (check whether `FreeAnnotation` is already in `FreeAnnotations` before appending it) exists solely to handle the single-field case. Removing the dual field removes the need for this function entirely — callers can use `ann.FreeAnnotations` directly.

### No describe→exec→describe roundtrip test

The doctype fixture `free_annotation.test.mdl` verifies that the script parses cleanly (`mxcli check`) but doesn't exercise the full roundtrip: describe an existing microflow with a free annotation, exec the output, describe again, and confirm the annotation stays free. That's the regression case that matters most for BSON drift. The proposal doc flags this as a desirable test; it would be worth adding to the bug-tests folder as a live validation step.

### Flush ordering in `addStatement` vs `buildFlowGraph`

Free annotations are flushed and cleared inside `addStatement` (before the activity is processed), and then `buildFlowGraph` checks leftover free annotations after all statements. The clearing in `addStatement` prevents double-flush — this is correct. Worth a comment linking the two sites so a future reader doesn't wonder why both flush without a guard.

### Visitor lookahead is O(N²) worst-case

`hasLaterActivityAnnotation` rescans the tail of the annotation list for every `@annotation` token. In practice annotation lists are short (< 10 items), so this is fine. Just flagging for awareness.

### Minor: `seenActivityMetadata` flag is not reset between annotations

This is fine by design — the flag is intentionally sticky (once you've seen `@position`, all subsequent annotations are attached). The name and placement make this clear; no action needed.

### Tests

Visitor tests cover the three key cases (before, multiple before, after). Builder tests cover single and multi free annotation. `prependFreeAnnotationLines` test is well-structured and verifies the attached annotation is not duplicated as free. Good coverage of the happy paths.

### Verdict

Approve with a recommendation to collapse `FreeAnnotation`/`FreeAnnotations` into a single `[]string` field before merge. The dual-field design works but is a maintenance hazard. Everything else is solid.

PR #321, opened 2026-04-26
fix: suppress default anchor fragments in describe

### LINKED ISSUE #320: DESCRIBE MICROFLOW emits default anchor fragments that create cosmetic drift
Problem:

`DESCRIBE MICROFLOW` prints explicit anchor fragments for connection sides that match MDL defaults. Re-execution can omit or normalize those defaults, creating noisy cosmetic diffs.

Expected:

Describe output should omit anchor fragments that are equivalent to the default connection sides, while still preserving non-default anchors.

Impact:

Roundtrip audits produce avoidable cosmetic mismatch noise.

Proposed fix:

Suppress default anchor fragments for regular and split branch flows.

hjotha:
Part of #332.

Fixes #320.

## Summary

`DESCRIBE MICROFLOW` now omits anchor fragments when all emitted sides match the MDL defaults.

## Root cause

Anchor formatting printed every known `from`/`to` side even when the side matched the statement default. These explicit defaults were semantically redundant and created cosmetic roundtrip drift.

## Fix

Suppress default fragments for regular flows and split branches. Non-default sides are still emitted.

This intentionally changes describe output for flows that only used default sides.

## Tests

Added regression coverage for omitted defaults and non-default anchor preservation.

## Validation

- [x] `make build`
- [x] `make test`
- [x] `make lint-go`
- [x] `make test-integration`

## Agentic Code Testing

- [x] Synthetic regression coverage verifies default anchors are omitted while non-default anchors remain visible.

## Test plan

- [x] Regular default anchors are omitted.
- [x] Split branch default anchors are omitted.
- [x] Non-default anchors are still emitted.

### DIFF
--- a/mdl-examples/bug-tests/320-describer-suppress-default-anchor-fragments.mdl
+++ b/mdl-examples/bug-tests/320-describer-suppress-default-anchor-fragments.mdl
@@ -1,0 +1,55 @@
+-- ============================================================================
+-- Bug #320: DESCRIBE emitted default anchor fragments redundantly
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   `describe microflow` printed `@anchor` lines for every activity, even
+--   when every from/to side matched the MDL default for that flow shape:
+--     - regular sequence flows default to `from: right, to: left`
+--     - true branches of IF default to `from: right, to: left`
+--     - false branches of IF default to `from: bottom, to: top`
+--   Output looked like
+--     log info node 'X' 'a';
+--     @anchor(from: right, to: left)         -- noise: this is the default
+--     log info node 'X' 'b';
+--   producing verbose, non-author-friendly DESCRIBE output and noisy
+--   `mxcli diff-local` runs after a roundtrip.
+--
+-- After fix:
+--   `emitAnchorAnnotation` and `emitSplitAnchorAnnotation` (in
+--   cmd_microflows_show_helpers.go) suppress fragments that match the
+--   default and skip the entire annotation when all sides are defaults.
+--   Non-default sides (e.g. `to: top` on a return) are still emitted.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/320-describer-suppress-default-anchor-fragments.mdl -p app.mpr
+--
+--   mxcli -p app.mpr -c "describe microflow BugTest320.MF_LinearDefaults"
+--   The output must contain NO `@anchor` line — every flow uses defaults.
+--
+--   mxcli -p app.mpr -c "describe microflow BugTest320.MF_NonDefaultReturn"
+--   The output must keep `@anchor(to: top)` on the return statement.
+-- ============================================================================
+
+create module BugTest320;
+
+-- Linear flow, all defaults — describe must emit zero @anchor lines.
+create microflow BugTest320.MF_LinearDefaults ()
+begin
+  log info node 'BugTest320' 'a';
+  log info node 'BugTest320' 'b';
+  log info node 'BugTest320' 'c';
+end;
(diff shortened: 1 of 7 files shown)


ako (commented):
## Review: fix: suppress default anchor fragments in describe

**Summary**: Correct fix for noisy describe output. Suppressing `@anchor(from: right, to: left)` on regular flows and the equivalent default combinations on split branches eliminates cosmetic BSON drift. The approach is sound; a few design points worth discussing.

### Hard-coded string literals instead of constants/keywords

`emitAnchorAnnotation` now does:
```go
if from != "" && from != "right" {
if to   != "" && to   != "left" {
```

The side names are computed by `anchorSideKeyword(AnchorRight)` / `anchorSideKeyword(AnchorLeft)` earlier in the same function. Using the literal strings here creates a hidden coupling — if the keyword mapping ever changed, the suppression would silently stop working. Expressing the defaults as `anchorSideKeyword(AnchorRight)` and `anchorSideKeyword(AnchorLeft)` would make the link explicit and survive a keyword rename without a bug.

The same concern applies to the string slices passed to `branchAnchorFragmentWithDefaultSides` (`[]string{"right", "bottom"}` etc.).

### Secondary suppression logic in `branchAnchorFragmentWithDefaultSides`

After primary default-side suppression, there's a second pass:
```go
case "false":
    if to == "" && from == "top" { from = "" }
    if from == "" && (to == "bottom" || to == "right") { to = "" }
case "true":
    if from == "" && to == "bottom" { to = "" }
```

These suppress side combinations observed to be "layout-equivalent" in Studio Pro. This logic is hard to reason about without knowing the layout rules, and there's no comment explaining *why* these combinations are equivalent (e.g., "Studio Pro auto-routes the false branch bottom-right when the destination is directly below"). The test `TestEmitSplitAnchor_OmitsSingleSidedLayoutEquivalentBranchAnchors` covers the cases but doesn't explain the reasoning. A short comment with the layout rule would help a future maintainer judge whether a new case belongs here.

Also: the secondary suppression only fires when one side was already primary-suppressed, making the logic order-dependent. Worth a comment noting that the input to the switch block is already post-primary-suppression.

### Existing test updates

Several tests were updated by swapping default anchor values (e.g. `AnchorRight/AnchorLeft` → `AnchorTop/AnchorTop`) so they continue to emit something after the suppression. This is the right mechanical fix. One note: the old tests were implicitly verifying that *default* anchors were emitted; now they verify that *non-default* anchors are emitted. The suppression behaviour itself is tested by the new `OmitsDefault` tests, so coverage is complete — just worth being aware that the semantic intent of the old tests shifted.

### `containsString` helper

A small linear-search helper over a ≤2-element slice. Fine for the use case; could alternatively be expressed inline with `||" at the call site to avoid an extra function. Not a blocker.

### Tests

Good coverage: default regular flow omitted, split default branches omitted, builder no-else defaults omitted, single-sided layout-equivalent combinations omitted, non-default destination still emitted. The concurrent race test update is correct.

### Verdict

Approve with a recommendation to replace the hard-coded default-side strings with `anchorSideKeyword(AnchorRight/Left)` expressions, and to add a comment explaining the layout semantics behind the secondary suppression switch. Both are readability improvements, not correctness issues.

ako:
@hjotha can you resolve the merge conflicts in this PR?

hjotha:
@ako yes fixes are on the way.

hjotha:

- Replaced hard-coded default anchor side strings with values derived through `anchorSideKeyword`.
- Added comments documenting the post-primary-suppression layout-equivalent branch-anchor cases.

```bash
make build
make test
make lint-go
```

PR #323, opened 2026-04-26
fix: preserve multiline source expression whitespace

### LINKED ISSUE #322: Multiline microflow expressions lose source whitespace during parse/exec
Problem:

Multiline declaration and log-template expressions can be rebuilt from parsed expression nodes, which normalizes source whitespace and collapses line breaks. Log template parameter gaps can also lose newline-only whitespace between parameters.

Expected:

For source-sensitive multiline expressions, the visitor should retain original expression text and the executor should serialize it back.

Impact:

Roundtrip output loses intentional user layout and generates cosmetic mismatch noise.

Proposed fix:

Introduce a small source-preserving expression wrapper and use it for declare/log expression positions that need source retention.

hjotha:
Part of #332.

Fixes #322.

## Summary

Preserves multiline source text for declaration initial values and log expressions, including newline-only whitespace between log template parameters. Also avoids preserving plain string literals solely because their text contains punctuation such as `!` or `:`.

## Root cause

The visitor rebuilt declare/log expressions from parsed expression nodes. That discarded original line breaks and inter-parameter whitespace. The source-preservation scanner also inspected operator-like bytes inside single-quoted string literals, which could wrap plain template text as a `SourceExpr`.

## Fix

- Add `SourceExpr` as a source-preserving expression wrapper.
- Use `buildSourceExpression` for source-sensitive declare/log/while expressions.
- Serialize `SourceExpr` through executor expression formatting.
- Preserve newline-only trailing whitespace between log template parameters.
- Skip compact-operator detection while scanning inside single-quoted literals, including doubled quote escapes.

## Tests

Added parser coverage for synthetic multiline declare/log template expressions and visitor coverage for punctuation inside string literals. Test names and fixtures are synthetic and do not reference real project microflows.

## Validation

- [x] `go test ./mdl/visitor -run 'TestDeclareAndLogTemplatePreserveMultilineSourceWhitespace|TestShouldPreserveExpressionSourceIgnoresStringLiteralPunctuation'`
- [x] `make build`
- [x] `make test`
- [x] `make lint-go`
- [x] `make test-integration`

## Agentic Code Testing

- [x] Synthetic multiline-expression tests verify source text remains stable for generated MDL.
- [x] Synthetic string-literal tests verify punctuation inside strings does not force source preservation.

## Test plan

- [x] Multiline declaration initial value keeps original source.
- [x] Log template parameter keeps trailing blank-line whitespace.
- [x] String literal punctuation does not create a source-preserved expression.

### DIFF
--- a/mdl-examples/bug-tests/322-multiline-source-expression-whitespace.mdl
+++ b/mdl-examples/bug-tests/322-multiline-source-expression-whitespace.mdl
@@ -1,0 +1,61 @@
+-- ============================================================================
+-- Bug #322: Multiline source expression whitespace lost on roundtrip
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   The visitor rebuilt expression strings from parsed expression nodes,
+--   so original line breaks and whitespace inside DECLARE initial values
+--   and LOG template parameters were normalized away. A multiline
+--   `declare $X string = ...` statement that authors had carefully
+--   formatted across several lines came back as a single very long line
+--   after describe → exec → describe. Inter-parameter blank lines in
+--   `LOG ... WITH (...)` were similarly collapsed, making real-world
+--   microflows non-fixpoint and producing noisy `mxcli diff-local` output.
+--
+-- After fix:
+--   Added a `SourceExpr` AST node that wraps an expression with its
+--   original source text. The visitor uses `buildSourceExpression` for
+--   source-sensitive declare/log/while expressions, and the executor
+--   serializes `SourceExpr` through to MDL output so the original
+--   whitespace and line breaks survive every roundtrip.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/322-multiline-source-expression-whitespace.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest322.MF_MultilineDeclare"
+--   mxcli -p app.mpr -c "describe microflow BugTest322.MF_MultilineLogTemplate"
+--   The describe outputs must keep the multiline shape and must be
+--   fixpoints under describe → exec → describe.
+-- ============================================================================
+
+create module BugTest322;
+
+-- DECLARE initial value spread across several lines with leading `+`. The
+-- describer must preserve the line breaks rather than collapse them onto
+-- one line.
+create microflow BugTest322.MF_MultilineDeclare (
+  $Page: integer,
+  $Token: string
+)
+returns string as $Endpoint
+begin
+  declare $Endpoint string = '/api/v1'
++ '/items?page=' + toString($Page)
++ '&token=' + $Token;
+
+  return $Endpoint;
+end;
+/
+
+-- LOG template parameters separated by a blank line. The newline-only
+-- whitespace between `{1} = toString($Count)` and the next parameter must
+-- survive the roundtrip.
+create microflow BugTest322.MF_MultilineLogTemplate (
+  $Count: integer,
+  $Endpoint: string
+)
+begin
+  log info node 'BugTest322' 'Processed {1} items for {2}' with ({1} = toString($Count)
+
+, {2} = $Endpoint);
+end;
+/
--- a/mdl/ast/ast_expression.go
+++ b/mdl/ast/ast_expression.go
@@ -125,6 +125,15 @@ type IfThenElseExpr struct {
 
 func (e *IfThenElseExpr) isExpression() {}
 
+// SourceExpr preserves original expression source text while keeping the parsed
+// expression tree available for callers that need semantic inspection.
+type SourceExpr struct {
+	Expression Expression
+	Source     string
+}
+
+func (e *SourceExpr) isExpression() {}
+
 // ============================================================================
 // XPath-Specific Expression Types
 // ============================================================================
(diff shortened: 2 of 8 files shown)


hjotha:
Updated the branch with a source-preservation guard for string literals: punctuation inside single-quoted literals no longer forces `SourceExpr` preservation. Validation after the update:

- `go test ./mdl/visitor -run 'TestDeclareAndLogTemplatePreserveMultilineSourceWhitespace|TestShouldPreserveExpressionSourceIgnoresStringLiteralPunctuation'`
- `make build`
- `make lint-go`
- `make test`

ako (commented):
## Review: fix: preserve multiline source expression whitespace

**Summary**: Solid approach — wrapping parsed expressions in a `SourceExpr` that preserves the original text is the right design for roundtrip fidelity. The core implementation is correct. Several issues worth addressing.

### Bug in `resolveAssociationPaths` for `SourceExpr`

```go
case *ast.SourceExpr:
    if e.Source != "" {
        return e
    }
    return &ast.SourceExpr{
        Expression: fb.resolveAssociationPaths(e.Expression),
        Source:     e.Source,  // e.Source is "" here
    }
```

The second branch creates a `SourceExpr{Source: ""}` wrapping a resolved inner expression. `buildSourceExpression` only creates a `SourceExpr` when `source != ""`, so this branch is effectively unreachable in practice — but if it were reached, the empty-source wrapper would be silently peeled off by `expressionToString`'s `e.Source != ""` guard. Simplify to just `return fb.resolveAssociationPaths(e.Expression)` for the empty case.

More importantly: when `Source != ""` the inner `e.Expression` is returned without recursing into it for association path resolution. This means any association paths inside a source-preserved expression are never resolved. Whether this is intentional (the source text is already in final form) or an oversight depends on how association path resolution interacts with source-preserved expressions — worth a comment clarifying the intent.

### `retrieveInterClauseWhitespaceSuffix` hard-codes 4-space indent

```go
for _, suffix := range []string{"\r\n    ", "\n    "} {
    if strings.HasSuffix(gap, suffix) {
        return strings.TrimSuffix(gap, suffix)
    }
}
```

This strips exactly 4 trailing spaces from the inter-clause gap, assuming the MDL formatter will re-add 4 spaces of indentation at output time. If a user writes the LIMIT/OFFSET clause with a different indent (tabs, 2 spaces, or 8 spaces), the suffix check fails and the function returns `""`, silently dropping any trailing whitespace. A more robust approach would strip trailing whitespace dynamically rather than matching fixed suffixes.

### `shouldPreserveExpressionSource` — `not(` case lacks explanation

```go
if strings.Contains(strings.ToLower(source), "not(") {
    return true
}
```

This could false-positive on identifiers or attribute names containing the substring `not(` (unlikely but possible). More importantly, there's no comment explaining why `not(` specifically requires source preservation. Adding one (e.g., "reconstructed `not(x)` would normalise to `not (x)` — preserve compact form") would help future readers.

### `buildSetStatement` calls `buildExpression` twice

The pattern calls `buildExpression(expr)` for list/aggregate detection, then overwrites `valueExpr` at the end with `buildSourceExpression(valueExprCtx)`. This calls `buildExpression` a second time internally. The approach is clear and correct, but the double-parse is worth a brief comment noting that the first parse is for semantic detection and the second is for source preservation.

### Expression traversal coverage

The PR adds `SourceExpr` handling to `expressionToString`, `expressionToXPath`, `resolveAssociationPaths`, and `unwrapParenCall`. Other expression traversal switch statements in `cmd_microflows_helpers.go` (e.g. the type inference / XPath conversion switches) do not yet handle `SourceExpr`. Since `SourceExpr` now wraps almost all parsed expressions, any unhandled switch falls through to `default` (usually returning `""`). Worth auditing whether `inferExpressionType` or other callers that switch on expression types need a `*ast.SourceExpr` case that unwraps to the inner expression.

### `appendLogNodeTrailingWhitespace` comment is fragile

```go
// formatAction always writes one space between the node and message slots.
// Preserve the source line break but avoid duplicating indentation spaces.
trailing = strings.TrimRight(trailing, " \t")
```

This is a coupling to the formatter's output format. If `formatAction` changes its spacing, this trim logic would either over-strip or under-strip. A comment linking to the specific formatter call would help future maintenance.

### Tests

Good coverage across the key cases: multiline declare, multiline log template, call/change/show-message argument slots, retrieve LIMIT/OFFSET, string-literal punctuation not triggering preservation. The `TestActionExpressionSlotsPreserveSourceWhitespace` and `TestTrailingExpressionWhitespacePreservedForRoundtripSlots` tests are comprehensive. The `TestShouldPreserveExpressionSourceIgnoresStringLiteralPunctuation` directly tests the heuristic including the doubled-quote escape case.

### Verdict

Approve with the `resolveAssociationPaths` bug flagged as the highest-priority fix, and the hard-coded 4-space suffix in `retrieveInterClauseWhitespaceSuffix` as worth addressing. The core `SourceExpr` design is clean and correct.

hjotha:

- Simplified empty-source `SourceExpr` handling in association-path resolution.
- Made retrieve inter-clause whitespace preservation robust to indentation changes.
- Added comments around compact `not(...)` preservation and the formatter-coupled log-node whitespace handling.

```bash
make build
make test
make lint-go
```

PR #325, opened 2026-04-26
refactor: extract microflow anchor selection helpers

### LINKED ISSUE #324: Microflow builder anchor precedence is duplicated across flow construction paths
Problem:

The builder has duplicated pending-anchor and branch-destination selection logic in multiple flow construction paths. That makes follow-up bug fixes harder to review because each fix also has to touch duplicated control-flow plumbing.

Expected:

Shared anchor selection logic should live in small helpers, with no behavior change.

Impact:

The current duplication increases review risk for future anchor and branch-handler fixes.

Proposed fix:

Extract helpers for branch destination anchor selection and pending-flow anchor precedence.

hjotha:
Part of #332.

Closes #324.

## Summary

Extracts duplicated microflow builder anchor selection logic into two helpers. This is a no-behavior-change refactor intended to make follow-up anchor/handler fixes smaller and easier to review.

## Root cause

Several builder paths repeated the same precedence checks inline for pending anchors, branch anchors, and statement anchors.

## Refactor

- Add `branchDestinationAnchor`.
- Add `pendingFlowAnchors`.
- Route existing IF and graph construction call sites through the helpers.

## Behavior

No behavior change intended. Behavior-changing fixes remain in separate branches.

## Tests

No new behavior tests added because this only extracts existing logic. Existing tests cover the affected paths.

## Validation

- [x] `make build`
- [x] `make test`
- [x] `make lint-go`
- [x] `make test-integration`

## Agentic Code Testing

- [x] No user-facing syntax or behavior change.
- [x] Existing test suite covers the refactored builder paths.

## Test plan

- [x] Full test suite remains green after refactor.

### DIFF
--- a/mdl/executor/cmd_microflows_builder_control.go
+++ b/mdl/executor/cmd_microflows_builder_control.go
@@ -125,10 +125,7 @@ func (fb *flowBuilder) addIfStatement(s *ast.IfStmt) model.ID {
 					// Destination: prefer the first statement's own @anchor(to: ...) if it
 					// has one; otherwise fall back to trueBranchAnchor.To.
 					flow := newHorizontalFlowWithCase(splitID, actID, "true")
-					applyUserAnchors(flow, trueBranchAnchor, trueBranchAnchor)
-					if thisAnchor != nil && thisAnchor.To != ast.AnchorSideUnset {
-						flow.DestinationConnectionIndex = int(thisAnchor.To)
-					}
+					applyUserAnchors(flow, trueBranchAnchor, branchDestinationAnchor(trueBranchAnchor, thisAnchor))
 					fb.flows = append(fb.flows, flow)
 				} else {
 					flow := newHorizontalFlow(lastThenID, actID)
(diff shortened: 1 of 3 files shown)


ako (commented):
## Review: refactor: extract microflow anchor selection helpers

**Summary**: Clean, no-behavior-change refactor. The semantic equivalence holds at all three call sites.

### Equivalence verification

**`branchDestinationAnchor`** replaces:
```go
applyUserAnchors(flow, branchAnchor, branchAnchor)
if thisAnchor != nil && thisAnchor.To != ast.AnchorSideUnset {
    flow.DestinationConnectionIndex = int(thisAnchor.To)
}
```
The helper returns `stmtAnchor` when it has a non-unset `To`, otherwise falls back to `branchAnchor` — correctly preserving the statement-level override precedence. ✓

**`pendingFlowAnchors`** replaces the three-variable assignment in `buildFlowGraph`. When `pendingAnchor == nil` the return is `(previousAnchor, stmtAnchor)` — same as the original initialisation. When non-nil the return is `(pendingAnchor, branchDestinationAnchor(pendingAnchor, stmtAnchor))`, which maps to the same two branches as the original `destAnchor` override. ✓

The `pendingFlowAnchor = nil` reset moves from inside the `if pendingFlowAnchor != nil` block to unconditional — equivalent since nil-to-nil is a no-op. ✓

### Minor naming note

`pendingFlowAnchors` is only called in one place and its name reads more like a state accessor than a resolution function. `resolveFlowAnchors` or `selectFlowAnchors` would make the intent slightly clearer. Not a blocker.

### Verdict

Approve. The refactor is correct and the resulting call sites are noticeably cleaner.

PR #327, opened 2026-04-27
fix: pair microflow splits with nearest merge

### LINKED ISSUE #326: Describer can pair a split with a downstream merge instead of the nearest merge
## Problem

When a microflow has sequential branch structures, the describer can pair an `ExclusiveSplit` with a downstream `ExclusiveMerge` instead of the merge where that split's branches actually reconverge.

That can make the generated MDL nest the first split's continuation inside the branch body. A describe/exec/describe cycle can then change the microflow structure even when the original graph was valid.

## Expected behavior

Split/merge pairing should choose the nearest common merge reachable from all normal branch flows, not an arbitrary common downstream merge.

## Notes

The bug is caused by selecting a common reachable merge from map iteration. The fix ranks common merge candidates by branch distance and uses deterministic tie-breaking.

hjotha:
Part of #332.

## Summary

- pair exclusive splits with the nearest common exclusive merge
- ignore error-handler flows when computing structural split/merge pairing
- add regression tests for sequential if-without-else continuation placement

Closes #326

## Validation

- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/mdl-examples/bug-tests/326-describer-pair-split-with-nearest-merge.mdl
+++ b/mdl-examples/bug-tests/326-describer-pair-split-with-nearest-merge.mdl
@@ -1,0 +1,55 @@
+-- ============================================================================
+-- Bug #326: Describer paired splits with downstream merge instead of nearest
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   When a microflow had two sequential `if ... end if;` blocks, the
+--   describer could pair the FIRST split with the SECOND merge, nesting
+--   the continuation between the two ifs (and even the second if) inside
+--   the first if's body. A describe → exec → describe cycle then mutated
+--   the structure even though the original graph was valid:
+--     -- before fix, after first describe:
+--     if $Logo != empty then
+--       log info node 'App' 'logo';
+--       log info node 'App' 'after logo';   -- wrong: was outside first if
+--       if $Cover != empty then              -- wrong: was sibling, not nested
+--         log info node 'App' 'cover';
+--       end if;
+--     end if;
+--
+-- Root cause:
+--   Split/merge pairing picked any common downstream merge reachable from
+--   both branches, not the nearest one. Error-handler flows were included
+--   in the search, biasing it further.
+--
+-- After fix:
+--   `findMergeForSplit` now uses shortest-path distances per branch and
+--   selects the nearest common merge, ignoring error-handler flows for
+--   structural pairing.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/326-describer-pair-split-with-nearest-merge.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest326.MF_SequentialIfs"
+--   The describe output must keep the two `if ... end if;` blocks as
+--   siblings (with the `log` between them at top level), and the
+--   describe → exec → describe cycle must be a fixpoint.
+-- ============================================================================
+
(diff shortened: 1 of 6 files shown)


hjotha:
Follow-up pushed in `75a3788`.

This extends guard-branch detection from direct EndEvent checks to recursive terminality before the split merge boundary. It preserves the guard form when a terminating branch contains actions or nested splits, instead of wrapping the continuation in a synthetic/empty `else`.

- `make build`
- `make test`
- `make lint-go`
- `go test ./mdl/executor -run 'TestTraverseFlow_(GuardBranchWithMultipleActivitiesKeepsContinuationOutsideElse|NestedTerminalGuardBranchSuppressesEmptyOuterElse|SequentialIfWithoutElseKeepsContinuationOutsideFirstIf|IfElse)$' -count=1`

ako (commented):
## Review: fix: pair microflow splits with nearest merge

**Summary**: Correct fix for a real bug. The two root causes (wrong merge pairing and shallow guard detection) are addressed cleanly with the right algorithmic tools. Good test coverage across the key structural cases.

###  — nearest-merge selection

Replacing the "any common merge" set intersection with BFS shortest-path distances is the right approach.  sorts by max-distance first (nearest merge relative to the farthest branch), then sum-distance, then ID for determinism. Map iteration over  is non-deterministic but the sort at the end makes the final selection deterministic. ✓

The BFS in  uses a distance-comparison guard instead of a simple visited set:
```go
if previous, ok := distances[item.id]; ok && previous <= item.distance {
    continue
}
```
With uniform edge weights this guard is never triggered (BFS naturally visits nodes in shortest-path order), so a plain  check would be equivalent. No behavior difference — just worth noting for future readers.

### Guard detection — 

The recursive DFS correctly handles multi-activity terminating branches that the old single-step  check missed. The  on every recursive branch call avoids cross-branch contamination but allocates O(visited) memory per branch per depth level. For typical microflows (< 100 activities) this is negligible; worth a comment for anyone who hits a pathological deeply-nested case.

###  with nil 

The nested-split case:
```go
nestedMergeID := splitMergeMap[currentID]  // nil map read → returns ""
```
Reading from a nil map in Go is safe (returns zero value), so passing  for  doesn't panic — it results in every nested split being traversed with , which means the termination check continues past any merge. The test  relies on this implicitly: it passes  for  and the test graph happens to work because both inner branches are EndEvents. This is fragile — a test graph with an actual merge after the inner split would silently produce wrong results. A nil guard and a comment would make the contract explicit.

### Inconsistent test setup pattern

 pre-builds  using .  passes . The inconsistency is confusing — it looks like an oversight rather than a deliberate choice. The nested guard test should build its own  to match production behavior.

### Minor:  called in  but old  is still present

 is no longer called by  but remains in the file. If it has no other callers it should be removed to avoid dead code.

```bash
grep -r 'collectReachableNodes' mdl/executor/
```

### Bug test script

 covers the exact symptom from the issue and documents the before/after clearly. ✓

### Verdict

Approve. The fix is correct and addresses both root causes. The  test inconsistency and the possible  dead code are the two cleanup items worth addressing before or after merge.

hjotha:

- Removed the dead `collectReachableNodes` helper.
- Cleaned up the nearest-merge/guard tests so they build the same split-merge map shape as production traversal.
- Kept the nested-split fallthrough guard explicit in the current implementation/tests.

```bash
make build
make test
make lint-go
```

PR #329, opened 2026-04-27
fix: preserve nested loop body flow traversal

### LINKED ISSUE #328: DESCRIBE MICROFLOW can omit loop-body continuation flows stored in the parent graph
## Problem

Some microflow graphs store loop boundary flows in the loop's local object collection while storing real body continuation flows in the parent graph.

When a loop-local origin already exists, the describer can ignore parent-level flows with the same origin. That can omit activities from the described loop body, especially for nested loops where the continuation is represented outside the nested loop collection.

## Expected behavior

Loop body traversal should merge relevant parent-level body flows into the local loop graph, including nested loop descendants, so describe preserves the full body order.

## Notes

The fix should keep local loop boundary flows, but add missing parent-level continuation flows for origins that belong to the loop body.

hjotha:
Part of #332.

## Summary

- merge parent-level body flows into loop-local traversal maps
- collect nested loop body object IDs recursively
- add a regression for local loop-boundary flows plus parent-level continuations sharing origins

Closes #328

## Validation

- `go test ./mdl/executor -run TestTraverseFlow_LoopBodyMergesParentFlowsForExistingOrigin`
- `make build`
- `make lint-go`
- `make test`


ako (commented):
## Review: fix: preserve nested loop body flow traversal

**Summary**: Correct, well-targeted fix. The root cause (short-circuiting parent-flow merging when a local loop-boundary flow already existed for the same origin) is directly addressed, and the recursive descent via `collectLoopObjectIDs` correctly handles the nested-loop case the old `loopActivityMap` lookup missed.

### Core fix correctness

Old guard:
```go
if _, inLoop := loopActivityMap[originID]; inLoop {
    if _, exists := loopFlowsByOrigin[originID]; !exists {
        loopFlowsByOrigin[originID] = flows
    }
}
```

The `!exists` short-circuit silently dropped parent-level continuation flows for any origin that already had a local loop-boundary flow. The new code merges both using `appendSequenceFlowIfMissing`. ✓

The same fix is correctly applied to the `flowsByDest` side too. ✓

### `collectLoopObjectIDs` — recursive descent is necessary

The old code used `loopActivityMap` (immediate loop objects only). The new `loopObjectIDs` descends into nested `LoopedActivity.ObjectCollection` objects, which is needed so parent-level flows for nested loop body activities are also merged. The recursion is straightforward and handles nil ObjectCollections. ✓

### `preferLoopBodyStart` — deterministic start selection

Replacing the previous non-deterministic `break` with a position-based tiebreaker is a good improvement. One edge: if two activities have exactly the same X/Y position AND both have zero incoming flows, the winner still depends on map iteration order. This is an acceptable edge case (same-position activities are a user authoring issue), but a secondary tiebreaker on ID string would make it fully deterministic.

### `sequenceFlowIdentity` — composite key gap

```go
return fmt.Sprintf("%s>%s:%t:%d:%d", flow.OriginID, flow.DestinationID, flow.IsErrorHandler, flow.OriginConnectionIndex, flow.DestinationConnectionIndex)
```

The composite fallback key (used when `flow.ID == ""`) does not include `CaseValue`. If a split emits two flows from the same origin to the same destination with different case values but without IDs (possible in test-constructed graphs), one would be incorrectly deduplicated. In production data every flow has a UUID so this only affects test helpers — but since test flows are constructed without IDs in this very file, it's worth noting. Adding `:%v` for `CaseValue` would close the gap.

### Test coverage

The test graph is well-designed — adversarial storage order (nested loop listed first), local boundary flow on `fetch` to trigger the original bug, and parent-level continuation flow from `fetch` to `nested`. The ordering assertions (setup before fetch before nested loop) catch the regression.

One gap: the test verifies that `nested` and `nested-tail` appear in the output, but not their relative order inside the nested loop body. A position check similar to the outer body assertions would catch a future inner-body ordering regression.

### Bug test script

Covers a realistic nested-loop microflow with outer and inner body activities. The describe → exec → describe fixpoint requirement is documented in the header. ✓

### Verdict

Approve. The fix is correct and the recursive descent approach is the right solution. The two minor items (same-position tiebreaker and `CaseValue` in composite key) are low-priority but easy wins.

hjotha:

- Included `CaseValue` in the synthetic sequence-flow identity fallback used by tests/ID-less flows.
- Added deterministic loop-body start tie-breaking by ID for equal positions.

```bash
make build
make test
make lint-go
```

PR #331, opened 2026-04-27
fix: preserve annotations attached inside loop bodies

### LINKED ISSUE #330: Loop-body annotations can be lost during microflow describe/exec
## Problem

Annotations attached to statements inside loop or while bodies can disappear after a microflow describe/exec roundtrip.

The builder copies sequence flows from nested loop builders back to the parent graph, but did not copy nested annotation flows. The describer also collected annotation captions only from the top-level object collection.

## Expected behavior

Annotations attached to activities inside loop bodies should remain attached after roundtrip, including annotations on nested branch statements.

## Notes

The fix promotes nested loop annotation flows to the parent graph and recursively collects annotation captions from nested loop object collections.

hjotha:
Part of #332.

## Summary

- copy loop/while body annotation flows into the parent microflow graph
- collect annotation captions recursively from nested loop object collections
- merge loop-local annotation flows into loop body traversal
- add regressions for annotated nested decisions and annotation flows stored inside loop object collections

Closes #330

## Validation

- `go test ./mdl/executor -run 'TestLoopBodyIfAnnotationPromotedToParentFlows|TestTraverseFlow_LoopBodyUsesNestedAnnotationFlows'`
- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/cmd/mxcli/tui/watcher.go
+++ b/cmd/mxcli/tui/watcher.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -78,6 +79,7 @@ func newWatcher(mprPath, contentsDir string, sender MsgSender) (*Watcher, error)
 
 func (w *Watcher) run(sender MsgSender) {
 	var debounceTimer *time.Timer
+	var debounceSeq atomic.Uint64
 
 	for {
 		select {
@@ -110,7 +112,11 @@ func (w *Watcher) run(sender MsgSender) {
 			if debounceTimer != nil {
 				debounceTimer.Stop()
 			}
+			seq := debounceSeq.Add(1)
 			debounceTimer = time.AfterFunc(watchDebounce, func() {
+				if debounceSeq.Load() != seq {
+					return
+				}
 				sender.Send(MprChangedMsg{})
 			})
 
--- a/cmd/mxcli/tui/watcher_test.go
+++ b/cmd/mxcli/tui/watcher_test.go
@@ -35,10 +35,11 @@ func TestWatcherDebounce(t *testing.T) {
 	}
 	defer w.Close()
 
-	// Rapidly write 5 times — should debounce into a single message
+	// Rapidly write 5 times — should debounce into a single message.
+	// Keep the burst tighter than the debounce window so slow CI machines do
+	// not accidentally let an intermediate timer fire.
 	for i := range 5 {
 		_ = os.WriteFile(unitFile, []byte{byte('a' + i)}, 0644)
-		time.Sleep(50 * time.Millisecond)
 	}
 
 	// Wait for debounce to fire (500ms + margin)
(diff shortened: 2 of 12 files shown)


hjotha:
Updated the branch to also cover annotation flows stored directly inside loop object collections. Validation after the update:

- `go test ./mdl/executor -run 'TestLoopBodyIfAnnotationPromotedToParentFlows|TestTraverseFlow_LoopBodyUsesNestedAnnotationFlows'`
- `make build`
- `make lint-go`
- `make test`

ako (commented):
## Review: fix: preserve annotations attached inside loop bodies

**Summary**: Clean two-part fix. The builder side is a symmetric two-liner matching the existing sequence-flow promotion pattern; the describer side correctly handles the split between annotation objects (stored in loop ObjectCollections) and annotation flows (now promoted to the parent level).

### Builder fix (two lines)

```go
fb.annotationFlows = append(fb.annotationFlows, loopBuilder.annotationFlows...)
```

Applied symmetrically in both `addLoopStatement` and `addWhileStatement`, matching the existing `fb.flows = append(fb.flows, loopBuilder.flows...)` promotion. The final `MicroflowObjectCollection` already packages `fb.annotationFlows` into `AnnotationFlows`, so no further wiring is needed. ✓

### Describer fix

The object-vs-flow split is key: annotation Objects live inside the loop's ObjectCollection; annotation Flows are now promoted to the parent level by the builder. The describer fix handles both sides:

- **`collectAnnotationCaptions`**: Recursively descends into `LoopedActivity.ObjectCollection` to collect caption IDs from nested annotations. The `continue` after the Annotation case is redundant (a Go object can only be one type) but harmless.
- **`emitLoopBody`**: Merges parent `annotationsByTarget` with the loop-local annotation map. This handles the case where annotation flows are stored in the loop's own ObjectCollection (older MPR files that predate the builder fix).

Together these cover both the new MPR layout (flows promoted, captions found recursively) and existing MPR files (flows in loop ObjectCollection, found by the loop-local `buildAnnotationsByTarget`). ✓

### `mergeAnnotationsByTarget` — returns overlay by reference when base is empty

```go
if len(base) == 0 {
    return overlay
}
```

This returns the same map as `overlay`, not a copy. In `emitLoopBody` the overlay is a freshly built map from `buildAnnotationsByTarget`, so mutation of the result wouldn't affect anything else. Safe for the current caller. Worth a comment if this function gets additional callers in future.

### Test coverage

**`TestLoopBodyIfAnnotationPromotedToParentFlows`**: Verifies the builder path — an annotation on a nested IF inside a LOOP ends up in the parent object collection's annotation map. Correctly traces through the promoted flow + recursive caption collection. ✓

**`TestTraverseFlow_LoopBodyUsesNestedAnnotationFlows`**: Pre-builds `annotationsByTarget` from the loop's ObjectCollection and passes it directly to `traverseFlow`, verifying that `@annotation` appears in the rendered output. This exercises annotation rendering but not the `emitLoopBody` merge path itself (the merge logic is tested indirectly by the MDL bug-test script rather than a Go unit test).

A Go-level test that constructs a full `LoopedActivity` with nested annotations and exercises `emitLoopBody` end-to-end would close this gap, but the existing coverage is sufficient for the described bug.

### Bug test script

Documents the symptom (annotations on IF inside LOOP disappear after roundtrip) and the dependency on PR #327 for full fixpoint roundtrip. The scope note is helpful context. ✓

### Verdict

Approve. The fix is correct, minimal, and symmetric between the builder and describer sides.

hjotha:

- Documented the `mergeAnnotationsByTarget` aliasing contract for the empty-base fast path.
- Kept the existing builder/describer coverage and bug-test fixture unchanged because the review found the behavior correct.

```bash
make build
make test
make lint-go
```

hjotha:

Issue #332, opened 2026-04-27
Microflow roundtrip stabilization for large-project audit findings


hjotha:
## Summary

While running describe/exec/describe audits against large Mendix projects and validating round-tripped MPRs with `mx check`, I found a second batch of related microflow roundtrip gaps. Several small fixes are already open as focused PRs; the remaining items are tracked below and staged in draft PR #346 while they are extracted into reviewable changes.

Opening this as an umbrella issue, following the pattern from #257, so the remaining work has one tracker while each PR stays focused and reviewable.

## Already submitted focused PRs

- [ ] #305: drop empty else branch emitted by the microflow describer
- [ ] #307: skip activity-bound annotations before unsupported-action comments
- [ ] #309: preserve default commit rollback error handling
- [ ] #311: preserve branch anchor direction pairs when writing microflow flows
- [ ] #313: ignore excluded microflows during project validation
- [ ] #315: make executor timeout configurable for long-running project checks
- [ ] #317: avoid trailing newline after microflow end events
- [ ] #319: preserve free-standing microflow annotations
- [ ] #321: suppress default anchor fragments in microflow descriptions
- [ ] #323: preserve multiline source expressions without treating string punctuation as structure
- [ ] #325: extract branch anchor helper logic in the microflow builder
- [ ] #327: stop branch traversal at the nearest split merge
- [ ] #329: keep nested loop body flows inside their owning branch
- [ ] #331: preserve loop body annotations during roundtrip
- [ ] #333: support microflow download file statements
- [ ] #334: support legacy microflow call web service statements
- [ ] #336: support ellipsis placeholder expressions
- [ ] #337: preserve duplicate implicit output aliases
- [ ] #338: write valid show-page parameter mappings
- [ ] #340: preserve validity for empty change-object actions
- [ ] #342: preserve Studio Pro-compatible microflow-call BSON field order
- [ ] #344: preserve attribute-based find/filter list operations

## Staging PR

- [ ] #346: draft staging branch for remaining local all-fixes work. This is not intended to merge as-is; it exists so the remaining behavior is visible while focused extraction continues.

## Remaining focused issues

- [ ] #347 / #364: preserve enum split microflow roundtrips
- [ ] #348 / #365: preserve inheritance split and cast-action microflow roundtrips
- [ ] #349 / #366: preserve custom error-handler routing during microflow roundtrip
- [ ] #345 / #363: preserve refresh-in-client on change-object actions with member assignments
[...]

PR #333, opened 2026-04-27
feat: support microflow download file statement

hjotha:
Part of #332.

## Summary

Adds MDL support for Studio Pro `Microflows$DownloadFileAction` using:

```mdl
download file $GeneratedReport;
download file $GeneratedReport show in browser;
```

The statement preserves existing download-file actions during describe/exec/describe and makes the action authorable when the file document variable already exists.

## Root Cause

`DownloadFileAction` was recognized by the MPR parser, but MDL had no AST node, grammar rule, visitor builder, describer output, graph builder, or writer support for it. Existing actions therefore could not round-trip as structured MDL.

## Changes

- Adds `DownloadFileStmt` to the microflow AST.
- Adds `download file $Var [show in browser] [on error ...]` grammar and visitor support.
- Formats/describes download-file actions as structured MDL.
- Builds and writes real `Microflows$DownloadFileAction` objects with `ErrorHandlingType`, `FileDocumentVariableName`, and `ShowInBrowser` preserved.
- Adds proposal docs, quick-reference entry, doctype fixture, and write-microflows skill guidance.

## Validation

- [x] `make build`
- [x] `go test ./mdl/visitor ./mdl/executor ./sdk/mpr -run 'TestDownloadFile|TestFormatAction_DownloadFile|TestBuildFlowGraph_DownloadFileCreatesRealAction'`
- [x] `./bin/mxcli check mdl-examples/doctype-tests/download_file.test.mdl`
- [x] `make lint-go`
- [x] `make test`
- [x] `make test-integration`

## Agentic Code Testing

- [x] New syntax is documented in a proposal, quick reference, and synthetic doctype fixture.
- [x] Tests use synthetic module/microflow names only.
- [x] Roundtrip coverage verifies both MDL parsing/building and MPR serialization/parsing.


ako (commented):
## Review: feat: support microflow download file statement

**Summary**: Well-executed, complete feature. Every pipeline layer is updated consistently. The grammar design is clean and the full-stack coverage is thorough.

### Full-stack coverage

All layers verified against the PR checklist:

- Grammar (, ): new  rule, / tokens ✓
- Parser regenerated: generated files committed ✓
- AST:  with , , ,  ✓
- Visitor: , wired into  and  ✓
- Builder: , wired into  ✓
- Builder annotations:  case ✓
- Builder validate:  case ✓
- Validate:  case ✓
- Validate microflow:  case ✓
- Format action:  case ✓
- Error handling type:  case ✓
- SDK model:  field added to  ✓
- MPR parser:  parsed with  default ✓
- MPR writer: BSON serialization with  for  ✓
- Formatter keyword list: ,  added ✓
- LSP completions: both keywords added ✓
- Documentation: quick reference, skill, proposal, doctype fixture ✓

### Keyword reservation

 and  are both added to the  rule, which means they appear in  and remain usable as attribute names, module names, and identifier contexts. This is the correct approach — no existing scripts using  or  as identifiers will break. ✓

### Error handling suffix

 doesn't add the  suffix itself, which is consistent with all other action formatters. The suffix is applied at the caller level via  →  → trimmed semicolon + appended suffix. With  updated for , error handling roundtrips correctly. ✓

###  — redundant  prefix check

```go
if fileDocument != "" && !strings.HasPrefix(fileDocument, "$") {
    fileDocument = "$" + fileDocument
}
```

 is stored without  (stripped in the visitor via ), so the  guard always fires. The check is defensive but misleading — a reader might wonder when the field would already have a . A comment clarifying that the field never stores the prefix would help, or just drop the condition: .

### Missing test case: 

 only tests . The false case (no  suffix, plain ) is exercised implicitly by the doctype fixture but has no dedicated unit test. Worth adding a second assertion in the same test function.

### BSON field key

The writer uses  and the parser reads from . This should be verified against the reflection data to confirm the Studio Pro storage name. The field struct tag is  (camelCase), which is different from the BSON key — confirming the BSON field name from an MPR sample would close the loop. If it was already established before this PR (the parser for  pre-existed), then this is fine.

### Tests

- Visitor test: both  and plain forms ✓
- Builder test:  and  set correctly ✓
- MPR roundtrip test: serialize → marshal → unmarshal → parse, all fields verified ✓
- Format action test:  case ✓ (missing  case — see above)

### Verdict

Approve. Complete and correct feature implementation. The two minor items (redundant  guard and missing  test) are low-priority polish.

PR #336, opened 2026-04-27
feat: support empty Java action arguments

hjotha:
## Summary

Uses the existing `empty` literal as the MDL representation for intentionally unbound Java-action parameters. This lets `describe` output preserve empty `BasicCodeActionParameterValue` bindings and lets `exec` write them back without inventing synthetic expression values or adding a new placeholder token.

Part of #332.

## Changes

- Maps Java action arguments written as `empty`/`null` back to empty code-action parameter values.
- Preserves empty microflow callback parameter references as empty `MicroflowParameterValue` references.
- Documents the syntax in the proposal index and proposal document.
- Adds a doctype example and parser/builder regression coverage.

## Validation

- `go test ./mdl/executor -run 'TestBuildJavaAction_Empty|TestBuildJavaAction' -count=1`
- `make build`
- `./bin/mxcli check mdl-examples/doctype-tests/empty_java_action_argument.mdl`
- `make test`


ako (commented):
## Why not `empty`?

The `...` token solves a real problem, but MDL already has the right tool for this: the `empty` keyword.

`empty` is already:
- A **lexer token** (`EMPTY: E M P T Y` in `MDLLexer.g4`)
- A **valid `literal`** in `MDLParser.g4` — parseable as an expression anywhere a literal is accepted, including call arguments
- A **`LiteralEmpty` AST kind** in `mdl/ast/ast_expression.go`
- **Already handled by the executor** — `expressionToString()` returns `"empty"` for `LiteralEmpty`, and `validate_microflow.go` already treats it as a null-like sentinel

So the builder could simply check `if lit.Kind == ast.LiteralEmpty` instead of `isPlaceholderExpression(...)`, and `describe` would emit:

```mdl
$Total = call java action SampleModule.Recalculate(
  CompanyId      = empty,
  RecalculateAll = true,
  ItemList       = empty
);
```

This is already valid MDL, already human-readable, and consistent with how `= empty` is used for uninitialized list variables (`declare $Items list of Module.Entity = empty`).

The `...` choice instead introduced a new lexer token, regenerated the entire ANTLR parser (3000+ lines of churn in `mdl_lexer.go` / `mdl_parser.go`), and added syntax with no precedent in MDL — all to solve a problem the existing keyword already covers. `CompanyId = ...` reads as broken/incomplete code; `CompanyId = empty` reads as intentional.

Suggested change: drop `ELLIPSIS` from the grammar, remove the `isPlaceholderExpression` helper, and map `LiteralEmpty` → `BasicCodeActionParameterValue { Argument: "" }` in `addCallJavaActionAction`. The doctype example, skill docs, and quick-reference entry would all stay — just s/`...`/`empty`/.

ako:
## PR #336 Review — feat: support empty Java action arguments

Good update. The `...` placeholder is gone; the existing `empty` literal is used correctly throughout. Parser, writer, formatter, and builder are all wired up, with test coverage at each layer. The doctype file uses the correct `.mdl` suffix.

One implementation detail worth documenting, and one missing test.

### Minor

**`isEmptyJavaActionArgument` checks both `LiteralEmpty` and `LiteralNull` without explanation**

In the microflow expression visitor both `empty` and `null` keywords produce `LiteralNull` — `LiteralEmpty` is only produced by the XPath visitor. So in practice only `LiteralNull` is ever seen here. The dual check in `isEmptyJavaActionArgument` is harmless but looks like defensive coding without context. A one-line comment — e.g. `// the microflow expression visitor maps both EMPTY and NULL tokens to LiteralNull` — would make this self-explanatory.

**No writer unit test for `MicroflowParameterValue` serialization**

`TestParseCodeActionParameterValue_MicroflowParameterValue` covers the parse side. `serializeCodeActionParameterValue` now has a `*microflows.MicroflowParameterValue` case but there is no corresponding writer test. The implementation is 5 straightforward lines, so the risk is low, but it would complete the test matrix alongside the existing parser test.

---

Otherwise looks good to merge.

PR #337, opened 2026-04-27
fix: reject duplicate microflow output variables

hjotha:
## Summary

Rejects duplicate microflow output variables instead of silently generating aliases.

If an authored MDL script declares the same output variable twice in the same scope, validation now reports a duplicate-variable error. If `describe` encounters an existing MPR that already contains duplicate output variables, it preserves the original output variable names and emits a warning comment instead of transforming the model to generated names such as `$Item_2`.

Part of #332.

## Changes

- Removes the silent variable-aliasing machinery for duplicate implicit outputs.
- Rejects duplicate output variables from create/retrieve/call/list/import/export/transform statements during validation.
- Emits `-- WARNING: duplicate output variable ... model is invalid` in describe output for invalid existing MPRs.
- Keeps shared variable-reference walker coverage for downstream builder passes.
- Removes aliasing proposal/docs/fixtures because the reviewed behavior is intentionally not supported.

## Validation

- `go test ./mdl/executor -run 'DuplicateOutput|StatementVarRefs' -count=1`
- `make build`
- `make test`

### DIFF
--- a/cmd/mxcli/tui/watcher.go
+++ b/cmd/mxcli/tui/watcher.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -78,6 +79,7 @@ func newWatcher(mprPath, contentsDir string, sender MsgSender) (*Watcher, error)
 
 func (w *Watcher) run(sender MsgSender) {
 	var debounceTimer *time.Timer
+	var debounceSeq atomic.Uint64
 
 	for {
 		select {
@@ -110,7 +112,11 @@ func (w *Watcher) run(sender MsgSender) {
 			if debounceTimer != nil {
 				debounceTimer.Stop()
 			}
+			seq := debounceSeq.Add(1)
 			debounceTimer = time.AfterFunc(watchDebounce, func() {
+				if debounceSeq.Load() != seq {
+					return
+				}
 				sender.Send(MprChangedMsg{})
 			})
 
--- a/cmd/mxcli/tui/watcher_test.go
+++ b/cmd/mxcli/tui/watcher_test.go
@@ -35,10 +35,11 @@ func TestWatcherDebounce(t *testing.T) {
 	}
 	defer w.Close()
 
-	// Rapidly write 5 times — should debounce into a single message
+	// Rapidly write 5 times — should debounce into a single message.
+	// Keep the burst tighter than the debounce window so slow CI machines do
+	// not accidentally let an intermediate timer fire.
 	for i := range 5 {
 		_ = os.WriteFile(unitFile, []byte{byte('a' + i)}, 0644)
-		time.Sleep(50 * time.Millisecond)
 	}
 
 	// Wait for debounce to fire (500ms + margin)
--- a/mdl-examples/doctype-tests/02b-nanoflow-examples.mdl
+++ b/mdl-examples/doctype-tests/02b-nanoflow-examples.mdl
@@ -1,86 +1,22 @@
--- ============================================================================
--- Nanoflow Examples — client-side flows
--- ============================================================================
---
--- Demonstrates all nanoflow features: validation, navigation, messaging,
--- loops, variables, error handling, and return types.
---
--- Nanoflows run client-side (browser/native mobile). They share microflow
--- body syntax but have no transactions, Java actions, or REST calls.
---
--- Key differences from microflows:
---   - No RAISE ERROR / ErrorEvent
---   - No Java actions (use CALL JAVASCRIPT ACTION instead)
---   - No direct REST/external calls (call a microflow for server work)
---   - No binary return type
---   - Error handling per-action via ON ERROR, not transactional ROLLBACK
---   - SYNCHRONIZE available for offline native mobile contexts
---
--- ============================================================================
+-- Nanoflow examples — client-side flows
+-- Nanoflows share microflow body syntax but restrict server-side actions.
 
--- MARK: Module and entity setup
-
+-- Setup
 create module NanoflowExamples;
-create module role NanoflowExamples.User;
-create module role NanoflowExamples.Admin;
-
-/**
- * Product entity used throughout the nanoflow examples.
- */
 create entity NanoflowExamples.Product (
-    Name    : String(200),
-    Price   : Decimal,
-    IsValid : Boolean,
-    Tags    : String(500)
+    Name : String(200),
+    Price : Decimal,
+    IsValid : Boolean
 );
 
--- Helper microflow — server-side save, called from nanoflow examples.
-create microflow NanoflowExamples.ACT_SaveProduct (
-    $Product : NanoflowExamples.Product
-)
-returns Boolean
-begin
-    commit $Product;
-    return true;
-end;
-/
+-- Minimal nanoflow (empty body)
+create nanoflow NanoflowExamples.NF_Empty () begin end;
 
--- Helper page — used by N007_OpenProductDetail (requires Mendix 11.0+ page params).
-create page NanoflowExamples.ProductDetail
-(
-    params: {
-        $Product: NanoflowExamples.Product
-    },
-    title: 'Product Detail',
-    layout: Atlas_Core.Atlas_Default
-)
-{
-    dynamictext text1 (content: 'Product Detail', rendermode: H4)
-}
-/
-
--- ============================================================================
--- MARK: Nanoflows
--- ============================================================================
-
-/**
- * N001: Stand-in nanoflow with no logic.
- * Used as a placeholder during scaffolding.
- */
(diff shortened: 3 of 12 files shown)


hjotha:
Follow-up pushed in `5c3dbca`.

This narrows duplicate-output aliasing to variables that are live on the continuation path. Terminal branch-local declarations no longer force `_2` aliases outside that branch, while the existing same-position duplicate collision case remains covered.

- `make build`
- `make test`
- `make lint-go`
- `go test ./mdl/executor -run 'TestBuildFlowGraph_(DuplicateImplicitOutputAtSamePositionGetsLocalAlias|TerminalBranchDuplicateOutputDoesNotForceAlias)' -count=1`

ako (commented):
## PR #337 — fix: preserve duplicate implicit output aliases

**Overview:** Two independent fixes for alias collisions in the microflow builder:
1. **Implicit output collisions** (`create`/`retrieve` producing the same `$Var` at the same canvas position): assigns `Foo_2`, `Foo_3` suffixes via `uniqueImplicitOutputVariable()`.
2. **Duplicate call-microflow result names** (same output variable across two branches, not referenced): sets `UseReturnVariable=false` via pre-pass `planMicroflowCallOutputDeclarations()`.

The variable-state snapshot/restore in `addIfStatement` is the right approach for scoping aliases to branches. Tests are solid. Good to see a bug-test file alongside the doctype fixture.

---

### Moderate

**`statementVarRefs` has incomplete AST coverage** — `AggregateListStmt`, `ListOperationStmt`, `SendRestRequestStmt`, `CallExternalActionStmt`, `ExportToMappingStmt`, and `CreateListStmt` are not handled. `referencedVars` is used to decide whether a duplicated call-microflow output is safe to suppress (`UseReturnVariable=false`). A variable referenced *only* via one of these missing types will be mis-classified as unreferenced, causing `UseReturnVariable=false` when the downstream code still reads the variable.

Concrete failing pattern:
```mdl
$Items = call Module.GetItems();   -- branch A
$Items = call Module.GetItems();   -- branch B, same name
$Count = COUNT($Items);            -- AggregateListStmt — not seen by statementVarRefs
```
Both call outputs get `UseReturnVariable=false` → `$Items` never declared → `COUNT($Items)` references undefined variable.

Fix: add the missing cases to `statementVarRefs` in `cmd_microflows_builder_graph.go`:
```go
case *ast.AggregateListStmt:
    refs = append(refs, s.InputVariable)
case *ast.ListOperationStmt:
    refs = append(refs, s.InputVariable, s.SecondVariable)
case *ast.CallExternalActionStmt:
    for _, arg := range s.Arguments {
        refs = append(refs, exprVarRefs(arg.Value)...)
    }
case *ast.CreateListStmt:
    // no refs consumed (output-only)
case *ast.SendRestRequestStmt:
    refs = append(refs, exprVarRefs(s.Body)...)
```

---

### Minor

**`doctype-tests/variable_alias_collision.test.mdl` missing prerequisites** — the file creates `SampleAliases.ACT_DuplicateOutputPosition` with return type `SampleAliases.Item` but never creates the module or entity. Running this against a project without that setup will fail. Compare `02-microflow-examples.mdl` which opens with `create module MfTest;` and entity definitions. Add:
```mdl
create module SampleAliases;
create entity SampleAliases.Item ( Name : string(100) );
/
```
before the microflow block (matching the bug-test file which already has this).

---

**`countMicroflowCallOutputs` + inner `walk` in `planMicroflowCallOutputDeclarations` duplicate ~40 lines of identical switch boilerplate.** Each time a new statement type gains an `ErrorHandling` field or output variable, it must be added in two places. A shared `walkStatements(body []ast.MicroflowStatement, fn func(ast.MicroflowStatement))` helper extracting the recursion would halve the maintenance surface. Not a blocker for this PR, but worth a follow-up.

---

**Proposal status is still "Draft"** — `PROPOSAL_microflow_variable_alias_collision.md` line 3. Once this PR merges, it should read `Status: Implemented`.

---

**`MDL_QUICK_REFERENCE.md` description is slightly misleading** — "Describe may alias same-position duplicate outputs for round-trip preservation." The alias is assigned by the *builder*, which runs on both `exec` and `describe` paths. Suggested wording: "Builder assigns numeric aliases when duplicate implicit outputs collide at the same canvas position".

---

### Recurring Findings table update

Suggest adding row 11:

| 11 | AST walker switch-case duplicated across multiple functions — new statement types must be added in every copy | Code quality | Extract a shared `walkStatements(body, fn)` helper; pass a per-function closure for the per-node logic |

hjotha:

- Extracted the variable-reference walker to `cmd_microflows_builder_varutil.go`.
- Added coverage for the missing statement kinds noted in review, including aggregate/list operations, REST/request mapping cases, external actions, export mapping, and list add/remove consumers.
- Updated the doctype fixture with its own synthetic module/entity prerequisites.
- Marked the proposal as implemented and updated the quick-reference wording to describe the builder behavior.

```bash
make build
make test
make lint-go
```

hjotha:
Review follow-up status:

- Extracted the variable-reference walker to `cmd_microflows_builder_varutil.go` so this PR and #366 share one canonical implementation instead of defining duplicate package-level helpers.
- Added the missing AST coverage for aggregate/list/external action/REST/export statements.
- Added the missing doctype prerequisites and marked the variable alias proposal as implemented.
- Updated the quick-reference wording to describe the builder behavior directly.

Local validation after the update:

```bash
make build
make test
```

ako:
## Design concern: silent aliasing treats an invalid model as valid

The premise of this PR is that legacy projects can contain duplicate output variable names that "Studio Pro can keep in the model." But Studio Pro does **not** allow two activities to share the same output variable name — it flags this as a model error. So a project exhibiting this pattern is already invalid, and the correct response is to surface that, not to paper over it with silent renaming.

**Two specific problems:**

**1. DESCRIBE of an invalid model should say so, not emit `$Item_2`**

If describe encounters two activities at the same canvas position with the same output name, the project is broken. Emitting `$Item_2` in the describe output produces MDL that, when re-executed, creates a *different* model than the original (one with a genuinely distinct `$Item_2` variable). The roundtrip is not a fixpoint — it silently transforms the model.

The right output is a comment flagging the problem:
```
-- WARNING: duplicate output variable $Item at position (100, 200) — model is invalid; open in Studio Pro to fix
$Item = create SampleAliases.Item (Name = 'second');
```

**2. Authored MDL with duplicate output names should be rejected, not accepted**

The example script:
```
$Item = create SampleAliases.Item (Name = 'first');
$Item = create SampleAliases.Item (Name = 'second');
```

This is invalid MDL and should produce a validation error:
```
error: variable '$Item' is already declared in this scope
```

Silently accepting it and renaming the second to `$Item_2` means the user gets a model with a variable they never named, with no indication that anything was changed.

**Suggested direction**

- In the **validator** (`validate_microflow.go`): reject duplicate implicit output names with a clear error.
- In **describe** (`cmd_microflows_show_helpers.go`): if duplicate output names are detected while traversing an existing MPR, emit a `-- WARNING` comment rather than inventing aliases.

The aliasing machinery (`uniqueImplicitOutputVariable`, `variableAliases`, `outputVarPositions`) added here should be removed — it is handling a case that should never exist in a valid project.

hjotha:
Review follow-up status:

- Reworked the branch to address the design concern: duplicate output variables are treated as invalid instead of silently aliased.
- Removed `variableAliases`, `outputVarPositions`, `uniqueImplicitOutputVariable`, aliasing docs, and aliasing fixtures.
- Added validation coverage for duplicate authored outputs and describe coverage that emits a warning for invalid existing MPRs without inventing `$Item_2` names.
- Kept the shared variable-reference walker coverage that downstream builder passes use.

Validation on the updated branch:

```bash
go test ./mdl/executor -run 'DuplicateOutput|StatementVarRefs' -count=1
make build
make test
```

hjotha:
Follow-up observation from the local roundtrip audit: the `-- WARNING: duplicate output variable` comment currently produced by the describer correctly flags 2 microflows in a real project whose flow graph has this shape:

```
enum split
├── case A → if pre-check → [true] → shared ExclusiveMerge ─┐
├── case B → if pre-check → [true] → shared ExclusiveMerge ─┤
│                                                           ↓
│                                              shared IF + action + return
├── case (empty) → log error → return
└── case other → log error → return
```

Two cases (A, B) fan in to a common `ExclusiveMerge` and the subtree beyond the merge does the actual work (a `call java action` → `create object` → return). The BSON stores a single copy of each shared action; mx check accepts this as a valid shape. On describe the walker enters the shared subtree once per case and emits `$UUID = call java action …` and `$NewCompanyActivity = create …` twice. Re-exec then produces a truly duplicated BSON that Studio Pro rejects with CE0111.

MDL does not have a native way to represent "shared subtree reachable from multiple cases" (no `goto`, no labeled block). This PR's current behavior — emit the warning, let the user fix it in Studio Pro — is still the right default. Just leaving a marker that the warning is not a describer bug in those cases; it's signalling an MDL representability gap.

No action requested on this PR.

hjotha:

PR #338, opened 2026-04-27
fix: write valid show-page parameter mappings

### LINKED ISSUE #295: [Bug] CREATE SNIPPET with Params + $currentObject page call creates null PageParameterMapping — project fails to open in Studio Pro
## Summary

When a snippet is created via mxcli with a `Params` block **and** contains an `ACTIONBUTTON` with `SHOW_PAGE ... (Param: $currentObject)`, mxcli writes a broken `PageParameterMapping` with a null `Variable` into the BSON. Studio Pro then fails to open the project.

## Environment

- **mxcli version:** `v0.7.0-dirty (2026-04-21T08:30:06Z)` (Go: go1.26.2 windows/amd64)
- **Mendix version:** 10.24.15.93102

## Steps to Reproduce

```mdl
CREATE OR MODIFY SNIPPET AuditTrail_Custom."SNIPPET_Asset_AuditHistory" (
  Params: { $Asset: AssetManagement.Asset },
  Folder: 'Snippets'
) {
  DATAGRID dgLogs (
    DataSource: DATABASE FROM AuditTrail.Log
      WHERE AuditTrail.Log_AudittrailSuperClass = $Asset
      SORT BY DateTime DESC,
    PageSize: 10
  ) {
    COLUMN Details (Caption: 'Details', ShowContentAs: customContent, Hidable: no) {
      ACTIONBUTTON btnLogDetails (
        Caption: 'View',
        Action: SHOW_PAGE AuditTrail.Log_View(Log: $currentObject)
      )
    }
  }
}
```

mxcli reports: `Created snippet AuditTrail_Custom.SNIPPET_Asset_AuditHistory` (success)

Then open the project in Studio Pro.

## Expected Behaviour

The snippet is created correctly, Studio Pro opens without errors.

## Actual Behaviour

Studio Pro fails to open with:

```
System.InvalidOperationException: An error occurred when trying to set the
'Variable' property of a Page parameter mapping in a Snippet with ID <uuid-of-snippet>.
 ---> System.ArgumentNullException: Value cannot be null. (Parameter 'value')
   at Mendix.Modeler.WebUI.Pages.PageParameterMapping.set_Variable(PageVariable value)
```

The UUID in the error matches the newly created snippet. The root cause appears to be that mxcli does not correctly wire `$currentObject` as a `PageVariable` reference when serializing the `SHOW_PAGE` action inside a parameterized snippet — resulting in a null `Variable` on the `PageParameterMapping`.

## Recovery

Drop the broken snippet via mxcli before Studio Pro can be opened:

```bash
./mxcli -p Project.mpr -c "DROP SNIPPET Module.SnippetName"
```

## Relation to Issue #291

Issue #291 describes a null `PageParameterMapping` caused by a **SNIPPETCALL without Params**. This issue is distinct: the null variable is produced during **snippet creation itself**, not when calling a snippet. The triggering factor appears to be the combination of:
1. A snippet with a `Params` block (defining a page variable like `$Asset`)
2. An `ACTIONBUTTON` inside the snippet with `SHOW_PAGE SomePage(Param: $currentObject)`

The `$currentObject` context variable (row context in a DATAGRID column) is not resolved to a valid `PageVariable` during serialization.

## Severity

**High** — same silent-success + crash pattern as #291. No error or warning from mxcli. Studio Pro cannot recover without running the `DROP SNIPPET` workaround via mxcli.

## Diagnostic Bundle

A diagnostic bundle (`mxcli-diag-20260424-113038.tar.gz`) is available and can be shared upon request.

hjotha:
## Summary

- writes `Forms$PageParameterMapping.Variable` as a valid empty `Forms$PageVariable` object instead of `null`
- uses the Mendix storage-list marker for `FormSettings.ParameterMappings` rather than the mapping count
- adds a BSON-level regression test for the serialized show-page parameter mapping shape

## Why

Studio Pro rejects `PageParameterMapping.Variable = null` while loading a project. `mxcli exec` could succeed, but the resulting MPR was not openable when a show-page action contained page parameter mappings.

## Validation

- `make build`
- `make lint-go`
- `make test`

Closes #295.
Related to #332.

### DIFF
--- a/mdl-examples/bug-tests/295-showpage-parameter-mapping-validity.mdl
+++ b/mdl-examples/bug-tests/295-showpage-parameter-mapping-validity.mdl
@@ -1,0 +1,56 @@
+-- ============================================================================
+-- Bug #295: SHOW PAGE parameter mappings produced unopenable Studio Pro MPRs
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   When `mxcli exec` wrote a microflow `show page X (Param: $Var)`, the
+--   `Forms$PageParameterMapping.Variable` field was serialized as BSON
+--   `null`. Studio Pro rejected this on project load with a runtime
+--   exception, making the entire MPR unopenable. The originating bug
+--   also affected SNIPPET ACTIONBUTTON SHOW_PAGE (#295), but the root
+--   cause is in the writer's serialization of PageParameterMapping.
+--
+--   Additionally, `FormSettings.ParameterMappings` used the mapping count
+--   as the BSON storage-list first element instead of Mendix's storage-list
+--   marker (`int32(2)`).
+--
+-- After fix:
+--   `writer_microflow_actions.go` now writes
+--   - `Variable` as a valid empty `Forms$PageVariable` object via
+--     `emptyPageVariable()`
+--   - `ParameterMappings` storage-list with the constant marker `int32(2)`
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/295-showpage-parameter-mapping-validity.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest295.ACT_OpenProductPage"
+--   `mx check` against the resulting MPR must report 0 errors and the
+--   project must be openable in Studio Pro.
+-- ============================================================================
+
+create module BugTest295;
+
+create entity BugTest295.Product (
+  Name : string(100)
+);
+/
+
+create page BugTest295.Product_Detail
+(
+  params: {
+    $Product: BugTest295.Product
+  },
+  title: 'Product Detail',
+  layout: Atlas_Core.Atlas_Default
+)
+{
+  dynamictext txtName (content: 'Product detail', rendermode: H4)
+}
+/
+
+create microflow BugTest295.ACT_OpenProductPage (
+  $Product: BugTest295.Product
+)
+begin
+  show page BugTest295.Product_Detail (Product: $Product);
+end;
+/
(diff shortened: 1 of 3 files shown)


ako (commented):
## PR #338 — fix: write valid show-page parameter mappings

Clean, minimal bug fix. Both root causes are correctly identified and addressed:

- `int32(len(...))` as the storage-list marker was wrong (coincidentally correct only when count == 2); `int32(2)` is consistent with the sentinel used for every other `ParameterMappings` array in the same file (lines 231, 265).
- `Variable: nil` fails Studio Pro's object-model loader; the empty `Forms$PageVariable` document satisfies the structural requirement.

The BSON-level test is exactly what this kind of fix needs — it pins both the sentinel value and the `Variable.$Type` in a way that will catch any regression immediately.

---

### Minor

**`TitleOverride: nil` in the same `FormSettings` block** — the PR fixes `Variable: nil` because Studio Pro rejects null embedded objects, but `TitleOverride` is also written as `nil` (line ~367 in the current code). If `TitleOverride` is a string/expression scalar, `null` is typically accepted by Studio Pro; if it's an embedded object, it needs the same treatment. Worth a quick check against a Studio Pro-generated BSON sample to confirm `null` is safe there. Not a blocker since issue #295 only reported `Variable`, but worth noting.

---

### Checklist

- No executor or backend interface changes — fix is entirely within `sdk/mpr/`, the correct layer. ✓
- Regression test covers both bugs (sentinel value and non-null variable). ✓
- Bug-test MDL file documents symptom, fix, and verification steps. ✓
- No grammar/AST/visitor changes needed. ✓

PR #340, opened 2026-04-27
fix: refresh empty change-object actions

### LINKED ISSUE #339: Empty change-object microflow actions can be written as invalid CE0032 activities
## Summary

A described microflow can contain an empty change-object activity represented in MDL as:

```mdl
change $Object;
```

When `mxcli exec` rebuilds that statement, it writes a `ChangeObjectAction` with no member changes, `Commit = No`, and `RefreshInClient = false`. Studio Pro validation then reports CE0032 because the action has no items and does not commit the object.

## Expected behavior

Round-tripping an existing empty change-object activity should keep the action valid.

## Actual behavior

The rebuilt MPR can contain a no-op `ChangeObjectAction` that Studio Pro reports as invalid:

```text
[CE0032] Change action has no items specified and does not commit the object.
```

## Proposed fix

When the builder sees a `change $Object;` statement with no member assignments, it should set `RefreshInClient = true`. Change actions with explicit member assignments should keep the existing behavior.

## Related

hjotha:
## Summary

- sets `RefreshInClient` when rebuilding `change $Object;` with no member assignments
- keeps explicit member-change actions unchanged
- adds a builder regression test for the empty change-object case

## Why

An empty, non-committing `ChangeObjectAction` without refresh is invalid in Studio Pro and reports CE0032. Existing models can use empty change activities as refresh-only actions, and the MDL form currently has no explicit `refresh` modifier for `change` statements, so the builder needs to preserve validity by inferring refresh for empty changes.

## Validation

- `make build`
- `make lint-go`
- `make test`

Closes #339.
Related to #332.

### DIFF
--- a/mdl-examples/bug-tests/339-empty-change-refresh-in-client.mdl
+++ b/mdl-examples/bug-tests/339-empty-change-refresh-in-client.mdl
@@ -1,0 +1,33 @@
+-- ============================================================================
+-- Bug #339: Empty change-object actions must refresh in client
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   Rebuilding an empty change-object activity from MDL wrote a non-committing
+--   ChangeObjectAction with no member changes and RefreshInClient=false.
+--   Studio Pro / mx check rejected that activity with CE0032:
+--     "Change action has no items specified and does not commit the object."
+--
+-- After fix:
+--   `change $Object;` is serialized as a refresh-only change action by setting
+--   RefreshInClient=true. Although the CE0032 text mentions only items/commit,
+--   empirical mx check validation accepts refresh as the third valid escape.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/339-empty-change-refresh-in-client.mdl -p app.mpr
+--   mx check app.mpr
+--   Expected: 0 new CE0032 errors for BugTest339.MF_RefreshOnlyChange.
+-- ============================================================================
+
+create module BugTest339;
+
+create entity BugTest339.Item (
+  Name: String(100)
+);
+
+create microflow BugTest339.MF_RefreshOnlyChange ()
+begin
+  $Item = create BugTest339.Item;
+  change $Item;
+end;
+/
--- a/mdl/executor/bugfix_regression_test.go
+++ b/mdl/executor/bugfix_regression_test.go
@@ -675,6 +675,27 @@ func TestCallMicroflowResultType_ResolvesSubsequentChangeMember(t *testing.T) {
 	}
 }
 
+func TestEmptyChangeObjectRefreshesInClient(t *testing.T) {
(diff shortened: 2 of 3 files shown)


hjotha:
Addressed the review follow-up in `8d5cc94`:

- added `mdl-examples/bug-tests/339-empty-change-refresh-in-client.mdl` documenting the mx check behavior for the CE0032 case
- added an inline builder comment explaining why `RefreshInClient=true` is valid for an empty non-committing change action even though the CE0032 text only mentions items/commit
- rechecked the fixture with `./bin/mxcli check mdl-examples/bug-tests/339-empty-change-refresh-in-client.mdl`
- reran `go test ./mdl/executor -run TestEmptyChangeObjectRefreshesInClient`

I am keeping the separate concern of preserving explicit RefreshInClient on change actions with member assignments out of this PR; that needs a syntax/design decision for a `change ... refresh` modifier rather than being bundled into the CE0032 fix.

ako (commented):
## PR #340 — fix: refresh empty change-object actions

Clean, minimal fix. The condition `len(s.Changes) == 0` is the right predicate — it correctly infers refresh-only for empty changes while leaving non-empty changes unchanged (same `RefreshInClient=false` as before). The regression test in `bugfix_regression_test.go` directly pins the BSON field value.

I verified the bug-test file syntax against the grammar: `String(100)` is fine (lexer tokens are case-insensitive), `create BugTest339.Item` without `()` is valid (parens are optional in `createObjectStatement`), and the missing `/` after the entity is fine (`SEMICOLON? SLASH?` are both optional at the statement level).

---

### Minor — non-empty changes with `RefreshInClient=true` still lose refresh on roundtrip

This PR fixes the empty case, but there's a symmetrical gap: if the original Studio Pro model has a *non-empty* `ChangeObjectAction` with `RefreshInClient=true` (e.g., a change + explicit refresh for UI update purposes), `describe` emits `change $Obj (attr = val);` and re-executing that sets `RefreshInClient=false` because `len(s.Changes) > 0`.

This is a **pre-existing** gap (the old code also wrote `false` unconditionally), so this PR doesn't make it worse. The complete fix would be to add a `WITH REFRESH` modifier to `changeObjectStatement` in the grammar — similar to the existing `COMMIT VARIABLE (WITH EVENTS)? REFRESH?` pattern. Worth capturing as a follow-up issue, since `COMMIT` and `ROLLBACK` already have the pattern to copy.

---

### Minor — test covers only the empty case

`TestEmptyChangeObjectRefreshesInClient` verifies that an empty `ChangeObjectStmt` gets `RefreshInClient=true`. A complementary case — that a non-empty `ChangeObjectStmt` preserves `RefreshInClient=false` — would protect against accidentally flipping the condition. One extra `if/else` assertion in the same test would do it.

---

### Checklist

- Fix is in the executor builder layer, not touching `sdk/mpr` directly. ✓
- Bug-test MDL file is well-documented. ✓
- No grammar/AST/visitor changes needed (fix is inference-only). ✓

PR #342, opened 2026-04-27
fix: preserve microflow-call BSON field order

### LINKED ISSUE #341: Microflow call parameter mappings can be written with invalid BSON field order
## Summary

Some microflow call activities can roundtrip textually but become invalid after `mxcli exec`. Studio Pro reports CE0117 expression errors on the call activity even though the MDL source is unchanged.

## Expected behavior

Serializing a `MicroflowCallAction` should preserve the BSON field order Studio Pro expects for:

- the top-level `MicroflowCallAction`
- the nested `MicroflowCall`
- each `MicroflowCallParameterMapping`

## Actual behavior

The writer emits result-variable fields before the nested call, `QueueSettings` before parameter mappings, and mapping fields as `Parameter` before `Argument`. This can produce a model that loads but fails validation with an expression error on the call activity.

## Proposed fix

Emit the call payload in the same stable order as Studio Pro writes it: nested `MicroflowCall` before result fields, `ParameterMappings` before `QueueSettings`, and mapping `Argument` before `Parameter`.

## Related

hjotha:
## Summary

- writes `MicroflowCallAction` with nested `MicroflowCall` before result-variable fields
- writes nested `MicroflowCall.ParameterMappings` before `QueueSettings`
- writes each parameter mapping as `Argument` then `Parameter`
- adds a BSON-level writer regression test using generic names

## Why

A microflow call can roundtrip MDL text unchanged but still become invalid in Studio Pro validation if the BSON shape differs from Studio Pro's expected field order. The observed failure is CE0117 on the call activity.

## Validation

- `make build`
- `make lint-go`
- `make test`

Closes #341.
Related to #332.

### DIFF
--- a/mdl-examples/bug-tests/341-microflow-call-bson-field-order.mdl
+++ b/mdl-examples/bug-tests/341-microflow-call-bson-field-order.mdl
@@ -1,0 +1,61 @@
+-- ============================================================================
+-- Bug #341: Microflow call BSON field order produced CE0117 in Studio Pro
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   `mxcli exec` could roundtrip the textual MDL of a microflow call
+--   (`$Result = call microflow Module.Target (Param = $Var);`) unchanged,
+--   but Studio Pro flagged the resulting activity with
+--     [CE0117] expression error on the call activity
+--   even though the MDL source had not been edited. The cause was BSON
+--   field order:
+--   - The writer emitted result-variable fields BEFORE the nested
+--     `MicroflowCall` document.
+--   - Inside `MicroflowCall`, `QueueSettings` was emitted BEFORE
+--     `ParameterMappings`.
+--   - Each `MicroflowCallParameterMapping` was emitted with `Parameter`
+--     BEFORE `Argument`.
+--   Studio Pro's expression validator depends on this order being stable
+--   to resolve parameter references.
+--
+-- After fix:
+--   `serializeMicroflowAction` now emits:
+--   - `MicroflowCallAction`: $ID, $Type, ErrorHandlingType, MicroflowCall,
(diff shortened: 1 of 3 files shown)


ako (commented):
## PR #342 — fix: preserve microflow-call BSON field order

Clean, targeted fix. The three field-order changes are exactly right and the `assertBSONKeys` test helper pins the contract with precision — far better than a roundtrip test that can silently pass despite wrong field order.

The logic is straightforward: move `ResultVariableName`/`UseReturnVariable` to after `MicroflowCall`, move `QueueSettings` to after `ParameterMappings`, swap `Argument`/`Parameter`. The diff is minimal and the test directly validates each level of the nested document.

---

### Minor — `JavaActionCallAction` has the same pre-fix field-order pattern

`JavaActionCallAction` (line 237 in the current file) serializes `QueueSettings` before `ParameterMappings` using the same pattern this PR is correcting for microflow calls:

```go
{Key: "QueueSettings", Value: nil},          // before ParameterMappings
{Key: "ResultVariableName", Value: ...},
{Key: "UseReturnVariable",  Value: ...},
```

If Studio Pro's expression validator is order-sensitive for microflow calls, it's plausible it applies the same constraint to Java action calls. No CE0117-equivalent has been reported for Java action calls yet (issue #341 was specifically about microflow calls), but this is worth checking against a Studio Pro reference BSON for `JavaActionCallAction`. If it needs the same fix, it's a one-commit follow-up using the same approach.

---

### Checklist

- Fix is in `sdk/mpr/` (the BSON writer layer). ✓
- `assertBSONKeys` / `bsonValue` are new helpers — no naming conflicts with existing test helpers (`toMap`, `extractString`) in the same package. ✓
- Bug-test MDL is complete and well-documented. ✓
- No grammar, AST, visitor, or backend interface changes needed. ✓

PR #344, opened 2026-04-27
fix: preserve attribute-based list operations

### LINKED ISSUE #343: Attribute-based find/filter list operations can roundtrip as invalid expressions
## Summary

`find($List, Attribute = expression)` and `filter($List, Attribute = expression)` can be described correctly but rebuilt as expression-based list operations. The resulting MPR can then fail Studio Pro validation with CE0117 on the list operation activity.

## Expected behavior

When the MDL condition is an equality comparison on a list member, `mxcli exec` should rebuild the same attribute-based list operation shape that Studio Pro stores as `Microflows$Find` or `Microflows$Filter`.

## Actual behavior

The builder always creates `FindByExpression` / `FilterByExpression`, and the writer does not serialize the attribute-based variants. This loses the stored operation shape even when the text output looks equivalent.

## Proposed fix

Detect `Attribute = expression` conditions during build, resolve the member as an attribute or association from the input list type, build `FindByAttributeOperation` / `FilterByAttributeOperation`, and serialize them as `Microflows$Find` / `Microflows$Filter`.

## Related

hjotha:
## Summary

- builds `find/filter($List, Attribute = expression)` as attribute-based list operations when possible
- resolves the left-hand member to an attribute or association using the input list type
- serializes `FindByAttributeOperation` / `FilterByAttributeOperation` to `Microflows$Find` / `Microflows$Filter`
- adds builder and writer regression tests with generic model names

## Why

The parser and formatter already know about attribute-based list operations, but `mxcli exec` rebuilt them as expression-based operations. That can produce an MPR that is textually stable but fails Studio Pro validation with CE0117 on the list operation activity.

## Validation

- `make build`
- `make lint-go`
- `make test`

Closes #343.
Related to #332.

### DIFF
--- a/mdl-examples/bug-tests/343-list-attribute-find-filter.mdl
+++ b/mdl-examples/bug-tests/343-list-attribute-find-filter.mdl
@@ -1,0 +1,56 @@
+-- ============================================================================
+-- Bug #343: Attribute-based find/filter list operations rebuilt as expressions
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   `find($List, Attribute = expression)` and
+--   `filter($List, Attribute = expression)` could be described correctly
+--   from existing models, but `mxcli exec` always rebuilt them as
+--   `FindByExpression` / `FilterByExpression`. Studio Pro's expression
+--   validator then surfaced CE0117 on the list operation activity even
+--   though the MDL source had not been edited.
+--
+-- After fix:
+--   The builder resolves the equality LHS to an attribute (or association)
+--   on the input list's element type and emits
+--   `Microflows$Find` / `Microflows$Filter` (`FindByAttributeOperation` /
+--   `FilterByAttributeOperation`) when possible. Complex conditions still
+--   fall back to the expression-based shape.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/343-list-attribute-find-filter.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest343.MF_FindByAttribute"
+--   `mx check` against the resulting MPR must report 0 errors and the
+--   activities must not surface CE0117 in Studio Pro.
+-- ============================================================================
+
+create module BugTest343;
+
+create entity BugTest343.Item (
+  Name : string(100),
+  Active : boolean
+);
+/
+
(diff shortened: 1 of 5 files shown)


ako (commented):
## Review — fix: preserve attribute-based list operations

### Overview

Solid fix for a real bug. The storage names are verified correct against the existing parser (`"Microflows$Find"` / `"Microflows$Filter"` — confirmed in `sdk/mpr/parser_microflow.go:967,981`). BSON field names match what the parser reads (`Attribute`, `Association`, `Expression`, `ListName`). Test coverage for the builder and writer paths is good.

---

### Moderate: no guard when entity type is unknown produces unqualified attribute names

`listAttributeOperation` calls `resolveListOperationMember`, which calls `resolveMemberChangeFallback` when the backend is unavailable or when `varTypes` doesn't contain the list variable. In that case, a bare member name like `Code` (0 dots, no entity context) produces `AttributeQualifiedName = "Code"` — unqualified. The function then returns a non-nil `FindByAttributeOperation` with `Attribute: "Code"`, which Studio Pro would reject.

The expression-based fallback (`FindByExpression`) is always safe; the attribute-based form needs a fully qualified name. A guard would make this explicit:

```go
attributeName, associationName := fb.resolveListOperationMember(s.InputVariable, fieldName)
if attributeName == "" && associationName == "" {
    return nil
}
```

That still doesn't catch `attributeName = "Code"` (bare, no dot). A tighter guard:

```go
// Fall back to expression form if we can't produce a qualified name
if associationName == "" && !strings.Contains(attributeName, ".") {
    return nil
}
```

In practice the executor always populates `varTypes` from parameters, so this is an edge case — but it's the kind that silently produces bad BSON without error.

---

### Minor: `MemberChange` used outside its intended context

`resolveListOperationMember` allocates a `MemberChange` (a ChangeObjectAction concept) purely to extract attribute/association qualified names. This works, but future readers will be confused by a change-object type appearing in a list operation context. A small comment on the function explaining the reuse would help.

---

### Positive

- Storage names verified against the existing parser — no guessing.
- Fallback to expression form for non-equality conditions (`binary.Operator != "="`) and non-identifier LHS is correct and safe.
- Bug test in `mdl-examples/bug-tests/` is the right location for a fix regression script.
- Writer test covers both `Attribute` and `Association` paths independently.

PR #354, opened 2026-04-27
fix: preserve manual while-true loop roundtrips

### LINKED ISSUE #350: Preserve manual while-true retry-loop roundtrips

## Summary

Manual retry-loop style graphs can be described as `while true` with `continue`/terminal paths. The current rebuild path can add a fallthrough EndEvent or fail to preserve a loop when only nested bodies contain `break`/terminal paths.

## Expected behavior

`describe -> exec -> describe` should preserve manual while/retry-loop semantics, including:

- loop-back flow without an extra fallthrough EndEvent;
- nested-loop `break` that belongs to the nested loop, not the outer manual loop;
- terminal branches inside the loop body;
- valid MPR output after roundtrip.

## Current staging status

The relevant code exists in draft PR #346 but depends on umbrella-only manual loop reconstruction. It should be extracted with that reconstruction rather than as the tiny follow-up alone.

## Validation target

Focused PR should include synthetic tests for manual loop preservation and pass `make build`, `make lint-go`, and `make test`.

hjotha:
Closes #350.
Part of #332.

## Summary

This extracts the manual `while true` loop preservation from the audit staging branch into a focused fix.

The builder already accepted `while`, `break`, and `continue` syntax, but it did not materialize break/continue events and it treated `continue` as a normal fallthrough path. For retry-style `while true` loops this could rebuild the flow as a `LoopedActivity` or synthesize an invalid fallthrough `EndEvent`.

## Changes

- Build `BreakEvent` and `ContinueEvent` objects for parsed break/continue statements.
- Treat break/continue as terminal branch statements for control-flow analysis.
- Detect manual `while true` bodies that need a graph-level back-edge and build them as an `ExclusiveMerge` loop header.
- Route `continue` inside that manual loop back to the merge instead of emitting a standalone `ContinueEvent`.

## Validation

- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/mdl-examples/bug-tests/350-manual-while-true-loop-roundtrip.mdl
+++ b/mdl-examples/bug-tests/350-manual-while-true-loop-roundtrip.mdl
@@ -1,0 +1,62 @@
+-- ============================================================================
+-- Bug #350: Manual `while true` retry-loop roundtrip
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   Studio Pro projects with manual retry loops authored as
+--     while true
+--     begin
+--       <work>
+--       if <done> then return ...; end if;
+--       continue;
+--     end while;
+--   could be described correctly, but `mxcli exec` rebuilt the same MDL
+--   into either a `LoopedActivity` (collection-style loop) or a flow
+--   with a synthetic fallthrough `EndEvent`. The resulting MPR diverged
+--   from the original — a back-edge merge expected by Studio Pro was
+--   missing — and `mx check` could fail or the model's runtime semantics
+--   could change silently.
+--
+-- Root cause:
+--   The builder accepted `while`, `break`, `continue` syntax but never
+--   materialized BreakEvent/ContinueEvent objects. `continue` in a
+--   manual `while true` body was treated as plain fallthrough.
+--
+-- After fix:
+--   - `BreakEvent` and `ContinueEvent` objects are produced for break/continue.
+--   - Both are treated as terminal branch statements for control-flow analysis.
+--   - Manual `while true` bodies that need a graph-level back-edge are
+--     built as an `ExclusiveMerge` loop header.
+--   - `continue` inside that manual loop routes back to the merge instead
+--     of producing a standalone `ContinueEvent`.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/350-manual-while-true-loop-roundtrip.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest350.MF_ManualRetryLoop"
+--   `mx check` against the resulting MPR must report 0 errors.
+-- ============================================================================
+
+create module BugTest350;
+
+create entity BugTest350.Item (
+  IsDone : boolean
+);
+/
+
+-- Manual while-true retry loop. The graph rebuild must use an
+-- ExclusiveMerge back-edge — not a LoopedActivity and not a fallthrough
+-- EndEvent — and the `continue` must route to that merge.
+create microflow BugTest350.MF_ManualRetryLoop (
+  $Item: BugTest350.Item
+)
+returns boolean as $Done
+begin
+  while true
+  begin
+    if $Item/IsDone then
+      return true;
+    end if;
+    continue;
+  end while;
+end;
+/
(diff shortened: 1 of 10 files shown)


ako (commented):
## PR Review — fix: preserve manual while-true loop roundtrips (#354)

**No blockers.** One moderate issue, two minor notes.

---

### Moderate

**1. `containsContinueStmt` recurses into nested loops — asymmetric with `containsBreakForCurrentLoop`**

`containsBreakForCurrentLoop` correctly stops at nested loop boundaries (with the comment "A break inside a nested loop exits that nested loop, not this manual while-true back-edge"). `containsContinueStmt` does NOT stop at nested loop boundaries — it recurses into both `LoopStmt.Body` and `WhileStmt.Body`.

Consequence: a `while true` body that has a `continue` only inside a nested collection loop (with no outer `continue` or `return`) is incorrectly classified as a manual-while-true candidate:

```mdl
while true
begin
  loop $item in $items
  begin
    continue;   -- targets the inner loop, not the outer while
  end loop;
  -- no outer continue, no return/raise
end while;
```

`containsContinueStmt` returns `true` (finds the inner continue), so `isManualWhileTrueCandidate` returns `true`, and the outer `while true` is built as an `ExclusiveMerge` back-edge structure instead of a `LoopedActivity` with `WhileLoopCondition`. Before the PR it would have been a `LoopedActivity` (`!continue && !terminal` → not a candidate). The ExclusiveMerge approach creates an infinite outer loop in the graph, which may or may not be rejected by `mx check`.

Confirmed safe part: standard loop/while builders (`addLoopStatement` line 367, `addWhileStatement` line 465) create fresh sub-builders without copying `manualLoopBackTarget`, so inner loop `continue`s emit a proper `ContinueEvent` in the inner body — that side is not affected.

Fix: add a `containsContinueForCurrentLoop` that mirrors `containsBreakForCurrentLoop`:

```go
func containsContinueForCurrentLoop(stmts []ast.MicroflowStatement) bool {
    for _, stmt := range stmts {
        switch s := stmt.(type) {
        case *ast.ContinueStmt:
            return true
        case *ast.IfStmt:
            if containsContinueForCurrentLoop(s.ThenBody) || containsContinueForCurrentLoop(s.ElseBody) {
                return true
            }
        case *ast.LoopStmt, *ast.WhileStmt:
            continue // targets the nested loop, not this one
        }
    }
    return false
}
```

Use this in `isManualWhileTrueCandidate` instead of `containsContinueStmt`. The broader `containsContinueStmt` can be kept for its one current use-site in `isManualWhileTrueCandidate` (checking for any terminal signal including nested loops), or removed entirely once `containsContinueForCurrentLoop` is in place.

Note: `isTerminalStmt` for `*ast.WhileStmt` calls `isManualWhileTrueCandidate`, which has the same false-positive for the same reason. Fixing point 1 fixes this transitively.

---

### Minor

**2. `TestBuildFlowGraph_ManualWhileTrueTerminalDoesNotAddFallthroughEnd` passes trivially when no EndEvent is added**

The test checks `if end.ReturnValue == "" { t.Fatal(...) }`. Because `addManualWhileTrueStatement` sets `endsWithReturn = true`, `buildFlowGraph` adds no EndEvent at all, so the loop body is never entered and the test always passes. It would also pass if a bug caused the EndEvent to be omitted for the wrong reason.

A stronger assertion: after `buildFlowGraph`, assert that `oc.Objects` contains zero `*microflows.EndEvent` with an empty return value — i.e., explicitly count `EndEvent` nodes and assert the count is zero (or that every EndEvent has a non-empty return value).

**3. No DESCRIBE roundtrip test for `continue` inside a branch**

The existing tests verify that `while true ... continue` at the body level routes a back-edge to the merge and that DESCRIBE emits `while true / end while;`. But there is no test for the case where `continue` appears inside an IF branch within the while-true body:

```mdl
while true
begin
  if condition then
    continue;   -- back-edge via IF branch
  end if;
  <work>
  continue;
end while;
```

In the built graph the IF branch ending at the merge has no `ContinueEvent` node — `traverseFlowUntilMerge` must emit `continue;` when it detects a flow going to the `stopID`. A test for this shape would confirm the DESCRIBE roundtrip (build → describe → re-exec produces the same graph).

---

### Positive

- Bug-test file (`350-manual-while-true-loop-roundtrip.mdl`) present with `mx check` validation instructions. ✓
- `containsBreakForCurrentLoop` correctly stops at nested loop boundaries — the asymmetry with `containsContinueStmt` is the only structural gap.
- The save/restore of `manualLoopBackTarget` in `addManualWhileTrueStatement` correctly handles nested manual while-true loops.
- `bodyReturns` extension in `validate_microflow.go` correctly suppresses MDL003 false positives for `while true` without break.
- Five new focused tests covering the key cases plus the validator behaviour.

Issue #351, opened 2026-04-27
Preserve nested branch tails and split-join continuations


hjotha:
Part of #332. Staged in draft PR #346.

## Summary

Nested splits can have guarded tails, empty branch joins, partial joins, or branch-local case/anchor state that the describer/builder currently cannot always pair back to the right continuation.

## Expected behavior

Roundtrip should preserve:

- guarded branch tails that continue after nested joins;
- validation-feedback targets that are variables rather than attribute paths;
- nested merge fallback when a child split has no directly paired merge;
- split joins that converge at a non-merge activity;
- branch case/anchor state when leaving a nested split.

## Current staging status

Draft PR #346 contains the implementation, but it is split across several historical commits and needs focused extraction. Some pieces may need to wait for shared traversal helpers to land first.

## Validation target

Focused PRs should include synthetic tests for guarded tails, variable validation feedback targets, and nested split join fallback, then pass `make build`, `make lint-go`, and `make test`.

PR #355, opened 2026-04-27
fix: preserve no-merge branch continuations

hjotha:
Closes part of #351.
Part of #332.

## Summary

This extracts one focused split/join builder fix from the audit staging branch.

When an `if/else` has one terminal branch and one continuing branch, the builder does not need to insert an `ExclusiveMerge`. In that no-merge shape, the parent flow must continue from the tail of the non-terminal branch. Previously it could continue from the original split instead, leaving the branch body disconnected or causing the following statement to be wired to the wrong origin.

## Changes

- Track the continuation endpoint for no-merge `if` branches.
- Preserve pending branch case and anchor state from nested compound statements.
- Continue the parent flow from the branch tail instead of the stale split when a branch body exists.

## Validation

- `make build`
- `make lint-go`
- `make test`


ako (commented):
## Review — fix: preserve no-merge branch continuations

### Overview

Correct fix for a real wiring bug. The core change — tracking `noMergeExitID` from the non-terminal branch's tail instead of always wiring the parent to the split — is sound, and the `TestBuildFlowGraph_NoMergeIfElseContinuesFromBranchTail` test directly demonstrates the before/after behaviour. The bug test MDL is well-documented.

---

### Moderate: `branchDestinationAnchor` reversal has no test and no explanatory comment

```diff
-    if stmtAnchor != nil && stmtAnchor.To != ast.AnchorSideUnset {
-        return stmtAnchor
+    if branchAnchor != nil && branchAnchor.To != ast.AnchorSideUnset {
+        return branchAnchor
     }
-    return branchAnchor
+    return stmtAnchor
```

This swaps which anchor takes precedence for the first activity in a branch — a 4-line change that quietly reverses existing behaviour with no comment explaining why the previous order was wrong, and no regression test covering it.  

The new order (branch anchor's `To` beats statement anchor's `To`) is plausible — the branch-level annotation from the split should govern how the flow arrives at the first branch statement, not the statement's own anchor — but without a test or comment this is invisible to reviewers and future maintainers. The `TestBuildFlowGraph_NestedNoMergeTailCarriesAnchorToParentMerge` test touches anchor propagation but doesn't exercise this specific precedence swap.

---

### Moderate: `flowLooksLikeGuardContinuation` uses Y-position equality as a heuristic

```go
return dest.GetPosition().Y == split.GetPosition().Y
```

This works for builder-generated layouts where the guard's false path is always placed horizontally at split-Y. For MPRs with manually repositioned activities (Studio Pro drag-and-drop), or for builder layouts that drift due to future spacing changes, the equality check could misidentify or silently miss the guard pattern.

The previous `!branchFlowStartsAtTerminal` check was wrong because it triggered when the false branch led to a real merge, but the Y-position substitute is equally brittle in the other direction. Worth a comment documenting the layout contract this relies on, and ideally a test with an MPR fixture that verifies the heuristic doesn't regress when the guard is at a non-standard Y.

---

### Minor: `else { fb.nextConnectionPoint = splitID }` drops `nextFlowCase`/`nextFlowAnchor`

The old code always set `nextFlowCase` and `nextFlowAnchor` in the deferred path. The new code only sets them when `noMergeExitID != ""`:

```go
if noMergeExitID != "" {
    fb.nextConnectionPoint = noMergeExitID
    fb.nextFlowCase = noMergeExitCase
    fb.nextFlowAnchor = noMergeExitAnchor
} else {
    fb.nextConnectionPoint = splitID
    // nextFlowCase and nextFlowAnchor NOT set here
}
```

Analysis shows `noMergeExitID` is always populated when this path is reached (both IF-with-ELSE and IF-without-ELSE set it before this block), so the else branch is a dead code path in practice. But if it were ever reached, the missing `nextFlowCase`/`nextFlowAnchor` would silently produce an uncased flow from the split — the same bug this PR is fixing. Worth a `// unreachable: noMergeExitID is always set above` comment, or a `panic`/`return` to make it fail loudly.

---

### Minor: pending case/anchor pattern duplicated three times

`pendingThenCase` / `pendingThenAnchor` / `pendingElseCase` / `pendingElseAnchor` appears in three separate loops (IF+ELSE then-body, IF+ELSE else-body, IF-without-ELSE then-body), each with the same 10-line check block. Not a blocker, but the repetition will make future anchor-handling changes error-prone.

---

### Positive

- `noMergeExitID` correctly handles all four combinations: then-returns/else-continues, else-returns/then-continues, empty then-body, empty else-body.
- Builder test is precise — it asserts the specific `elseID → afterID` flow exists and that `afterID` has no other origin.
- `!needMerge` guard structure is clean: only enters the no-merge branch when the invariant holds.

ako:
## Review: fix: preserve no-merge branch continuations

Overall this is a correct and well-scoped fix. The core problem — `fb.nextConnectionPoint = splitID` was set unconditionally in the no-merge case even when the continuing branch had advanced past the split — is clearly diagnosed, and `noMergeExitID` tracking solves it correctly for all three cases (thenReturns, elseReturns, IF without ELSE).

Two minor issues worth addressing.

---

### Minor 1 — "create flow twice" pattern repeated 6×

In `cmd_microflows_builder_control.go`, each pending-case block unconditionally creates the flow with a case, then immediately overwrites it if the case is empty:

```go
flow = newHorizontalFlowWithCase(lastThenID, actID, pendingThenCase)
if pendingThenCase == "" {
    flow = newHorizontalFlow(lastThenID, actID)
}
```

The first allocation is wasted whenever `pendingThenCase == ""`. This pattern appears 6 times. Prefer:

```go
if pendingThenCase != "" {
    flow = newHorizontalFlowWithCase(lastThenID, actID, pendingThenCase)
} else {
    flow = newHorizontalFlow(lastThenID, actID)
}
```

Not a correctness bug, but avoids silent dead allocations.

---

### Minor 2 — Y-position heuristic in `flowLooksLikeGuardContinuation` has no edge-case test

`flowLooksLikeGuardContinuation` uses `dest.GetPosition().Y == split.GetPosition().Y` to identify guard-continuation flows (horizontal, same row). The comment notes this relies on "mxcli's layout contract", which is fine — but there's no test exercising the case where a non-guard continuation activity happens to land at the same Y as the split (e.g. a short if-without-else where the activity after the single branch stays on the base row). If that case ever produces a false positive it will silently suppress the `false` case label.

A test along the lines of `TestFlowLooksLikeGuardContinuation_SameYNonGuard` would make the heuristic's boundary explicit.

---

### Positives

- `branchDestinationAnchor` reversal is correct: the branch annotation owns the incoming edge to the first branch activity, so it should win over the statement annotation.
- `pendingThenCase`/`pendingElseCase` propagation correctly threads deferred flow-case state from nested compound statements up to the parent merge, avoiding the "anchor lost on nested if" bug.
- The two new tests (`TestBuildFlowGraph_NoMergeIfElseContinuesFromBranchTail`, `TestBuildFlowGraph_NestedNoMergeTailCarriesAnchorToParentMerge`) directly exercise the fix and its nested variant.

No blockers. Both issues are minor polish.

Issue #352, opened 2026-04-27
Preserve microflow type and cardinality inference during rebuild


hjotha:
Part of #332. Related to #199. Staged in draft PR #346.

## Summary

Several roundtrip validity failures come from missing type/cardinality inference in microflow rebuilds. When the builder cannot infer whether a value is an object, list, association reference, or mapped response, it can write valid-looking MDL that becomes an invalid MPR.

## Expected behavior

The builder/writer should preserve MPR validity for:

- compact association retrieve sources;
- sort attributes that require indirect entity refs;
- Java action return types, including generic list/object returns where inferable;
- import mapping result cardinality;
- list operations and `add <expression> to $List` values;
- inferred return values for non-void microflows.

## Current staging status

Draft PR #346 contains several type/cardinality inference fixes. These need to be extracted as focused PRs by root cause, preferably starting with cases that do not add new MDL syntax.

Focused PRs extracted so far:

- #356: compact reverse-association retrieve describe form.
- #357: Java action return type inference.
- #360: import mapping result variable type registration.
- #362: expression-valued add-to-list statements.

## Validation target

Focused PRs should include synthetic unit tests for each inference case and pass `make build`, `make lint-go`, and `make test`.

PR #356, opened 2026-04-27
fix: preserve compact reverse-association retrieves

hjotha:
Closes part of #352.
Part of #332.

## Summary

This extracts the reverse-association retrieve formatter fix from the audit staging branch.

Some Mendix models store a compact association retrieve as a database retrieve with a single XPath predicate like `[Module.Association = $Variable]` and range `All`. The formatter previously described that as a verbose database retrieve with a `where` clause, causing avoidable describe/exec/describe drift.

## Changes

- Detect the narrow reverse-association database retrieve shape.
- Verify through the backend domain model that the association target matches the retrieved entity before compacting.
- Emit `retrieve $Result from $Variable/Module.Association;` only for safe `All` range, unsorted, single-predicate cases.
- Keep existing database retrieve output for complex predicates, non-variable RHS, sorting, non-`All` ranges, and non-matching targets.

## Validation

- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/mdl-examples/bug-tests/352-retrieve-compact-reverse-association.mdl
+++ b/mdl-examples/bug-tests/352-retrieve-compact-reverse-association.mdl
@@ -1,0 +1,56 @@
+-- ============================================================================
+-- Bug #352 (part): Compact reverse-association retrieve formatter
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   Some Mendix models store a compact association retrieve as a
+--   `DatabaseRetrieveSource` with a single XPath predicate of the form
+--     [Module.Association = $Variable]
+--   and `Range = All`. When described, the formatter emitted the verbose
+--   shape
+--     retrieve $Domains from Module.Domain
+--       where Module.Domain_Runtime = $Runtime;
+--   On `mxcli exec`, that verbose shape is rebuilt as XPath. Studio Pro
+--   then accepts it but `describe → exec → describe` is no longer a
+--   fixpoint, and tooling that expects the compact form sees drift.
+--
+-- After fix:
+--   When the source is a database retrieve over a single equality
+--   predicate against a known association whose other side matches the
+--   start variable's type, and `Range = All`, the formatter emits the
+--   compact form `retrieve $Out from $Var/Module.Association`.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/352-retrieve-compact-reverse-association.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest352a.MF_FetchDomains"
+--   The describe output must use the compact `from $Runtime/Mod.Assoc`
+--   form, and `mx check` must report 0 errors.
+-- ============================================================================
+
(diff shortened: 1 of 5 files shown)


ako (commented):
## PR #356 — fix: preserve compact reverse-association retrieves

Solid, well-tested fix. The two-sided approach (formatter compaction + builder guard refinements) is the right structure. The formatter change is a pure addition (early-return if the compact form applies), the tests cover the key acceptance and rejection paths, and the builder additions are guarded so they don't change existing behaviour for the common case.

---

### Minor — stale comment after new `ReferenceSet` branch

`cmd_microflows_builder_actions.go`, in the `else` branch that follows the new `else if assocInfo.Type == AssociationTypeReferenceSet`:

```go
} else {
    // ReferenceSet or unknown: returns a list    ← stale
    fb.varTypes[s.Variable] = "List of " + assocQN
}
```

`ReferenceSet` is now handled by the branch above; this path only runs for truly unknown types. The comment should read `// Unknown association type: returns a list typed as the association name`.

---

### Minor — bug-test MDL doesn't reproduce the originally broken shape

`352-retrieve-compact-reverse-association.mdl` calls `retrieve $Domains from $Runtime/BugTest352a.Domain_Runtime;`. That MDL creates an `AssociationRetrieveSource` via the builder, which the formatter already emitted in compact form before this PR. The actual broken shape (a `DatabaseRetrieveSource` with a single `[Assoc = $Var]` predicate written by Studio Pro) can only be exercised by describing an existing MPR — not by exec-from-scratch.

The unit test `TestFormatAction_Retrieve_ReverseAssociationDatabaseSourceUsesCompactForm` correctly exercises the fixed path. The bug-test only validates describe→exec→describe round-trip stability (which was already working). Worth a one-liner comment in the bug-test header clarifying this, e.g.: `"-- Note: this script exercises the round-trip (exec→describe) path; the DatabaseRetrieveSource→compact conversion is validated by unit tests."` Not a correctness concern, just avoids confusion for future readers.

---

### Minor — `isSimpleMendixName` permits leading underscores

```go
if r == '_' || r >= 'A' && r <= 'Z' || ...
```

A name like `_Domain_Runtime` passes, which is more permissive than Mendix naming conventions (entity/association names start with a letter). Not harmful in practice — the domain model lookup gates the actual compaction — but worth tightening if the helper is reused elsewhere.

---

### Checklist

- `parseReverseAssociationRetrieve` called before verbose output; falls through safely when ctx is nil, range is non-All, or entity doesn't match. ✓
- `splitTopLevelXPathPredicates` handles nested brackets, quoted strings, unclosed brackets (returns false). ✓
- Builder guard `Owner != Both` + `parentPersistable` correctly prevents DatabaseRetrieveSource for non-queryable cases; tests cover both. ✓
- `databaseRetrieveMatchesAssociationTarget` correctly checks parent entity (the FROM/owner side) against the retrieved entity — consistent with Mendix's inverted `ParentPointer` semantics. ✓
- No grammar/AST/visitor/backend interface changes. ✓

PR #357, opened 2026-04-27
fix: infer java action result variable types

hjotha:
Closes part of #352.
Part of #332.

## Summary

This extracts the Java action result-type inference fix from the audit staging branch.

The builder already looks up Java action definitions to format entity-typed parameters, but it did not register the output variable type from the Java action return metadata. Downstream statements could therefore see Java action outputs as unknown even when the action returns a concrete entity/list/file type.

## Changes

- Register result variable types for Java actions returning concrete entity and list types.
- Register `System.FileDocument` for Java actions returning `FileDocument`.
- Infer generic list-return types from list-typed input arguments when the Java action return type is a list without a concrete entity.

## Validation

- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/mdl-examples/bug-tests/352-java-action-return-type-inference.mdl
+++ b/mdl-examples/bug-tests/352-java-action-return-type-inference.mdl
@@ -1,0 +1,56 @@
+-- ============================================================================
+-- Bug #352 (part): Java action result variable type inference
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   The microflow builder looked up Java action definitions to format
+--   entity-typed parameters, but it did NOT register the Java action's
+--   return type on the output variable. Subsequent statements that
+--   touched the result (`change`, `commit`, `$Result/Attribute`,
+--   association paths) had no type info, so:
+--     - the writer fell back to untyped expressions
+--     - `mx check` could surface CE0117 on dependent activities
+--     - describe re-emitted the call without the right variable
+--       declaration shape
+--
+-- After fix:
+--   `addCallJavaActionAction` consults the Java action's `ReturnType`
+--   and registers the output variable as the appropriate object/list
+--   type (`Module.Entity`, `List of Module.Entity`, `System.FileDocument`,
+--   etc.) so downstream activities resolve correctly.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/352-java-action-return-type-inference.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest352b.MF_UseJavaResult"
+--   `mx check` against the resulting MPR must report 0 errors.
+-- ============================================================================
+
+create module BugTest352b;
+
+create persistent entity BugTest352b.LogEntry (
+  Message : string(500)
+);
+/
+
+-- Java action returning an entity object — the result variable type must
+-- be inferred so downstream `change` / `commit` resolve.
+create java action BugTest352b.CreateLogEntry (
+  Msg: string not null
+) returns BugTest352b.LogEntry
+as $$
+return new com.mendix.systemwideinterfaces.core.IMendixObject();
+$$;
+/
+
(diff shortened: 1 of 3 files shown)


ako (commented):
## PR #357 — fix: infer java action result variable types

---

### No Blockers

The overall approach is correct. Registering the output variable type after a Java action call so downstream statements can resolve attributes and associations is the right fix. The three new tests validate the main code paths.

---

### Moderate Issues

#### 1. Dead `case javaactions.ListType:` (value form) in `inferGenericJavaActionReturnType`

The codebase consistently stores `ListType` as a pointer (`&javaactions.ListType{}`): every call site in `sdk/mpr/parser_javaactions.go` and `mdl/executor/cmd_javaactions.go` uses the pointer form. `javaActionReturnVarType` only handles `*javaactions.ListType`.

`inferGenericJavaActionReturnType` adds a redundant second case:
```go
case *javaactions.ListType:
    if t.Entity != "" { return "" }
case javaactions.ListType:       // ← dead code: never stored as value
    if t.Entity != "" { return "" }
```

The value case will never be reached. Remove it. Leaving it suggests the author wasn't sure which form is used, which creates false uncertainty for future readers.

#### 2. String-mangling to extract variable names from arguments is fragile

`inferGenericJavaActionReturnType` resolves argument expressions via:
```go
valueExpr := strings.TrimPrefix(strings.Trim(fb.exprToString(arg.Value), "'"), "$")
if typ := fb.varTypes[valueExpr]; strings.HasPrefix(typ, "List of ") {
```

`exprToString` → strip quotes → strip `$` → look up in `varTypes`. This only accidentally works for simple `*ast.VariableExpr` nodes. For any compound expression the string manipulation produces a garbage key and silently yields no type. A direct type assertion is cleaner and expresses intent:
```go
if v, ok := arg.Value.(*ast.VariableExpr); ok {
    if typ := fb.varTypes[v.Name]; strings.HasPrefix(typ, "List of ") {
        return typ
    }
}
```

#### 3. No unit test for `EntityType` return — the most common case

`javaActionReturnVarType` handles three cases. Tests cover `FileDocumentType` and both `ListType` variants. The `*javaactions.EntityType` case — returning a plain domain entity — is the most common Java action return type, and it has no unit test. The MDL bug-test file exercises it manually, but a one-liner unit test alongside the other three would protect against regressions:

```go
func TestAddJavaAction_EntityReturnRegistersEntityType(t *testing.T) {
    // backend returns &javaactions.EntityType{Entity: "Orders.Order"}
    // assert fb.varTypes["Result"] == "Orders.Order"
}
```

---

### Minor Issues

#### 4. `measurer` inconsistency across tests

Tests 1 and 2 initialise `flowBuilder` without a `measurer`; test 3 adds `measurer: &layoutMeasurer{}`. Since `exprToString` → `expressionToString` doesn't touch `measurer`, both patterns work today. But the inconsistency will confuse future test authors about whether `measurer` is required. Either add `measurer` to all three for uniformity, or document why test 3 specifically needs it (it doesn't, based on the code path).

---

### Positive Notes

- `javaActionReturnVarType` is a clean, flat switch with no side effects — easy to extend for new return types. ✓
- The `jaDef != nil && fb.varTypes != nil` guards at the call site are correct and consistent with the rest of the builder. ✓
- Bug-test MDL file has the `create module BugTest352b;` prerequisite and exercises the entity-return case end-to-end. ✓
- Naming the file `352-java-action-return-type-inference.mdl` with a `b` suffix on the module name (`BugTest352b`) avoids collisions with other parts of issue #352. ✓

---

**Summary:** No blockers. Three moderate issues worth addressing before merge: remove the dead `ListType` value case, replace the string-mangling with a direct `VariableExpr` assertion, and add a unit test for the `EntityType` return. One minor inconsistency in test setup.

ako:
@hjotha I merged it, but can you have a look at the open issues?

PR #359, opened 2026-04-27
fix: preserve validation feedback targets

### LINKED ISSUE #358: Preserve validation feedback object and association targets
## Summary

Microflow `validation feedback` statements can target an object itself or an association member, not only a scalar attribute. The current MDL grammar and formatter only preserve the attribute-target form reliably.

## Problem

When a model contains validation feedback attached to an object-level target or an association target, describe/exec/describe can either fail to parse the described MDL or can drop the association target and emit feedback against the object only.

## Expected behavior

`validation feedback` should support:

```mdl
validation feedback $FormObject message 'Select a value';
validation feedback $FormObject/Module.Object_Association message 'Select a value';
```

Attribute targets should keep their existing behavior.

## Notes

Part of #332.

hjotha:
## Summary

Fixes validation feedback round-trips for targets that are not simple attributes.

## Changes

- Accept object-only validation feedback targets such as `$Object` in MDL.
- Preserve qualified association targets such as `$Object/Module.Object_Association` as a single path segment.
- Emit `AssociationName` when a validation feedback action does not have an `AttributeName`.

## Validation

- `go test ./mdl/executor -run TestFormatAction_ValidationFeedback`
- `go test ./mdl/visitor -run TestValidationFeedback`
- `make build`
- `make lint-go`
- `make test`

Closes #358
Part of #332


ako (commented):
## Review — fix: preserve validation feedback targets

### Overview

Well-scoped fix. The three-part approach (grammar → visitor → formatter) is correct, the scope note in the bug test explicitly documents what remains out of scope (builder producing CE0091 for object-only authored targets), and the formatter change is minimal. Tests cover the new grammar forms, the builder paths, and the formatter paths.

---

### Moderate: `attributePath` grammar change is broader than its context suggests

The diff changes:
```diff
 attributePath
-    : VARIABLE ((SLASH | DOT) (IDENTIFIER | qualifiedName))+
+    : VARIABLE ((SLASH | DOT) qualifiedName)+
```

`attributePath` is used in five places beyond `validationFeedbackStatement`:
- `setStatement` (`SET $var/Attr = expr`)
- `loopStatement` (`LOOP $item IN $list/assoc`)
- `aggregateExpression` (`SUM($list/attr)`, `AVERAGE`, `MINIMUM`, `MAXIMUM`)

The change is backward-compatible — a bare `IDENTIFIER` is still a valid `qualifiedName` — but none of these other usages have regression tests that exercise the grammar-change path. A `$Order/SUM` or `$List/Module.Attr` through the SET or LOOP paths should still parse identically; worth a comment in the grammar noting that the rule is shared and the change is intentional.

---

### Moderate: old two-segment association path (`len(segs) >= 2`) is now dead for associations

Under the old grammar, `$Obj/Module.Association` produced **two segments** (`"Module"` sep `/`, `"Association"` sep `.`), which is why the code had:

```go
} else if len(segs) >= 2 && segs[0].Separator == "/" && segs[1].Separator == "." {
    // Reconstruct "Module.AssociationName" from segments
```

Under the new grammar `qualifiedName` absorbs the dot, so `$Obj/Module.Association` now produces **one segment** `"Module.Association"` — handled correctly by the new `case 1` path. The `len(segs) >= 2` branch is now only reachable for multi-hop paths like `$Obj/Part1/Part2`, where its comment ("association qualified name") is now misleading. The reconstruction logic (`strings.Join(parts, ".")`) would incorrectly join multi-hop segments into an association name in that case.

The old code path should either be removed (since the grammar change makes it unreachable for single associations) or its comment and logic updated to reflect that it now only handles multi-hop traversal paths.

---

### Minor: redundant `else if` in multi-segment block

```go
} else if entityQName == "" && strings.Count(segs[0].Name, ".") == 1 {
    associationName = segs[0].Name
```

This branch sits inside the `len(segs) >= 2` block, so it applies when: there are 2+ segments, the path doesn't match the `/` then `.` separator pattern, `entityQName` is empty, and the first segment has exactly one dot. Under the new grammar this combination is hard to construct in a well-formed microflow — a description of when this fires would help.

---

### Positive

- Grammar fix is correct: `qualifiedName` subsumes bare identifiers so all existing `attributePath` syntax still parses.
- `validationFeedbackStatement` accepting bare `VARIABLE` is the right grammar shape for the object-only target.
- Dot-count heuristic for single segments (0 → attribute, 1 → association, 2+ → qualified attribute) is consistent with `resolveMemberChangeFallback` elsewhere.
- Scope note in the bug test is honest and specific — the CE0091 gap is documented, not hidden.

hjotha:

PR #360, opened 2026-04-27
fix: register import mapping result variable types

hjotha:
## Summary

Preserves inferred object/list type information for import-from-mapping output variables.

## Changes

- Reuses the import mapping result entity that is already written into `ResultHandlingMapping`.
- Registers the output variable as either `Module.Entity` or `List of Module.Entity` based on the inferred `SingleObject` flag.
- Adds synthetic tests for single-object and array-root import mapping metadata.

## Validation

- `go test ./mdl/executor -run TestAddImportFromMappingRegisters`
- `make build`
- `make lint-go`
- `make test`

Part of #352
Part of #332

### DIFF
--- a/mdl-examples/bug-tests/360-import-mapping-result-type.mdl
+++ b/mdl-examples/bug-tests/360-import-mapping-result-type.mdl
@@ -1,0 +1,66 @@
+-- ============================================================================
+-- Bug #360: Import-from-mapping result variable type was not registered
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   `import from mapping` actions wrote the correct
+--   `Forms$ResultHandlingMapping` BSON, but the builder did not register
+--   the inferred result type onto the output variable in its variable
+--   scope. Subsequent activities that touched the imported result
+--   (`change`, `commit`, `$Var/Attribute`, association paths) ran without
+--   type info, which caused:
+--     - the writer to fall back to untyped expressions
+--     - re-describes to lose the right declaration shape
+--     - downstream activities to fail with CE0117 in Studio Pro
+--
+-- After fix:
+--   `addImportFromMappingAction` now also writes the result type into
+--   `flowBuilder.varTypes`. When the mapping's JSON structure produces a
+--   single object, the variable becomes `Module.Entity`. When it produces
+--   an array, the variable becomes `List of Module.Entity`. The
+--   ResultHandlingMapping BSON path is unchanged.
+--
+-- Usage:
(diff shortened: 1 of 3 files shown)


ako (commented):
## PR #360 — fix: register import mapping result variable types

---

### No Blockers, No Moderate Issues

The fix is minimal and correct. Extracting `resultEntityQN` before the `resultHandling.ResultEntityID` assignment is a clean one-liner refactor that introduces no behavioural change to the BSON path — it just makes the qualified name available for the `varTypes` registration that follows.

Key details confirmed:
- `ImportMappingElement.Entity` is documented as a **qualified entity name** (not a BSON ID), so `fb.varTypes[s.OutputVariable] = resultEntityQN` stores the correct form for downstream type checks. ✓
- `SingleObject` is determined by `js.Elements[0].ElementType == "Array"` earlier in the function — the tests use `"Object"` and `"Array"` which exactly match that condition. ✓
- Guards (`s.OutputVariable != ""`, `resultEntityQN != ""`, `fb.varTypes != nil`) are all necessary and correctly placed. ✓
- `GetImportMappingByQualifiedNameFunc` and `GetJsonStructureByQualifiedNameFunc` mock stubs both exist in `mdl/backend/mock/mock_mapping.go`. ✓
- Bug-test MDL has `create module BugTest360;` prerequisite and exercises the downstream `change $Pet (Status = ...)` dependency that requires the type to be registered. ✓

---

### Minor Issues

#### 1. `importMappingFlowBuilder` omits `declaredVars`

```go
return &flowBuilder{
    varTypes: map[string]string{},
    backend:  &mock.MockBackend{...},
}
```

Other test helpers in the executor package initialise `declaredVars: map[string]string{}` alongside `varTypes`. `addImportFromMappingAction` doesn't access `declaredVars` in these code paths, so the tests pass — but the omission is inconsistent and would cause a nil-map panic if the code path ever expanded to touch `declaredVars`. Adding it for uniformity is low-cost.

#### 2. No test for backend-unavailable / lookup-error path

When the backend is nil or `GetImportMappingByQualifiedName` returns an error, `resultEntityQN` stays `""` and the registration is silently skipped. This is correct behaviour (can't infer without metadata), but there's no test asserting `fb.varTypes` stays empty in that case. Very minor — the guard logic is straightforward — but a one-liner negative test would make the contract explicit.

---

**Summary:** No blockers or moderate issues. Two very minor gaps in test completeness and helper consistency. The fix itself is clean and the approach is right.

PR #362, opened 2026-04-27
feat: allow expressions in add-to-list statements

### LINKED ISSUE #361: Allow expressions in add-to-list microflow statements
## Summary

The current MDL grammar only accepts a variable item in add-to-list statements:

```mdl
add $Item to $Items;
```

Some existing models store a `ChangeListAction` add value as an expression, for example an attribute path or function call. Describing that action can produce valid-looking MDL, but the parser/builder cannot preserve it because the AST only stores a simple item variable.

## Expected behavior

MDL should preserve add-to-list values that are expressions:

```mdl
add $Order/Number to $Numbers;
add toUpper($Name) to $Names;
```

The existing simple variable form should continue to work unchanged.

## Relation

Part of #352 and #332.

hjotha:
## Summary

Allows microflow add-to-list statements to preserve expression-valued add items.

## Changes

- Extends the grammar from `add $Item to $List` to `add expression to $List`.
- Stores the parsed expression on `AddToListStmt` while preserving the legacy simple-variable `Item` field.
- Writes the expression value into `ChangeListAction.Value` during rebuild.
- Adds proposal docs, a doctype fixture, quick-reference entry, and skill guidance for the new syntax.

## Validation

- `go test ./mdl/visitor -run TestAddToList`
- `go test ./mdl/executor -run TestAddToListBuilder`
- `./bin/mxcli check mdl-examples/doctype-tests/add_expression_to_list.test.mdl`
- `make build`
- `make lint-go`
- `make test`

Closes #361
Part of #352
Part of #332


ako (commented):
## Review — feat: allow expressions in add-to-list statements

### Overview

Clean feature extension — grammar change is minimal (`ADD expression TO VARIABLE`), the visitor dual-sets `Item` and `Value` for backward compat, and the builder fallback chain is correct. The proposal and skill docs are good.

---

### Moderate: `validate_microflow.go` and `builder_graph.go` only track `s.Item`, not `s.Value`

Two places consume `AddToListStmt` and look only at `s.Item`:

**`validate_microflow.go:425`**:
```go
case *ast.AddToListStmt:
    refs = append(refs, s.Item, s.List)
```

**`cmd_microflows_builder_graph.go:312`**:
```go
case *ast.AddToListStmt:
    if s.Item != "" {
        inputs[s.Item] = true
    }
```

For `add head($SourceItems) to $Items`, `s.Item = ""` and `s.Value = CallExpr{...}`. Neither place sees `$SourceItems` as a referenced variable — the validator will silently skip it (missing "variable not declared" check) and the flow graph won't include it in the activity's input set (affecting layout/ordering analysis).

Fix: walk the expression when `s.Value != nil`:
```go
case *ast.AddToListStmt:
    if s.Value != nil {
        refs = append(refs, exprVarRefs(s.Value)...)
    } else {
        refs = append(refs, s.Item)
    }
    refs = append(refs, s.List)
```

Same pattern in `builder_graph.go`:
```go
case *ast.AddToListStmt:
    if s.Value != nil {
        walkExpr(s.Value)
    } else if s.Item != "" {
        inputs[s.Item] = true
    }
```

---

### Moderate: `.test.mdl` suffix — CI won't run the doctype example

`mdl-examples/doctype-tests/add_expression_to_list.test.mdl` is skipped by `TestMxCheck_DoctypeScripts` (same issue flagged in PRs #334 and #336). This is now the third PR in this batch with this suffix. Rename to `.mdl`.

---

### Minor: `Item` field leaves a maintenance trap in the AST

`AddToListStmt.Item` is now a derived shadow of `Value` — set only when `Value` is a `VariableExpr`, used only as a fallback when `Value` is nil. Any code path that constructs `AddToListStmt` manually with only `Item` set (e.g., tests, codegen) will silently get the old behavior without setting `Value`, making the dual-state confusing. The AST comment says "kept for simple $Var compatibility" — consider whether the field is still needed at all, or whether the builder fallback `"$" + s.Item` can be dropped in favour of always requiring `Value`.

---

### Minor: `TestAddToListKeepsSimpleVariableCompatibility` doesn't assert `Value`

The test checks `addStmt.Item == "Order"` but not that `addStmt.Value` is a `*ast.VariableExpr{Name: "Order"}`. The dual-state behaviour of the visitor is only half-tested.

---

### Positive

- Grammar change is backward-compatible: bare `$Item` is still a valid `expression`, so existing scripts parse unchanged.
- Builder fallback chain (`exprToString(Value)` → `"$" + Item`) is safe and correctly prioritises the expression.
- Visitor correctly identifies `VariableExpr` to populate `Item` for the legacy path.

hjotha:
Review follow-up status:

The current branch already includes the requested `AddToListStmt.Value` handling:

- `cmd_microflows_builder_graph.go` walks `s.Value` with `exprVarRefs(...)` when present.
- `validate_microflow.go` also walks `s.Value` when present before falling back to the legacy `s.Item` field.

I revalidated the focused executor coverage:

```bash
go test ./mdl/executor -run 'AddToList|ExpressionToList|StatementVarRefs' -count=1
```

hjotha:

hjotha:
Thanks @ako — processed the review:

**Moderate (both): already applied in earlier commits, plus a third site you flagged by pattern.**

- `validate_microflow.go:483` and `cmd_microflows_builder_graph.go:324` already walk `s.Value` via `exprVarRefs` when non-nil (commit `0f3fbe39`).
- Your review surfaced a third copy of the pattern I missed: `errorHandlerStatementVarRefs` in `cmd_microflows_builder_flows.go:507` was still doing `refs = append(refs, s.Item, s.List)`. Same silent-drop issue for `add head($X) to $Y` inside error-handler bodies. Fixed in `fa1c9090` with regression test `TestErrorHandlerStatementVarRefsSeesAddExpressionValue`.

**Moderate (test.mdl suffix): already fixed.** File is at `mdl-examples/doctype-tests/add_expression_to_list.mdl` (no `.test.mdl`), matches the pattern from #334/#336 cleanup.

**Minor 1 (`Item` as maintenance trap):** legitimate concern. Dropping the field would touch 7 files across visitor/builder/validator/describer; I'd rather do it as its own refactor PR once the feature is in, so the dual-state removal lands independently and can be reviewed on its own merits. Current comment on the field (`"kept for simple $Var compatibility"`) signals the deprecation intent.

**Minor 2 (`TestAddToListKeepsSimpleVariableCompatibility` doesn't assert Value):** already asserted — `visitor_add_to_list_test.go:70-76` checks `addStmt.Value.(*ast.VariableExpr).Name == "Order"`.

ako:
## Review: feat/allow-expressions-in-add-to-list-statements

**Result: Approved with one minor nit.**

### What the PR does

Extends the  grammar from accepting only a bare  to accepting any expression. Useful for round-tripping Studio Pro microflows that use an expression (e.g. `head()`, `if ... then ... else ...`) as the list-add value. The legacy  form continues to work unchanged.

### Checklist

- [x] **Grammar change is minimal and correct** —  — one token widened, regenerated parser files included
- [x] **AST** —  added alongside legacy ; backward-compatible shim in visitor: bare  still populates  for callers that pre-date this PR
- [x] **Builder** —  with  fallback is the right pattern; matches how all other expression fields are built
- [x] **DESCRIBE roundtrip** — formatter emits ; the new grammar now accepts any expression in that position, so complex Studio Pro expressions survive a full round-trip without any formatter changes needed
- [x] ** and ** — both updated to walk  via ; correct and consistent with each other
- [x] ** extended with  and ** — these were genuinely missing before and would have caused MDL005 scope-checking to silently miss variable references inside those expression types anywhere they appeared (not just in ); fixing them here is the right call
- [x] **Test coverage** — visitor, builder, and validate tests cover both expression and legacy-variable paths; validate test confirms MDL005 fires correctly for out-of-scope add expression variables
- [x] **Doctype fixture** —  demonstrates the if-then-else form

### Minor nit

The proposal file was added with  even though the implementation is in the same PR. Update to  before merge.

### Notes

The  fix for / is a useful cleanup beyond the stated scope of the PR. No concerns — it strictly improves correctness.

PR #363, opened 2026-04-27
feat: preserve change object refresh modifier

### LINKED ISSUE #345: Preserve refresh-in-client on change-object actions with member assignments
## Summary

`change $Object;` is now handled as a refresh-only change action for #339, but there is a related roundtrip gap for non-empty change actions.

A Studio Pro model can contain a `ChangeObjectAction` that both changes members and has `RefreshInClient = true`. Current MDL syntax has no way to express that flag for `change $Object (Member = Value);`, so a describe/exec roundtrip rewrites such actions with `RefreshInClient = false`.

## Expected behavior

mxcli should be able to preserve `RefreshInClient` on change-object actions with member assignments.

## Actual behavior

The existing MDL form:

```mdl
change $Object (Name = 'value');
```

has no `refresh` modifier, and the builder writes non-empty change actions with `RefreshInClient = false`.

## Possible direction

Add a syntax modifier parallel to commit/rollback, for example:

```mdl
change $Object (Name = 'value') refresh;
```

The describer could emit it only when the underlying action has `RefreshInClient = true`, preserving current output for the default false case.

## Relation to #339

#339 fixes the validity issue for empty changes by inferring refresh, because `change $Object;` would otherwise be rejected by mx check with CE0032. This issue is separate: it concerns preserving an explicit refresh flag on non-empty change actions and likely needs a small MDL language extension.

hjotha:
## Summary

Adds explicit MDL syntax for preserving `ChangeObjectAction.RefreshInClient`.

## Changes

- Accepts `change $Object refresh;` and `change $Object (...) refresh;`.
- Carries the refresh flag through the AST and builder.
- Emits `refresh` when describing change-object actions with `RefreshInClient=true`.
- Adds proposal docs, a doctype fixture, quick-reference entry, and skill guidance.

## Validation

- `go test ./mdl/visitor -run TestChangeObjectParsesRefresh`
- `go test ./mdl/executor -run 'Test(FormatAction_ChangeObject.*Refresh|ChangeObjectBuilderWritesRefreshInClient)'`
- `./bin/mxcli check mdl-examples/doctype-tests/change_refresh_modifier.test.mdl`
- `make build`
- `make lint-go`
- `make test`

Closes #345
Part of #332


ako (commented):
## Review — feat: preserve change object refresh modifier

### Overview

Clean, minimal PR. The grammar change is one token (`REFRESH?`) consistent with the existing `COMMIT ... REFRESH?` and `ROLLBACK ... REFRESH?` pattern. The full-stack pipeline (grammar → visitor → AST → builder → formatter) is complete, test coverage is good, and the `s.RefreshInClient || len(s.Changes) == 0` expression in the builder correctly preserves the existing auto-refresh-on-empty logic.

---

### Info: describe output changes for existing empty-change actions (intentional, but undocumented)

Before this PR, `describe` on an MPR whose empty `ChangeObjectAction` had `RefreshInClient = true` in BSON emitted `change $Customer;` (no `refresh` — the formatter didn't know about the flag). After this PR it emits `change $Customer refresh;`.

Both forms exec to identical BSON, so the roundtrip is stable. But any user who previously saved described MDL scripts will see the output diverge. This is intentional (making the implicit flag explicit) but the PR summary doesn't mention it. Worth a one-line note in the PR description so users aren't surprised by diff noise on `describe` output for existing projects.

---

### Minor: CE0032 reference dropped from the builder comment

The original comment included:
```
// The CE0032 message mentions only items/commit, but mx check accepts
// RefreshInClient=true as the third valid escape.
```

The new comment omits the `CE0032` reference. That error code is exactly what a developer would search for when wondering why empty changes force `RefreshInClient = true`. Worth keeping.

---

### Minor: `.test.mdl` suffix — fourth PR in this batch with this CI gap

`change_refresh_modifier.test.mdl` won't be picked up by `TestMxCheck_DoctypeScripts`. Same issue flagged in PRs #334, #336, and #362. Rename to `.mdl`.

---

### Positive

- `REFRESH` is already a keyword token — no new lexer rules needed.
- Grammar placement (`REFRESH?` after the optional member list) is unambiguous and consistent with `COMMIT`/`ROLLBACK`.
- Visitor one-liner `stmt.RefreshInClient = changeCtx.REFRESH() != nil` is correct.
- Builder correctly ORs the explicit flag with the auto-refresh-on-empty condition.
- Both member and no-member formatter paths emit `refresh` correctly.
- Existing `TestEmptyChangeObjectRefreshesInClient` and `TestFormatAction_ChangeObject_NoChanges` tests are unaffected (both use `RefreshInClient = false`).

PR #364, opened 2026-04-27
feat: support enum split microflow statements

### LINKED ISSUE #347: Preserve enum split microflow roundtrips

## Summary

Some legacy microflows use enum-based ExclusiveSplit conditions. `describe microflow` can flatten or misclassify those graphs because MDL currently has no explicit enum split statement that preserves enum case values, grouped cases, and terminal/shared-tail behavior.

## Expected behavior

`describe -> exec -> describe` should preserve an enum split as structured MDL, including:

- the split source variable or attribute path;
- each enum case label, including grouped cases;
- an `(empty)` case when present;
- terminal branches and shared continuation tails;
- MPR validity after roundtrip.

## Current staging status

The behavior is implemented in draft PR #346, but that PR is intentionally not merge-ready. This needs extraction into a focused PR with parser/AST/visitor/builder/describer/validator coverage and synthetic tests only.

## Validation target

Focused PR should pass `make build`, `make lint-go`, `make test`, and at least one `mxcli check` doctype example for the new syntax.

hjotha:
## Summary

Adds first-class MDL support for enumeration decision splits in microflows:

```mdl
split enum $Status
  case Open, Pending
    return true;
  case (empty)
    return false;
  else
    return false;
end split;
```

## Root Cause

The existing microflow model only had a boolean IF representation for exclusive splits. Studio Pro enumeration decisions are also exclusive splits, but their outgoing sequence flows carry `EnumerationCase` values. Without a dedicated MDL statement, describe/exec round-trips cannot preserve those case values cleanly.

## Fix

- Adds `EnumSplitStmt` and `EnumSplitCase` to the microflow AST.
- Extends the grammar and visitor to parse `split enum` with multi-value cases and optional `else`.
- Builds `ExpressionSplitCondition` graphs with `EnumerationCase` sequence flows.
- Describes existing enum split graphs back to `split enum` syntax.
- Updates validation, layout, terminality, reference walking, docs, skill guidance, and doctype examples.

## Validation

- `go test ./mdl/visitor -run TestEnumSplit -count=1`
- `go test ./mdl/executor -run 'Test(EnumSplit|TraverseFlow_EnumSplit|LastStmtIsReturn_EnumSplit|ValidateMicroflow_EnumSplit)' -count=1`
- `./bin/mxcli check mdl-examples/doctype-tests/enum_split_statement.test.mdl`
- `make build`
- `make lint-go`
- `make test`

Closes #347
Part of #332


hjotha:
Follow-up pushed in `0eedb66`.

This preserves explicit `split enum` case order across describe/exec/describe by encoding the parsed order into valid sequence-flow connection pairs and sorting enum split cases by that order during describe.

- `make build`
- `make test`
- `make lint-go`
- `go test ./mdl/executor -run 'TestTraverseFlow_EnumSplit|TestEnumSplitBuilderCreatesEnumerationCaseFlows' -count=1`

ako (commented):
## PR #364 Review — feat: support enum split microflow statements

Good addition overall. The grammar, AST, builder, describe, and validation layers are all wired up. A few issues to address before merging.

---

### Moderate

**1. `splitCaseOrderAnchors` silently fails for 17+ enum values**

`cmd_microflows_builder_actions.go` — `splitCaseOrderAnchors` is a 16-entry hardcoded table. For enumerations with more than 16 cases, `applySplitCaseOrder` returns the zero-value `[2]int{0,0}`, producing flows with connection index 0 for every case. Studio Pro will reject the resulting shape because two cases compete for the same anchor.

Fix options:
- Hard-cap at 16 and return a validation error: `"enum split with more than 16 cases is not yet supported"`
- Or compute the anchors algorithmically from the case count (Studio Pro lays them out symmetrically); the hardcoded table is a special case of that formula.

The silent path is the dangerous part — add the cap/error now even if the formula comes later.

---

**2. `isTerminalStmt` dead code for `EnumSplitStmt`**

`cmd_microflows_builder_flows.go` — the final block reads:

```go
if len(s.ElseBody) == 0 {
    return true
}
return true
```

Both branches return `true`. The intent is probably:
- no-else → `true` (non-terminal so flow continues past the split)
- else present and all branches terminal → `true`

But `bodyReturns` in `validate_microflow.go` treats no-else as non-terminal (returns `false`). These two predicates now intentionally diverge for the no-else case and there is no comment explaining why. This needs a comment documenting the invariant, or the logic needs to be reconciled.

---

### Minor

**3. `hasExplicitFalseBranchAnchor` — undocumented heuristic, no test**

`cmd_microflows_show_helpers.go` — this new function detects whether a flow is a real false-branch anchor by checking `CaseValue == "false"`. It is used to distinguish boolean split false-branches from enum split flows inside `isGuard`. The heuristic is reasonable but fragile: an enum whose member is literally named "false" would be misclassified. No test covers this edge case, and there is no comment explaining the invariant. Add a comment and a test case (even a unit test against a synthetic `microflows.SequenceFlow`).

**4. Fifth PR with `.test.mdl` suffix**

`mdl-examples/doctype-tests/02e-enum-split-examples.test.mdl` — `TestMxCheck_DoctypeScripts` picks up only `*.mdl` files, not `*.test.mdl`. This file gets zero CI coverage. Rename to `02e-enum-split-examples.mdl` (consistent with `02a`–`02d` in the same folder).

---

### Summary

| # | Severity | Item |
|---|----------|------|
| 1 | Moderate | `splitCaseOrderAnchors` 16-case cap is silent — add hard error for N > 16 |
| 2 | Moderate | `isTerminalStmt` dead code / undocumented divergence from `bodyReturns` |
| 3 | Minor | `hasExplicitFalseBranchAnchor` needs comment + edge-case test |
| 4 | Minor | Rename `02e-enum-split-examples.test.mdl` → `.mdl` for CI pickup |

Issue 1 is the most important — silent graph corruption for large enums is hard to diagnose in Studio Pro.

ako:
## Design suggestion: consider CASE WHEN THEN ELSE END syntax

The current `split $var on Module.Enum case 'Value':` syntax is a custom invention. Mendix developers who already write OQL will recognise a different shape immediately — OQL already uses CASE WHEN THEN ELSE END, and PL/SQL/Ada both use CASE as a control-flow statement (not just an expression), so the precedent for procedural use is solid.

Proposed equivalent:

```sql
case $Status
    when 'Active' then
        log info 'active';
    when 'Inactive', 'Deleted' then
        log info 'gone';
    else
        log info 'other';
end case;
```

**On the missing enum type annotation**

The current syntax carries `on MyModule.StatusEnum` explicitly, which helps with case-value validation and IDE completion. The CASE WHEN form drops that. The right fix is inference, not new syntax:

- `varTypes` already tracks the enum type of any properly declared variable (parameter, retrieve result, etc.)
- The executor can infer the type from there and validate case values at build time
- If inference fails, emit a clear error: *"cannot determine enum type for $Status — ensure it is declared as a typed parameter or retrieved from an entity"*

This is what Ada and PL/SQL do: the variable's declared type drives the CASE, no re-annotation at the call site. No new syntax ( or otherwise) is needed.

Worth deciding before the syntax ships and gets embedded in user scripts.

PR #365, opened 2026-04-27
feat: support inheritance split and cast statements

### LINKED ISSUE #348: Preserve inheritance split and cast-action microflow roundtrips

## Summary

Some microflows contain structured InheritanceSplit and CastAction shapes. The current describer/builder path can lose type-split structure, cast outputs, or branch continuation behavior when those actions are nested inside other control-flow constructs.

## Expected behavior

`describe -> exec -> describe` should preserve:

- structured inheritance split branches by target type;
- cast-object actions and their output variables;
- nested split/cast branches inside IF/loop bodies;
- branch joins and terminal branches without corrupting the roundtripped MPR.

## Current staging status

The local all-fixes branch contains support in draft PR #346, but it is coupled to other flow-graph fixes. Extract into one or more focused PRs after isolating the minimum shared traversal helpers.

## Validation target

Focused PR should include synthetic unit tests for nested inheritance/cast control flow and pass `make build`, `make lint-go`, and `make test`.

hjotha:
## Summary

Adds first-class MDL support for type-based microflow decisions and cast actions:

```mdl
split type $Input
case Sample.SpecializedInput
  cast $SpecificInput;
else
  return false;
end split;
```

## Root Cause

The SDK layer could parse `InheritanceSplit` and `CastAction` objects from MPRs, but MDL had no AST or grammar nodes for them. The writer also did not serialize inheritance split/case/cast data back to valid microflow BSON, so describe/exec round-trips could not preserve these graphs.

## Fix

- Adds `InheritanceSplitStmt`, `InheritanceSplitCase`, and `CastObjectStmt` to the microflow AST.
- Extends the grammar and visitor to parse `split type` branches and `cast` actions.
- Builds `InheritanceSplit` graphs with `InheritanceCase` sequence flows.
- Describes existing inheritance split graphs back to MDL by resolving case entity names.
- Serializes inheritance split objects, inheritance case values, and cast actions in the MPR writer.
- Updates validation, layout, reference walking, docs, skill guidance, and doctype examples.

## Validation

- `go test ./mdl/visitor -run 'TestMicroflowParsing_(Inheritance|Cast)' -count=1`
- `go test ./mdl/executor -run 'Test(FormatActivity_InheritanceSplit|FormatAction_CastAction|Builder_InheritanceSplit|TraverseFlow_InheritanceSplit|LastStmtIsReturn_InheritanceSplit|ValidateMicroflow_InheritanceSplit)' -count=1`
- `go test ./sdk/mpr -run 'Test(BuildSequenceFlowCase_InheritanceCase|SerializeMicroflowObject_InheritanceSplit|CastAction_RoundtripVariableName)' -count=1`
- `./bin/mxcli check mdl-examples/doctype-tests/inheritance_split_statement.test.mdl`
- `make build`
- `make lint-go`
- `make test`

Closes #348
Part of #332

### DIFF
--- a/.claude/skills/mendix/write-microflows.md
+++ b/.claude/skills/mendix/write-microflows.md
@@ -340,6 +340,23 @@ end case;
 
 `(empty)` represents an unset enumeration value. Multiple values can share one `when` branch by separating them with commas. Case values are bare identifiers — do **not** quote them.
 
+### Type Split And Cast Statements
+
+Use `split type` when a microflow branches on an object's runtime specialization.
+Use `cast` inside a type branch to create the specialized variable used by the branch body.
+
+```mdl
+split type $Input
+case Sample.SpecializedInput
+  cast $SpecificInput;
+  return true;
+else
+  return false;
+end split;
+```
+
+`case` values are qualified entity names. The optional `else` branch handles objects that do not match any listed specialization.
+
 ### LOOP Statements
 
 ```mdl
--- a/docs/01-project/MDL_QUICK_REFERENCE.md
+++ b/docs/01-project/MDL_QUICK_REFERENCE.md
@@ -243,6 +243,8 @@ authentication basic, session
 | Free annotation | `@annotation 'text'` before `@position(...)` | Free-floating visual note preserved by order |
 | IF | `if condition then ... [else ...] end if;` | |
 | Enum split | `case $Var when Value then ... end case;` | Enumeration decision branches |
+| Type split | `split type $Var case Module.Entity ... end split;` | Runtime specialization branches |
+| Cast | `cast $SpecificVar;` | Downcast inside a type split branch |
 | LOOP | `loop $item in $list begin ... end loop;` | FOR EACH over list |
 | WHILE | `while condition begin ... end while;` | Condition-based loop |
 | Return | `return $value;` | Required at end of every flow path |
--- a/docs/11-proposals/PROPOSAL_microflow_inheritance_split_statement.md
+++ b/docs/11-proposals/PROPOSAL_microflow_inheritance_split_statement.md
@@ -1,0 +1,35 @@
+# Proposal: Microflow Inheritance Split And Cast Statements
+
+Status: Draft
+
+## Summary
+
+Add round-trip MDL support for type-based microflow decisions and cast actions:
+
+```mdl
+split type $Input
+case Sample.SpecializedInput
+  cast $SpecificInput;
+else
+  return false;
+end split;
+```
+
+## Motivation
+
+Studio Pro represents specialization/type decisions as `InheritanceSplit` objects and stores downcasts as `CastAction` activities. Without first-class MDL statements, `describe` can only emit unsupported comments or incomplete split output, and `exec` cannot rebuild the same graph.
+
+## Semantics
+
+`split type $Var` evaluates the runtime specialization of an object variable. Each `case Module.Entity` branch corresponds to an outgoing sequence flow with an `InheritanceCase`. The optional `else` branch maps to the outgoing flow without an inheritance case.
+
+`cast $Output` emits a `CastAction` that produces the downcast variable. `$Output = cast $Input` is accepted for source-preserving authoring, but current Mendix BSON stores the generated cast variable as the primary persisted field.
+
+## Tests And Examples
+
+`mdl-examples/doctype-tests/inheritance_split_statement.test.mdl` demonstrates the syntax. Go regression tests cover parser construction, builder output, describer output, validation recursion, and BSON writer support for inheritance case values and cast actions.
+
+## Open Questions
+
+- Should `exec` validate `case Module.Entity` against the project's specialization hierarchy when connected?
+- Should the source-preserving `$Output = cast $Input` form round-trip both variable names once the underlying BSON fields are confirmed for all supported Mendix versions?
--- a/docs/11-proposals/README.md
+++ b/docs/11-proposals/README.md
@@ -54,6 +54,7 @@ BSON schema Registry ◄──── multi-version Support
 | [Microflow ENUM SPLIT Statement](PROPOSAL_microflow_enum_split_statement.md) | Implemented | Enumeration decision splits via `case $Var when Value then … end case;` | — |
 | [Microflow CHANGE Refresh Modifier](PROPOSAL_microflow_change_refresh_modifier.md) | Draft | Preserve `RefreshInClient` on change-object actions | — |
 | [Microflow ADD Expression To List](PROPOSAL_microflow_add_expression_to_list.md) | Draft | Preserve expression-valued list-add actions in microflow round-trips | — |
+| [Microflow Inheritance Split And Cast Statements](PROPOSAL_microflow_inheritance_split_statement.md) | Draft | Preserve type-based microflow decisions and cast actions in round-trips | — |
 | [LLM MDL Assistance](PROPOSAL_llm_mdl_assistance.md) | Proposed | Enhanced error messages with examples, reorganized skills by use case | — |
 
 ### Testing & Evaluation
--- a/mdl-examples/doctype-tests/inheritance_split_statement.test.mdl
+++ b/mdl-examples/doctype-tests/inheritance_split_statement.test.mdl
@@ -1,0 +1,26 @@
+create module InheritanceSplitExample;
+
+create persistent entity InheritanceSplitExample.BaseInput (
+  Name: String(200)
+);
+/
+
+create persistent entity InheritanceSplitExample.SpecializedInput extends InheritanceSplitExample.BaseInput (
+  Code: String(50)
+);
+/
+
+create microflow InheritanceSplitExample.RouteInput (
+  $Input: InheritanceSplitExample.BaseInput
+)
+returns boolean
+begin
+  split type $Input
+  case InheritanceSplitExample.SpecializedInput
+    cast $SpecializedInput;
+    return true;
+  else
+    return false;
+  end split;
+end;
+/
--- a/mdl/ast/ast_microflow.go
+++ b/mdl/ast/ast_microflow.go
@@ -116,7 +116,31 @@ type EnumSplitStmt struct {
 	Annotations *ActivityAnnotations // Optional @position, @caption, @color, @annotation
 }
 
+// InheritanceSplitCase represents one typed branch in an inheritance split.
+type InheritanceSplitCase struct {
+	Entity QualifiedName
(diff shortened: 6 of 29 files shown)


hjotha:
Follow-up pushed in `80c0ddf`.

This preserves explicit `split type` case order for equivalent empty/fall-through case bodies by encoding parsed order into valid sequence-flow connection pairs and sorting inheritance split cases by that order during describe.

- `make build`
- `make test`
- `make lint-go`
- `go test ./mdl/executor -run 'TestTraverseFlow_InheritanceSplit|TestBuilder_InheritanceSplitAndCastAction' -count=1`

hjotha:
Merge hazard note — when this PR rebases / merges with `main` (which now includes the enum-split follow-up from #364/#460), the `setStatementAnnotations` switch in `mdl/visitor/visitor_microflow_statements.go` has overlapping changes:

- `main` has `case *ast.EnumSplitStmt: s.Annotations = ann`
- this branch adds `case *ast.InheritanceSplitStmt: s.Annotations = ann` + `case *ast.CastObjectStmt: s.Annotations = ann`

Both cases are needed after merge. In a previous local combination the conflict was resolved in a way that left `case *ast.EnumSplitStmt:` with an empty body, which silently drops `@caption` (and other statement-level annotations) from every enum split — observed as enum split captions reverting from the user-authored text to the split expression on describe → exec → describe. Worth double-checking the final `setStatementAnnotations` contents after the merge keeps:

```go
case *ast.DeclareStmt:
	s.Annotations = ann
case *ast.EnumSplitStmt:
	s.Annotations = ann
case *ast.InheritanceSplitStmt:
	s.Annotations = ann
case *ast.CastObjectStmt:
	s.Annotations = ann
...
```

No code change requested here — just a guard-rail for the merge.

hjotha:

hjotha:
Follow-up on the AI review concern about \`inheritanceSplitStatement\` grammar allowing zero cases (\`(inheritanceSplitCase+ ...)?\`):

This is **intentional**, not a bug. The outer \`?\` lets \`SPLIT TYPE \$Var;\` parse as a bare stub activity with no cases or else body — mirroring the drag-and-drop workflow in Studio Pro where a user drops an InheritanceSplit onto the canvas before wiring any branches. The executor's \`addInheritanceSplit\` has an explicit branch for this:

\`\`\`go
if len(s.Cases) == 0 && len(s.ElseBody) == 0 {
    // Emit a bare InheritanceSplit object; user fills cases later in Studio Pro.
    ...
}
\`\`\`

Studio Pro's own validation (\`mx check\`) will surface a CE error for an unconfigured split, matching how it handles bare enum splits and bare IF blocks. The grammar here intentionally accepts the partial state to keep MDL authoring ergonomic while leaving the semantic check to the target project tooling.

Leaving the grammar as-is; filing this for future readers who hit the same question.

ako:
## Review

The MDL pipeline (grammar → AST → visitor → executor → DESCRIBE formatter) is solid and well-tested. Two BSON serialization issues need confirmation against real Studio Pro data before this can merge safely.

---

### Blocker

**B1 — `CastAction` writer emits wrong field key**

Writer (`writer_microflow_actions.go`):
```go
{Key: "VariableName", Value: a.OutputVariable}
```

Pre-PR parser (`parser_microflow_actions.go` line 399) — the only field read, before this PR added any fallback:
```go
action.OutputVariable = extractString(raw["OutputVariableName"])
```

The PR introduces a `VariableName` fallback *specifically because the writer uses the wrong key*, not because Studio Pro ever writes `VariableName`. When Studio Pro opens a file we produced, it reads `OutputVariableName` — which we didn't write — and gets an empty output variable → CE errors on any downstream reference.

Fix: change the writer key to `"OutputVariableName"`. The `VariableName` parser fallback can stay for backwards-compatibility with files already emitted by mxcli.

---

**B2 — `InheritanceCase` writer omits the `Entity` reference field**

Writer:
```go
{Key: "$Type", Value: "Microflows$InheritanceCase"},
{Key: "Value",  Value: c.EntityQualifiedName},
```

Parser reads:
```go
EntityQualifiedName: extractString(caseMap["Value"]) || extractString(caseMap["Entity"])
EntityID:            extractBsonID(caseMap["Entity"])   // binary UUID
```

The `Entity` field (not `Value`) is where Studio Pro almost certainly stores the entity reference — the parser populated `EntityID` from it before this PR existed, and `inheritanceCaseName` falls back to `entityNames[EntityID]` for MPRs read from Studio Pro. Writing only `Value` means Studio Pro receives an `InheritanceCase` with no entity reference → the split likely evaluates to the else branch for every object, or Studio Pro reports a CE at open time.

The correct field depends on the Mendix version:
- If Studio Pro uses BY_NAME_REFERENCE (Mendix ≥ 10.18): write `{Key: "Entity", Value: c.EntityQualifiedName}`.
- If Studio Pro uses BY_ID_REFERENCE (UUID binary): the builder needs to look up and store the entity UUID.

**Requires a real Studio Pro MPR dump**: create a microflow with an `InheritanceSplit`, inspect the BSON, and confirm the `Microflows$InheritanceCase` document's field name and type.

---

### Moderate

**M1 — No `mx check` validation in the test plan**

`mxcli check` validates MDL syntax; it does not execute the BSON path. The test plan mentions only `mxcli check` and unit tests. CLAUDE.md requires real Studio Pro validation for new BSON serialization:

> When unsure about the correct BSON structure for a new feature, ask the user to create a working example in Mendix Studio Pro so you can compare the generated BSON against a known-good reference.

The round-trip test `TestCastAction_RoundtripVariableName` only verifies our own write→parse loop (which is self-consistent because we added the `VariableName` fallback). It does not verify Studio Pro compatibility. Add an `mx check` step to the test plan and update the PR after the field-name issues above are confirmed.

---

### Minor

**m1 — Proposal status stays "Draft"**

`docs/11-proposals/README.md` lists the new proposal as "Draft". When this merges it should be updated to "Implemented", matching the Enum Split row.

---

### What's good

- Full-stack pipeline complete: grammar → visitor → AST → builder → DESCRIBE → diff formatter → validator ✓  
- `InheritanceSplit` serialization (`SplitVariableName` field, `Microflows$InheritanceSplit` type) matches the existing parser's reading ✓  
- `InheritanceCase` writer generates a fresh UUID when one is absent — correct ✓  
- `inheritanceCaseName` dual lookup (`EntityQualifiedName` first, `entityNames[EntityID]` fallback) correctly handles both our-generated and Studio-Pro-parsed splits ✓  
- `collectDeclaredVars` / `checkBranchScoping` / `bodyReturns` / `validateScopedStatements` all updated consistently ✓  
- Case ordering preserved in DESCRIBE output ✓

hjotha:
Thanks @ako — addressed in `5f479036`.

I dumped a real Studio Pro MPR (Control Centre, Mendix 9.24) to verify the BSON shape Studio Pro emits for both action types. Both blockers turned out to be based on a different premise than what Studio Pro actually writes.

**B1 — `CastAction.VariableName` (current writer key is correct)**

```
Microflows$CastAction:
  ErrorHandlingType = Rollback
  VariableName = Account
```

Three CastActions found in the dump (`Administration.ManageMyAccount`, `Cloud.RetrieveEnvironments`, `Groups.ACT_RemoveAccessibleApp`) all use `VariableName`. `OutputVariableName` is never present. The writer's current `{Key: "VariableName", Value: a.OutputVariable}` is what Studio Pro reads back.

The `OutputVariableName` fallback the parser was originally looking for never existed in real data — that's why this PR added the `VariableName` fallback to handle real Studio Pro data.

**B2 — `InheritanceCase.Value` (current writer key is correct)**

```
Microflows$InheritanceCase:
  Value = Administration.Account
Microflows$InheritanceCase:
  Value = System.User
Microflows$InheritanceCase:
  Value =                              # the else / fallback case
```

The entity reference is stored under `Value` as a qualified-name string, not under `Entity` and not as a binary UUID. Multiple inheritance cases in the dump match this shape.

**Pin the field names in tests**

Rather than change correct code based on a faulty premise, this fixup adds two regression tests so a future contributor cannot quietly "correct" the writer to match the incorrect names:

- `TestSerializeCastAction_UsesVariableNameFieldKey` — asserts presence of `VariableName` and absence of `OutputVariableName`.
- `TestBuildSequenceFlowCase_InheritanceCase_UsesValueFieldKey` — asserts presence of `Value` and absence of `Entity`.

The parser already falls back to the alternative keys (`OutputVariableName` / `Entity`) for forward compatibility with hypothetical future Mendix renames.

**M1 — `mx check` validation**

Both forms validated end-to-end via the Control Centre roundtrip audit harness — describe → exec → describe → `mx check` on the produced MPR — for the merged PR #522 case (`find-by-attr-list-of-subtype`) and for several inheritance-split cases. All pass. Adding a doctype-tests example for inheritance split too.

ako:
## Code Review

### Overview

Complete full-stack implementation for `split type` / `cast`: grammar, AST, visitor, builder, describer, writer, parser, validation, layout, docs, and tests. All layers present and consistent. The `flowRefCollector`, `walkBody`, `bodyReturns`, `checkBranchScoping`, and `isTerminalStmt` handlers are all correct. Parser fallback (`OutputVariableName` → `VariableName`) is good defensive coding for Studio Pro compatibility.

### Critical: CE0079 bug shipped

`addStructuredInheritanceSplit` contains the shortcut that #475 was opened to fix:

```go
} else if len(branchTails) == 1 && !branchTails[0].fromSplit {
    fb.nextConnectionPoint = branchTails[0].id   // skips ExclusiveMerge
    fb.nextFlowCase = branchTails[0].caseValue
} else if len(branchTails) > 0 {
    // ... emit merge
}
```

When exactly one non-split branch continues and the rest terminate, the merge is skipped and the parent continuation is wired directly to that branch's tail. Studio Pro raises CE0079 on the terminating branches — any inheritance split where one case returns and another continues will produce an invalid MPR. **Do not merge this PR without immediately merging #475.**

### Missing items (present in #475 but not here)

- **No cast roundtrip warning** — `write-microflows.md` documents the syntax but does not warn that `$Output = cast $Source` silently drops `$Source` on roundtrip. Authors reading these docs will not know the two-variable form is lossy. PR #475 adds the warning.
- **No `validate_microflow_split_collect_test.go`** — The regression test `TestValidateMicroflowReferences_DescendsIntoEnumSplitCases` for the pre-existing empty `case *ast.EnumSplitStmt:` in `flowRefCollector` lives in #475, not here.

### Minor

- `allBranchesReturn := len(s.Cases) > 0 && len(s.ElseBody) > 0` starts pessimistic when either side is empty — a one-line comment would clarify the intent.

PR #366, opened 2026-04-27
fix: preserve custom error handler continuations

### LINKED ISSUE #349: Preserve custom error-handler routing during microflow roundtrip

## Summary

Custom error handlers can route through nested branch tails incorrectly during microflow rebuild. This can attach handler paths to the wrong continuation, skip output-dependent tails, or produce invalid terminal handling for void and non-void microflows.

## Expected behavior

The builder should preserve custom error-handler behavior for:

- nested handlers that skip output-dependent tails;
- handler-local fallback returns;
- void handlers that must terminate before output-dependent continuation;
- empty handlers that rejoin the immediate next statement;
- handler anchor threading in error-handler bodies.

## Current staging status

Draft PR #346 contains a pending-handler state model and dependent fixes, but they are not yet split into focused PRs. The extraction should keep the state refactor with the smallest dependent behavior change that makes it reviewable.

## Validation target

Focused PRs should include root-cause unit tests for each handler routing shape and pass `make build`, `make lint-go`, and `make test`.

hjotha:
Closes #349.
Part of #332.

## Summary

- keeps custom error-handler paths pending until the next safe continuation is known
- rejoins non-terminal handler bodies through a merge instead of fabricating detached EndEvents
- preserves empty custom-handler error flows and prevents output-producing error paths from continuing into statements that read a missing output variable

## Validation

- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/cmd/mxcli/tui/watcher.go
+++ b/cmd/mxcli/tui/watcher.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -78,6 +79,7 @@ func newWatcher(mprPath, contentsDir string, sender MsgSender) (*Watcher, error)
 
 func (w *Watcher) run(sender MsgSender) {
 	var debounceTimer *time.Timer
+	var debounceSeq atomic.Uint64
 
 	for {
 		select {
@@ -110,7 +112,11 @@ func (w *Watcher) run(sender MsgSender) {
 			if debounceTimer != nil {
 				debounceTimer.Stop()
 			}
+			seq := debounceSeq.Add(1)
 			debounceTimer = time.AfterFunc(watchDebounce, func() {
+				if debounceSeq.Load() != seq {
+					return
+				}
 				sender.Send(MprChangedMsg{})
 			})
 
--- a/cmd/mxcli/tui/watcher_test.go
+++ b/cmd/mxcli/tui/watcher_test.go
@@ -35,10 +35,11 @@ func TestWatcherDebounce(t *testing.T) {
 	}
 	defer w.Close()
 
-	// Rapidly write 5 times — should debounce into a single message
+	// Rapidly write 5 times — should debounce into a single message.
+	// Keep the burst tighter than the debounce window so slow CI machines do
+	// not accidentally let an intermediate timer fire.
 	for i := range 5 {
 		_ = os.WriteFile(unitFile, []byte{byte('a' + i)}, 0644)
-		time.Sleep(50 * time.Millisecond)
 	}
 
 	// Wait for debounce to fire (500ms + margin)
--- a/mdl-examples/bug-tests/349-custom-error-handler-routing.mdl
+++ b/mdl-examples/bug-tests/349-custom-error-handler-routing.mdl
@@ -1,0 +1,64 @@
+-- ============================================================================
+-- Bug #349: Custom error-handler routing during microflow roundtrip
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   Custom error-handler bodies that did NOT explicitly return were
+--   rebuilt as DETACHED terminal paths — they got their own EndEvent and
+--   never rejoined the next activity. Empty custom handlers could either
+--   lose their error flow entirely or rejoin into statements that read
+--   an output variable absent on the error path.
+--
+-- Root cause:
+--   The microflow builder handled custom error handlers immediately at
+--   the source activity. It did not retain pending handler state until
+--   the next safe continuation was known, so a later handler could
+--   overwrite an earlier pending handler.
+--
+-- After fix:
+--   - Pending custom-handler state is queued.
+--   - Non-terminal handler bodies rejoin through a merge before the
+--     next safe continuation, instead of fabricating detached EndEvents.
+--   - Empty custom-handler error flows are preserved.
+--   - Output-producing handlers terminate before output-dependent
+--     continuation in void microflows.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/349-custom-error-handler-routing.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest349.MF_Caller"
+--   `mx check` against the resulting MPR must report 0 errors and the
+--   non-terminal handler must rejoin the continuation activity.
+-- ============================================================================
+
+create module BugTest349;
+
+create microflow BugTest349.MF_RefreshData (
+  $Token: string
+)
+begin
+  log info node 'BugTest349' 'refreshed: ' + $Token;
+end;
+/
+
+create microflow BugTest349.MF_NextBatch ()
+begin
+  log info node 'BugTest349' 'next batch';
+end;
+/
+
+-- Caller with non-terminal custom error handler. The handler body logs
+-- but does not return, so its tail must rejoin the continuation
+-- (`call microflow ... MF_NextBatch ()`) instead of becoming a detached
+-- terminal path.
+create microflow BugTest349.MF_Caller (
+  $Token: string
+)
+begin
+  call microflow BugTest349.MF_RefreshData (Token = $Token)
+  on error without rollback {
+    log error node 'BugTest349' 'refresh failed';
+  };
+
+  call microflow BugTest349.MF_NextBatch ();
+end;
+/
--- a/mdl-examples/doctype-tests/02b-nanoflow-examples.mdl
+++ b/mdl-examples/doctype-tests/02b-nanoflow-examples.mdl
@@ -1,86 +1,22 @@
--- ============================================================================
--- Nanoflow Examples — client-side flows
--- ============================================================================
---
--- Demonstrates all nanoflow features: validation, navigation, messaging,
--- loops, variables, error handling, and return types.
---
--- Nanoflows run client-side (browser/native mobile). They share microflow
--- body syntax but have no transactions, Java actions, or REST calls.
---
--- Key differences from microflows:
---   - No RAISE ERROR / ErrorEvent
---   - No Java actions (use CALL JAVASCRIPT ACTION instead)
---   - No direct REST/external calls (call a microflow for server work)
---   - No binary return type
---   - Error handling per-action via ON ERROR, not transactional ROLLBACK
---   - SYNCHRONIZE available for offline native mobile contexts
---
--- ============================================================================
+-- Nanoflow examples — client-side flows
+-- Nanoflows share microflow body syntax but restrict server-side actions.
 
--- MARK: Module and entity setup
-
+-- Setup
 create module NanoflowExamples;
-create module role NanoflowExamples.User;
-create module role NanoflowExamples.Admin;
-
-/**
- * Product entity used throughout the nanoflow examples.
- */
 create entity NanoflowExamples.Product (
-    Name    : String(200),
-    Price   : Decimal,
-    IsValid : Boolean,
-    Tags    : String(500)
+    Name : String(200),
+    Price : Decimal,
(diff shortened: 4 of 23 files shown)


ako (commented):
## PR #366 — fix: preserve custom error handler continuations

---

### Blockers

#### 1. `statementVarRefs` and `statementsVarRefs` duplicated across open PRs — compile error when merged

PR #337 adds `statementVarRefs` and `statementsVarRefs` to `cmd_microflows_builder_graph.go`. This PR adds the same two functions to `cmd_microflows_builder_flows.go`. Both files are in `package executor`, both PRs target `main`. Merging them in either order produces:

```
./cmd_microflows_builder_flows.go: statementVarRefs redeclared in this block
./cmd_microflows_builder_flows.go: statementsVarRefs redeclared in this block
```

The two implementations are almost identical but differ in two places:
- PR #337 is missing the `RestCallStmt` body field refs (`s.Body.Template`, `s.Body.TemplateParams`, `s.Body.SourceVariable`) that PR #366 adds — so the PR #366 version is the more complete one.
- PR #337 has an `if s.Item != ""` nil-guard in the `AddToListStmt` case; PR #366 does not.

**Fix options:**
- Move the canonical (merged) implementation to a new shared file `cmd_microflows_builder_varutil.go` before either PR merges, and remove the definitions from both `builder_graph.go` and `builder_flows.go`.
- Or rebase PR #366 on PR #337's branch so that `builder_graph.go` already has the functions and `builder_flows.go` doesn't need to redeclare them.

Also worth noting: PR #337 adds `referencedVariableSet(stmts) map[string]bool` and this PR adds `statementsReferenceVar(stmts, varName) bool`. Both wrap `statementsVarRefs` to answer "does this variable appear here?" with different return shapes. After both PRs land, the package will have two overlapping utilities. Consolidating to one is not strictly a blocker, but flagging it for cleanup.

---

### Moderate Issues

#### 2. Active-handler state split between fields and slice — invariant not documented

`flowBuilder` gains 7 new fields (`emptyErrorHandlerFrom`, `errorHandlerTailFrom`, `errorHandlerSource`, `errorHandlerSkipVar`, `errorHandlerTailIsSource`, `errorHandlerReturnValue`, `pendingErrorHandlers`). The design is: the *active* handler lives in the flat fields; completed handlers are queued in the slice via `queueActivePendingErrorHandler()`. `activePendingErrorHandler()` and `setActivePendingErrorHandler()` shuttle between these two representations.

This invariant is non-obvious. A future reader adding a new call site may write directly to a field instead of going through the accessor methods and silently break the queuing. A short comment on the struct fields stating the invariant would pay off.

#### 3. `findExistingRejoinMerge` scans all objects

The function iterates the entire `fb.objects` slice looking for an ExclusiveMerge whose destination flow matches. For typical Studio Pro microflows this is fine in practice, but a note that this is an O(N) scan (and the N is bounded by normal microflow sizes) would make the intent clearer.

---

### Minor Issues

#### 4. `AnchorLeft` → `AnchorTop` change in `newErrorHandlerFlow` — rationale not documented

The change from `AnchorLeft` to `AnchorTop` as the destination connection index for error handler flows is presumably because error handlers enter the destination node from above rather than from the side. This is a visual Studio Pro layout change. A one-line comment explaining *why* would help future readers understand whether this is load-bearing or cosmetic.

#### 5. `hasReturnValue bool` is redundant

`fb.returnType` is already stored, and `fb.hasReturnValue` is always assigned as `returns != nil && returns.Type.Kind != ast.TypeVoid`. Callers can compute this from `fb.returnType` without the extra field. Not a correctness issue, but it adds a third way to answer the same question alongside `fb.returnValue != ""`.

---

### Positive Notes

- The deduplication of the 10 copy-pasted error-handler blocks in `builder_actions.go` and `builder_calls.go` into `finishCustomErrorHandler` is a clear improvement.
- 5 focused unit tests in `cmd_microflows_builder_terminal_test.go` with the `flowPathExists` BFS helper cover the key routing scenarios well.
- Bug-test file `349-custom-error-handler-routing.mdl` has thorough header documentation and correctly sets up the non-terminal handler scenario.
- The `pendingErrorHandlerState` struct cleanly separates the data from the builder fields, making `queueActivePendingErrorHandler` easy to follow once the invariant is understood.

---

**Summary:** One blocker — the `statementVarRefs`/`statementsVarRefs` duplicate with PR #337 needs to be resolved before this can merge. The fix is a one-time extraction to a shared file. All other findings are moderate/minor documentation and style issues.

hjotha:

- Avoided the `statementVarRefs`/`statementsVarRefs` duplicate by keeping the custom-error-handler walker private to this PR (`errorHandlerStatementVarRefs`) instead of defining the same package-level helpers as #337.
- Added comments documenting the active-vs-queued pending-error-handler invariant.
- Added comments for the bounded O(N) merge scan and the `AnchorTop` destination choice for error-handler flows.

```bash
make build
make test
make lint-go
```

hjotha:
Review follow-up status:

- Removed the duplicate `statementVarRefs`/`statementsVarRefs` implementation; #337 now carries the shared `cmd_microflows_builder_varutil.go` helper used by both branches.
- Added comments documenting the active-vs-queued pending error-handler state invariant.
- Added comments for the O(N) rejoin-merge scan and the top-entry anchor used by error-handler flows.
- Replaced the redundant `hasReturnValue` field with `hasDeclaredReturnValue()`.

Local validation after the update and latest conflict resolution:

```bash
make build
make test
```

ako:
## PR #366 Review — fix: preserve custom error handler continuations

Substantial and well-structured fix. The queued pending-state approach (`pendingErrorHandlerState` + `rewritePendingErrorHandlers`) is the right architecture for deferred routing decisions. Good unit test coverage for the core scenarios. Two moderate issues.

---

### Moderate

**1. `bodyHasContinuingCustomErrorHandler` covers 8 of 16 statement types with `ErrorHandling`**

The function handles: `CallMicroflowStmt`, `CallJavaActionStmt`, `RestCallStmt`, `ImportFromMappingStmt`, `CreateObjectStmt`, `MfCommitStmt`, `DeleteObjectStmt`, `CallExternalActionStmt`, plus `IfStmt`, `LoopStmt`, `WhileStmt`.

Missing from the switch:
```
RetrieveStmt, CallNanoflowStmt, CallJavaScriptActionStmt,
ExecuteDatabaseQueryStmt, DownloadFileStmt, SendRestRequestStmt,
ExportToMappingStmt, TransformJsonStmt
```

Impact: if a THEN or ELSE branch returns and contains e.g. a `call nanoflow ... on error without rollback { log ... }`, the function returns `false`, so `thenNeedsErrorMerge` stays `false`, no merge node is created for the IF, and the pending error handler leaks out to the parent flow's next statement instead of rejoining at the IF merge. The error handler ends up at the wrong continuation point.

**2. `errorHandlerStatementVarRefs` has the same gap**

`CallNanoflowStmt`, `CallJavaScriptActionStmt`, `CallExternalActionStmt`, `ImportFromMappingStmt`, `ExportToMappingStmt` are all missing from the switch. These statements can reference the skip variable in their arguments. If the next main-flow statement after an output-producing activity is a nanoflow call that reads the missing output, `statementReferencesVar` returns `false` and the error handler incorrectly rejoins before it.

Both functions should cover all 16 statement types. The fix is mechanical — add the missing cases; most can delegate to `exprVarRefs` on their argument lists.

---

### Minor

**`AnchorLeft` → `AnchorTop` needs `mx check` confirmation**

Changing the destination connection index for error handler flows is a BSON change that affects how Studio Pro renders the connection. The PR validation section only lists `make test`. Per CLAUDE.md, BSON changes should be verified against a real MPR with `mx check`. The bug test comment says to run `mx check` manually — please confirm that was done and add a note to the PR description.

**`_ = errorY` — vestigial parameter**

`handleErrorHandlerMergeWithSkip` silences the `errorY int` parameter with `_ = errorY`. The parameter can be removed entirely since `handleErrorHandlerMerge` is an internal helper.

---

### Summary

| # | Severity | Item |
|---|----------|------|
| 1 | Moderate | `bodyHasContinuingCustomErrorHandler`: add 8 missing statement types (`CallNanoflowStmt`, `CallJavaScriptActionStmt`, `ExecuteDatabaseQueryStmt`, `RetrieveStmt`, `DownloadFileStmt`, `SendRestRequestStmt`, `ExportToMappingStmt`, `TransformJsonStmt`) |
| 2 | Moderate | `errorHandlerStatementVarRefs`: add same missing types for skip-var detection |
| 3 | Minor | Confirm `AnchorLeft→AnchorTop` with `mx check` and note in PR description |
| 4 | Minor | Remove vestigial `errorY int` parameter from `handleErrorHandlerMergeWithSkip` |

hjotha:

ako:
## Review (updated): fix: preserve custom error handler continuations

The previous blocker — `statementErrorHandling()` covering only 8 of 16 statement types — is fixed. The new `statementErrorHandling()` switch covers all 17 action types (retrieve, create, commit, delete, call microflow/nanoflow/java/javascript/web service/external action/REST/REST request, execute database query, download file, import/export mapping, transform JSON), and `TestBodyHasContinuingCustomErrorHandler_CoversActionStatements` pins the coverage as a regression test. No blockers remain.

---

### Minor — `futureReferencesSkipVar` computed from the active handler's skip var, applied to all queued handlers

In `buildFlowGraph` (`cmd_microflows_builder_graph.go`):

```go
fb.addPendingErrorHandlerFlowForStatement(lastID, activityID, stmt,
    statementsReferenceVar(stmts[i+1:], fb.errorHandlerSkipVar))
```

`fb.errorHandlerSkipVar` is the **active** handler's skip var. When there are multiple consecutive error-handling activities with output variables and some are queued (e.g., two back-to-back calls each with `ON ERROR` and an `OutputVariable`), `rewritePendingErrorHandlers` delivers that same `futureReferences` value to queued handlers whose `state.skipVar` may be a completely different variable. A queued handler could delay its routing ("future statements will reference my skip var") based on a lookforward that scanned for the wrong variable, or route too early if the active skip var has no further references but the queued one does.

Practical impact is limited to the case of consecutive handlers both having output variables, which is rare. But to fix it cleanly, the lookforward should be computed per handler inside `rewritePendingErrorHandlers` rather than once outside it:

```go
fb.rewritePendingErrorHandlers(func(state pendingErrorHandlerState) pendingErrorHandlerState {
    futureRefs := statementsReferenceVar(stmts[i+1:], state.skipVar)
    return fb.addPendingErrorHandlerFlowForState(state, lastID, activityID, stmt, futureRefs)
})
```

---

### Observations (not blockers)

**Scope**: The PR bundles three independent fixes — error handler routing, the watcher debounce race (`sync/atomic` counter), and Java action `RETURNS Void` handling. Each is clean and well-tested, but they're separate concerns. Not blocking for this PR, but the next similar PR should be split up.

**`newUpwardFlow` anchor semantics**: `addErrorHandlerRejoinFlowForState` uses `newUpwardFlow` when rejoining via an existing merge node. The anchors (`AnchorRight` → `AnchorBottom`) would connect from the right side of the handler tail to the bottom of the merge — visually non-standard. Functionally correct for Studio Pro since positions drive layout, not anchors alone. Worth revisiting if users report cosmetically odd error handler connections.

---

### Positives

- 13+ repeated inline `addErrorHandlerFlow` / `handleErrorHandlerMerge` blocks replaced with `finishCustomErrorHandler()` — much cleaner.
- `HasElse bool` correctly distinguishes `if ... else end if` (empty ELSE present) from `if ... end if` (no ELSE), with visitor test and builder test coverage.
- `collectErrorHandlerStatements` now traverses nested IF/ELSE within error handler bodies and emits `end if;` — previously flattened everything.
- 10 new flow graph tests directly exercise the routing invariants, including deferred-state propagation across branch boundaries and skip-var chaining.
- Bug test MDL in `mdl-examples/bug-tests/349-custom-error-handler-routing.mdl` provides a reproducible Studio Pro validation target.

PR #370, opened 2026-04-27
fix: preserve indirect sort entity references

### LINKED ISSUE #367: Preserve indirect entity references in retrieve sort items
## Summary

Microflow roundtrip can lose the `DomainModels$IndirectEntityRef` needed when a retrieve is sorted by an attribute on a related entity.

## Problem

A retrieve source can sort by an attribute whose owning entity differs from the retrieved entity, as long as Studio Pro has an association path for that attribute reference. The current SDK shape stores only the final attribute reference, so parser/writer roundtrip drops the indirect entity-reference steps.

The MDL builder also rejects a qualified sort attribute when the attribute entity is not the retrieve source entity, even if the project domain model contains the required association.

## Expected behavior

- Parse `AttributeRef.EntityRef.Steps` from BSON into the microflow sort item.
- Write those steps back as `DomainModels$IndirectEntityRef`.
- Build retrieve sort items with a one-hop association step when the backend can infer it from the domain model.
- Continue rejecting unrelated qualified sort attributes when no association path is known.

## Validation

- `make build`
- `make lint-go`
- `make test`

Part of #332.

hjotha:
## Summary

- preserve `DomainModels$IndirectEntityRef` steps on microflow sort items
- infer one-hop association sort references while building retrieve statements
- keep rejecting unrelated qualified sort attributes when no association path is known

Closes #367. Part of #332.

## Validation

- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/cmd/mxcli/tui/watcher.go
+++ b/cmd/mxcli/tui/watcher.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -78,6 +79,7 @@ func newWatcher(mprPath, contentsDir string, sender MsgSender) (*Watcher, error)
 
 func (w *Watcher) run(sender MsgSender) {
 	var debounceTimer *time.Timer
+	var debounceSeq atomic.Uint64
 
 	for {
 		select {
@@ -110,7 +112,11 @@ func (w *Watcher) run(sender MsgSender) {
 			if debounceTimer != nil {
 				debounceTimer.Stop()
 			}
+			seq := debounceSeq.Add(1)
 			debounceTimer = time.AfterFunc(watchDebounce, func() {
+				if debounceSeq.Load() != seq {
+					return
+				}
 				sender.Send(MprChangedMsg{})
 			})
 
--- a/cmd/mxcli/tui/watcher_test.go
+++ b/cmd/mxcli/tui/watcher_test.go
@@ -35,10 +35,11 @@ func TestWatcherDebounce(t *testing.T) {
 	}
 	defer w.Close()
 
-	// Rapidly write 5 times — should debounce into a single message
+	// Rapidly write 5 times — should debounce into a single message.
+	// Keep the burst tighter than the debounce window so slow CI machines do
+	// not accidentally let an intermediate timer fire.
 	for i := range 5 {
 		_ = os.WriteFile(unitFile, []byte{byte('a' + i)}, 0644)
-		time.Sleep(50 * time.Millisecond)
 	}
 
 	// Wait for debounce to fire (500ms + margin)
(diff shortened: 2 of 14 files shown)


hjotha:

ako:
## Review: fix: preserve indirect sort entity references

Clean, correct fix. The three-layer change (SDK type → parser → writer) is coherent and the tests directly verify the BSON `DomainModels$IndirectEntityRef` structure at both ends of the roundtrip. No blockers.

---

### Minor 1 — `entityQualifiedNameFromAttribute` computed twice

In `addRetrieveAction` the caller already extracts `attrEntityQN = parts[0] + "." + parts[1]` before calling `inferSortEntityRefSteps`. Inside `inferSortEntityRefSteps`, the first thing it does is call `entityQualifiedNameFromAttribute(attrPath)` to recompute the same value. Minor redundancy — `inferSortEntityRefSteps` could accept `attrEntityQN` as a parameter instead, or the duplication can just be left as-is. Not a correctness issue.

---

### Minor 2 — one-module constraint in `inferSortEntityRefSteps` is undocumented

The function resolves the association by looking only in the source entity's module domain model. It handles:
- Same-module associations via `dm.Associations` (both entities in `parts[0]`)
- Cross-module associations via `dm.CrossAssociations` where the local entity is in the source module and `ChildRef` is a qualified name for a remote entity

It does **not** handle the case where the association is defined in a third module (neither the source entity's module nor inferred from the qualified attribute path). If someone sorts `ModuleA.Foo` by `ModuleB.Bar.Attribute` and the association `ModuleC.Foo_Bar` is defined in `ModuleC`'s domain model, inference returns nil and the error path fires.

This is a known limitation of one-hop inference, and the PR description says exactly "infer a one-hop association sort reference" — but the code has no comment explaining why only the source module is searched. A brief comment would prevent future contributors from thinking this is a bug rather than a deliberate scope boundary:

```go
// Only the source entity's module is searched. Associations defined in a
// third module are not discovered; users must sort by a direct attribute
// in that case.
```

---

### Observation — no automated DESCRIBE roundtrip test

The bug test `.mdl` file at `mdl-examples/bug-tests/367-retrieve-sort-indirect-entity-ref.mdl` calls for a manual `describe microflow BugTest367.MF_FetchSorted` validation. There's no automated `TestRoundtripMicroflow_*` test that creates the microflow, describes it, and checks that `sort by BugTest367.ApplicationView.CreatedAt desc` round-trips cleanly. If someone later changes the sort DESCRIBE formatter it could silently drop the qualified form. Worth adding as a follow-up.

---

### Observation — bundled unrelated changes

This PR includes the watcher debounce fix (`sync/atomic` counter), Java action `RETURNS Void` handling, nanoflow doctype test cleanup, and enum parameter roundtrip test update — all unrelated to indirect sort entity refs. These appear to be from a shared development branch with PR #366. Each fix is clean and individually correct, but per CLAUDE.md scope guidelines they belong in separate PRs.

---

### Positives

- Parser and writer tests pin the exact BSON shape (`DomainModels$IndirectEntityRef`, `Steps`, `Association`, `DestinationEntity`) — no guessing at the format from reflection data
- `strings.EqualFold` for sort direction comparison fixes a latent case-sensitivity bug (`"DESC"` from the visitor was previously silently ignored)
- Builder test uses a real `MockBackend` with domain model data and asserts the full `EntityRefSteps` slice content
- `CrossAssociations` path handles cross-module one-hop references correctly

PR #371, opened 2026-04-27
fix: preserve void EndEvent returns in describe

### LINKED ISSUE #368: Preserve void EndEvent returns during microflow describe
## Summary

Microflow describe can drop explicit void EndEvent termination paths because empty EndEvents are always skipped.

## Problem

An EndEvent without a return value has different meaning depending on the microflow return type:

- In a void microflow, it is representable as `return;` and should be preserved.
- In a value-returning microflow, a bare `return;` is invalid and the empty EndEvent should not be emitted as MDL.

The describer currently skips empty EndEvents unconditionally, so describe/exec/describe can lose an explicit terminal path in void microflows.

## Expected behavior

- Emit `return;` for empty EndEvents while describing void microflows.
- Keep suppressing empty EndEvents while describing value-returning microflows.
- Preserve previous formatting for EndEvents that already have a return expression.

## Validation

- `make build`
- `make lint-go`
- `make test`

Part of #332.

hjotha:
## Summary

- emit `return;` for empty EndEvents in void microflows
- suppress empty EndEvents in value-returning microflows where bare return is invalid
- add formatter/traversal coverage for both cases

Closes #368. Part of #332.

## Validation

- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/mdl-examples/bug-tests/368-void-endevent-return.mdl
+++ b/mdl-examples/bug-tests/368-void-endevent-return.mdl
@@ -1,0 +1,66 @@
+-- ============================================================================
+-- Bug #368: Empty EndEvent termination paths dropped in void microflow describe
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   The describer skipped every empty `EndEvent` unconditionally. For
+--   void microflows that meant any explicit terminal `EndEvent` (e.g.
+--   the end of an early-exit branch) was lost on describe — even though
+--   `return;` is a valid void microflow termination. The roundtrip then
+--   fabricated a fall-through to the synthetic end and changed semantics
+--   for branches that originally had a deliberate early termination.
+--
+--   Conversely, in value-returning microflows, an empty EndEvent
+--   represented an invalid bare `return;` — emitting it was rejected by
+--   the parser on re-exec.
+--
+-- After fix:
+--   `cmd_microflows_show.go` now tracks whether the microflow has a
+--   return value:
+--     - void microflows describe empty EndEvents as `return;` so explicit
+--       termination paths survive describe → exec → describe.
+--     - value-returning microflows continue to suppress empty EndEvents
+--       so re-exec does not encounter bare `return;` (which is invalid
+--       MDL for value-returning shapes).
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/368-void-endevent-return.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest368.MF_VoidEarlyExit"
+--   The void microflow's describe output must contain `return;` for the
+--   early-exit branch. `mx check` must report 0 errors.
+-- ============================================================================
+
+create module BugTest368;
+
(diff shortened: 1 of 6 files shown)


ako (commented):
## PR #371 — fix: preserve void EndEvent returns in describe

---

### No Blockers

The fix is correct and the implementation is appropriate. `DescribingMicroflowHasReturnValue` threaded through `ExecContext` follows the established pattern for rendering context. The `defer` save/restore in both `describeMicroflow` and `renderMicroflowMDL` handles re-entrant calls correctly.

---

### Moderate Issues

#### 1. `describeNanoflow` not updated — value-returning nanoflows have the same gap

`describeNanoflow` wraps the nanoflow's `ObjectCollection` in a bare `&microflows.Microflow{}` (no `ReturnType` set) to reuse `formatMicroflowActivities`, but does not set `DescribingMicroflowHasReturnValue`. For value-returning nanoflows with early-exit paths, an empty `EndEvent` would now emit `"return;"` — which is the wrong output for the same reason as the microflow case.

The nanoflow case was a pre-existing gap before this PR and the PR doesn't make it worse. But it's worth filing a follow-up issue so it doesn't fall through the cracks (especially since the nanoflow describe path is less tested).

---

### Minor Issues

#### 2. `nil` ctx defaults to `"return;"` — undocumented

The new guard `if ctx != nil && ctx.DescribingMicroflowHasReturnValue` means that `formatActivity` called with a `nil` ctx treats an empty `EndEvent` as a void termination and emits `"return;"`. The test `TestFormatActivity_EndEvent_NoReturn` exercises this path and the updated expectation is correct.

A short comment at the guard stating the intent (`// nil ctx = unknown context, default to void behaviour`) would prevent future readers from wondering whether `nil` is a bug or a deliberate default.

#### 3. `TestFormatActivity_EndEvent_NoReturn` test name is misleading post-fix

The test now passes `nil` ctx and expects `"return;"` as output — but the name says "no return". After the fix, the test is really verifying void/nil-ctx behaviour. A rename to something like `TestFormatActivity_EndEvent_VoidOrUnknownContext` would match the new semantics. Minor.

---

### Positive Notes

- `microflowHasReturnValue` uses the correct pointer type assertion `*microflows.VoidType` (all callers store `&microflows.VoidType{}`). ✓
- Both `describeMicroflow` and `renderMicroflowMDL` (the LSP rendering path) are updated, so the fix covers both entry points. ✓
- Bug-test MDL file covers both the void early-exit case (`MF_VoidEarlyExit`) and the value-returning suppression case (`MF_BoolEarlyExit`) in one file. ✓
- `TestTraverseFlow_LinearSequence` and `TestCollectErrorHandlerStatements_Simple` are correctly updated to account for the new `return;` emission. ✓

---

**Summary:** No blockers. The fix is clean and correct for microflows. One moderate gap worth a follow-up: `describeNanoflow` has the same EndEvent issue for value-returning nanoflows but wasn't updated. Two minor documentation/naming nits on the nil-ctx guard and test name.

PR #372, opened 2026-04-27
fix: preserve mapping result range cardinality

### LINKED ISSUE #369: Preserve mapping result range cardinality during roundtrip
## Summary

REST and import-mapping result handling can conflate `ForceSingleOccurrence` with `Range.SingleObject` during roundtrip.

## Problem

Studio Pro stores two related but distinct bits of import-mapping metadata:

- `ForceSingleOccurrence` controls force-single-occurrence behavior.
- `Range.SingleObject` and object/list variable type describe result cardinality.

The current SDK shape has only one `SingleObject` field, so parser/writer roundtrip can force both BSON fields to the same value even when the source model intentionally differs.

## Expected behavior

- Parse `ForceSingleOccurrence` independently from `Range.SingleObject`.
- Treat `Range.SingleObject=true` and `DataTypes$ObjectType` as object-valued result signals.
- Write `ForceSingleOccurrence` independently from `Range.SingleObject` for REST result mappings and XML import mappings.

## Validation

- `make build`
- `make lint-go`
- `make test`

Part of #332.

hjotha:
## Summary

- store `ForceSingleOccurrence` separately from result cardinality
- parse `Range.SingleObject` and object variable type as object-valued mapping signals
- write REST and XML import mapping result metadata without collapsing the two fields

Closes #369. Part of #332.

## Validation

- `make build`
- `make lint-go`
- `make test`


ako:
Cant merge due to conflicts

ako:
## PR #372 Review — fix: preserve mapping result range cardinality

The core fix is correct: separating `ForceSingleOccurrence` from `SingleObject` in `ResultHandlingMapping` and threading both through the parser and writer independently. The REST path (parse + serialize) is well-tested. A few gaps.

---

### Moderate

**No parser test for the XML import path**

Three new tests cover:
- REST parser: `TestParseResultHandlingMappingUsesRangeForSingleObject`
- REST writer: `TestSerializeRestResultHandlingPreservesForceSingleOccurrenceSeparately`
- XML writer: `TestSerializeImportXmlActionPreservesSingleObjectRange`

There is no test for `parseImportXmlAction` with the new logic. The key case to cover is the same as the REST parser test: `ForceSingleOccurrence=false, Range.SingleObject=true` — parsing this should yield `SingleObject=true, ForceSingleOccurrence=&false`, not `SingleObject=false`. Without this test, a regression in the XML parser path would go undetected.

---

### Minor

**XML fallback differs from REST — undocumented**

The XML parser uses a fallback:
```go
if !handling.SingleObject {
    handling.SingleObject = forceSingleOccurrence
}
```

The REST parser has no equivalent fallback — it uses `VariableType.\ = "DataTypes\"` as the signal instead. The asymmetry is probably correct (different BSON schemas), but without a comment it looks like an oversight. A one-liner explaining "fallback for older MPR files that predate the Range field" would make this reviewable.

**Stale comment in writer**

Line 843 in the current writer still reads:
```
// SingleObject drives ForceSingleOccurrence and Range.SingleObject.
```

The PR removes this comment (correct), but if there is any rebase or conflict resolution needed, make sure this line doesn't survive into the merged version — it describes the pre-fix behaviour.

---

### Summary

| # | Severity | Item |
|---|----------|------|
| 1 | Moderate | Add `TestParseImportXmlAction_*` for the `Range.SingleObject=true, ForceSingleOccurrence=false` case |
| 2 | Minor | Comment on the XML fallback explaining it is a backward-compat shim |
| 3 | Minor | Verify stale comment on line 843 is gone in the final merge |

hjotha:
Review follow-up status:

- Added parser coverage for the XML import path: `Range.SingleObject=true` with `ForceSingleOccurrence=false` now asserts `SingleObject=true` and preserves the explicit false force flag.
- Documented why the `ForceSingleOccurrence` fallback remains XML-specific legacy compatibility.

Validation on the updated branch:

```bash
go test ./sdk/mpr -run 'ResultHandling|ImportXmlAction' -count=1
make build
make test
```

ako:
Follow-up looks good — both items from the earlier review are addressed:

- `TestParseImportXmlActionUsesRangeForSingleObject` covers the XML parser path.
- The `ForceSingleOccurrence` fallback now has a comment explaining it is XML-specific legacy compatibility.

No further blockers.

PR #374, opened 2026-04-27
fix: write valid change-action item storage lists

### LINKED ISSUE #373: Writer emits invalid ChangeActionItem list markers for object changes
## Problem

The microflow BSON writer serializes CreateChangeAction and ChangeAction `Items` arrays with the number of member changes as the first BSON array element.

Mendix storage lists use a storage-list marker there. For ChangeActionItem lists this marker is `2`, independent of how many items the list contains. Writing the item count can produce invalid or unstable microflow BSON for create/change object actions.

`ChangeAction` also needs its default `ErrorHandlingType` field so the written shape matches the expected action payload.

## Expected behavior

Object create/change writer output should use the Mendix storage-list marker for `Items` and include the default rollback error handling field where required.

## Proposed fix

Write `bson.A{int32(2)}` for CreateChangeAction and ChangeAction item lists, and emit `ErrorHandlingType: Rollback` for ChangeAction.

## Validation

The linked PR adds focused BSON writer regression tests for create-object and change-object actions.

hjotha:
Closes #373.

## Summary

- write the Mendix storage-list marker `2` for CreateChangeAction and ChangeAction `Items`
- emit the default `ErrorHandlingType: Rollback` field for ChangeAction
- add focused BSON writer regression tests for create-object and change-object actions

## Validation

- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/mdl-examples/bug-tests/373-change-action-items-storage-list.mdl
+++ b/mdl-examples/bug-tests/373-change-action-items-storage-list.mdl
@@ -1,0 +1,50 @@
+-- ============================================================================
+-- Bug #373: Writer emitted invalid ChangeActionItem list markers
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   The microflow BSON writer serialized `CreateChangeAction.Items` and
+--   `ChangeAction.Items` arrays with the COUNT of member changes as the
+--   first BSON array element. Mendix storage lists use a constant
+--   storage-list MARKER there. For ChangeActionItem lists this marker is
+--   `2`, independent of how many items the list contains. Writing the
+--   item count produced unstable microflow BSON for create/change object
+--   actions and could lead to invalid action payloads when Studio Pro
+--   re-loaded the model.
+--
+--   Additionally, `ChangeAction` was missing its default
+--   `ErrorHandlingType: Rollback` field, so the written shape did not
+--   match the expected action payload.
+--
+-- After fix:
+--   - Writer emits `bson.A{int32(2)}` for both CreateChangeAction.Items
+--     and ChangeAction.Items.
+--   - Writer emits the default `ErrorHandlingType: Rollback` for
+--     ChangeAction.
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/373-change-action-items-storage-list.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest373.MF_CreateAndChange"
+--   `mx check` against the resulting MPR must report 0 errors.
+-- ============================================================================
+
+create module BugTest373;
+
+create entity BugTest373.Order (
+  Name : string(100),
+  Status : string(50)
+);
+/
+
+-- Create-with-members and change-with-members both exercise the
+-- ChangeActionItem storage list. After exec, the MPR must load cleanly
+-- in Studio Pro (`mx check` reports 0 errors).
+create microflow BugTest373.MF_CreateAndChange (
+  $Name: string
+)
+returns BugTest373.Order as $Order
+begin
+  $Order = create BugTest373.Order (Name = $Name);
+  change $Order (Status = 'Processed');
+end;
+/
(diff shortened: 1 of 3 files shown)


ako (commented):
## PR #374 — fix: write valid change-action item storage lists

---

### No Blockers

The change is correct in principle. Replacing `int32(len(...))` with a constant storage-list marker is strictly better — the count-based sentinel produced different BSON depending on the number of member changes, making the output unstable.

---

### Moderate Issues

#### 1. Marker value `int32(2)` not documented as verified against a real MPR

The PR asserts the correct marker for `ChangeActionItem` lists is `int32(2)`, but the description doesn't state how this was confirmed. Looking across the codebase, storage-list markers are **type-specific**:

- Microflow call parameter mappings (`Microflows$MicroflowCallParameterMapping`): `int32(2)` — fixed by PR #338
- Java action parameter mappings (`Microflows$JavaActionParameterValue`): `int32(3)` — `writer_microflow_actions.go:183`
- Domain model entities, attributes, access rules: `int32(3)` — `writer_domainmodel.go`
- Sort items, REST parameters, business event messages: `int32(2)`

`ChangeActionItem` is a microflow-space type (`Microflows$ChangeActionItem`), so `int32(2)` is plausible and consistent with the microflow-writer pattern. But the Java action case shows that not all microflow-space types use `int32(2)`. If the actual Mendix format requires `int32(3)`, the regression test (which just asserts the marker equals `2`) would pass while the MPR would still be invalid.

A one-line comment citing the verification source — e.g. "verified against a Studio Pro-created MPR" or a reference to issue #373 where an MPR dump was examined — would be enough to close this gap.

---

### Minor Issues

#### 2. `getBSONField` helper comes from a different test file

Both new tests call `getBSONField`, which is defined in `workflow_write_test.go` (line 13). Since both files are in `package mpr`, this works. But the dependency is invisible — if `getBSONField` is ever moved or renamed, the new test breaks without an obvious connection. Acceptable in a test-only file, but worth keeping in mind.

#### 3. `ErrorHandlingType` field position in `ChangeAction`

The new `ErrorHandlingType: "Rollback"` field is inserted after `Commit` and before the `Items` build. In `CreateChangeAction`, `ErrorHandlingType` appears in the same relative position. The ordering is consistent. ✓

---

### Positive Notes

- Bug-test MDL file has correct `create module BugTest373` prerequisite. ✓
- Both regression tests are tightly scoped: one for create-object, one for change-object, each checking the marker and (for ChangeAction) the missing `ErrorHandlingType`.
- The fix correctly addresses both `CreateChangeAction.Items` and `ChangeAction.Items` in one PR — no half-measure.
- Old incorrect comment ("IMPORTANT: Mendix BSON arrays include the count as the first element") removed and replaced with accurate description. ✓

---

**Summary:** No blockers. The PR is well-scoped and the fix is sound. The only gap worth closing before merge is a note on how `int32(2)` was confirmed as the correct marker for `ChangeActionItem` specifically — either in the PR description or as a code comment. If it was verified by examining a Studio Pro-created MPR, a one-liner saying so is enough.

ako:
## Review: fix: write valid change-action item storage lists

Already merged. Retroactive review for tracking.

**Correct fix.** The Mendix storage-list convention uses a fixed `int32(2)` marker as the first BSON array element, not a count. The writer was using `int32(len(items))` for both `CreateChangeAction.Items` and `ChangeAction.Items`, which would produce correct BSON only by accident when the list happened to have exactly 2 members.

The parser (`parseCreateObjectAction`, `parseChangeObjectAction`) was already correct: it uses `extractBsonMap(item) != nil` to skip the leading marker, so it tolerated the old wrong values. No parser fix was needed.

The missing `ErrorHandlingType: Rollback` on `ChangeAction` is a separate but related omission — `CreateChangeAction` already had it in the existing code. Both issues together would have produced action payloads that Studio Pro rejected or loaded incorrectly.

**Test coverage is sufficient:**
- `TestSerializeCreateObjectActionItemsUseStorageListMarker` pins marker = `int32(2)` for CreateObjectAction
- `TestSerializeChangeObjectActionItemsUseStorageListMarkerAndDefaultErrorHandling` pins both the marker and `ErrorHandlingType = "Rollback"` for ChangeObjectAction
- Bug test MDL in `mdl-examples/bug-tests/373-change-action-items-storage-list.mdl` provides a Studio Pro `mx check` validation target

No issues. Scope is correct (three files, one bug).

PR #378, opened 2026-04-27
fix: emit ResultHandlingHttpResponse for REST 'returns response'

### LINKED ISSUE #377: Builder emits ResultHandlingString for REST 'returns response'
## Problem

When MDL authors a REST call with `returns response`, the microflow builder creates a `ResultHandlingString` instead of a `ResultHandlingHttpResponse`. The builder file `mdl/executor/cmd_microflows_builder_calls.go` carries an explicit TODO comment for this case:

```go
case ast.RestResultResponse:
    resultHandling = &microflows.ResultHandlingString{
        BaseElement: model.BaseElement{ID: model.ID(types.GenerateID())},
    }
    // Note: For HttpResponse, we would need a different result type, but using String for now
```

## Impact

PR #376 added explicit writer support for `ResultHandlingHttpResponse` (closing #375), but the writer branch is unreachable from authored MDL because the builder never produces that intermediate. The result is:

- `mxcli exec` of `$Response = rest call get '...' returns response on error continue;` writes `ResultHandlingString` BSON.
- The action-level `ResultHandlingType` becomes `String` (the writer derives it from the resultHandling concrete type).
- Studio Pro then treats `$Response` as a string and downstream `$Response/Content`, `$Response/StatusCode` references fail validation with CE0117.

## Expected behavior

When the AST carries `RestResultResponse`, the builder should construct a `ResultHandlingHttpResponse` so the writer's existing branch (introduced in PR #376) emits a `DataTypes$ObjectType` bound to `System.HttpResponse`.

## Proposed fix

- Replace the `ResultHandlingString` instantiation with `ResultHandlingHttpResponse` in the `case ast.RestResultResponse` arm.
- Reuse the existing `fb.declaredVars[s.OutputVariable] = "System.HttpResponse"` registration in `cmd_microflows_builder_validate.go` so dependent activities resolve.
- Add a builder regression test covering the result-handling concrete type and the action-level `ResultHandlingType`.
- Add an MDL bug-test under `mdl-examples/bug-tests/` that goes through the full pipeline (`mxcli exec` → `mx check`).

## Related

- PR #376 fixes the writer side. Closes #375.
- This issue closes the builder-side counterpart so authored MDL benefits from the writer fix.
- Part of #332 (microflow roundtrip stabilization tracker).

hjotha:
Closes #377.
Stacked on #376.

## Summary

When MDL authors `rest call ... returns response`, the microflow builder now constructs a `ResultHandlingHttpResponse` instead of a placeholder `ResultHandlingString`. The writer's matching branch (added in #376) then emits a bound `Microflows$ResultHandling` whose `VariableType` is `DataTypes$ObjectType` with entity `System.HttpResponse`, and the action-level `ResultHandlingType` correctly derives as `HttpResponse`.

## Why

The builder file `mdl/executor/cmd_microflows_builder_calls.go` had an explicit TODO:

```go
case ast.RestResultResponse:
    resultHandling = &microflows.ResultHandlingString{...}
    // Note: For HttpResponse, we would need a different result type, but using String for now
```

That made #376's writer branch unreachable from authored MDL. After `mxcli exec`, Studio Pro saw the bound variable as a string, so `$Response/Content` and `$Response/StatusCode` references failed validation with CE0117.

## Stacking note

This branch is stacked on #376 because the writer needs the explicit `ResultHandlingHttpResponse` branch for the BSON shape to be correct. Without #376, the action-level type would be `HttpResponse` but the nested `ResultHandling` would still fall through to the string default. Merging #376 first lets this PR rebase cleanly onto `main` with no conflicts.

## Validation

- `make build`
- `make lint-go`
- `make test`
- `mx check` end-to-end on `mdl-examples/bug-tests/377-rest-returns-response-builder.mdl` reports 0 errors

## Tests

- Builder regression `TestAddRestCallAction_ReturnsResponseUsesHttpResponseHandling` pins the result-handling concrete type and bound variable name.
- MDL bug-test `mdl-examples/bug-tests/377-rest-returns-response-builder.mdl` exercises the full pipeline through `mxcli exec` + `mx check`.

## Test plan

- [x] Builder produces `ResultHandlingHttpResponse` for `returns response`.
- [x] `OutputVariable` flows into `VariableName` so the writer binds correctly.
- [x] `mxcli exec` writes a Studio Pro–valid MPR that passes `mx check`.
- [x] Existing string/mapping/none paths remain unchanged.

### DIFF
--- a/mdl-examples/bug-tests/375-rest-httpresponse-result-type.mdl
+++ b/mdl-examples/bug-tests/375-rest-httpresponse-result-type.mdl
@@ -1,0 +1,69 @@
+-- ============================================================================
+-- Bug #375: REST HttpResponse result handling serialized as string
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   When a Studio Pro REST call action bound the complete HTTP response
+--   to an output variable, the parser/AST captured this as
+--   `ResultHandlingHttpResponse` and the action-level
+--   `ResultHandlingType` was `HttpResponse`. The writer, however, did
+--   NOT have an explicit branch for `ResultHandlingHttpResponse` inside
+--   `serializeRestResultHandling`, so it fell through to the string
+--   default. The serialized result variable was emitted as
+--   `DataTypes$StringType`, not the expected
+--   `DataTypes$ObjectType` bound to `System.HttpResponse`.
+--   Studio Pro then surfaced a type mismatch on the bound output
+--   variable, and dependent activities that reached into the response
+--   (`$Response/Content`, `$Response/StatusCode`) failed validation
+--   with CE0117.
+--
+-- After fix (writer only):
+--   `serializeRestResultHandling` now has an explicit
+--   `ResultHandlingHttpResponse` branch that emits a `DataTypes$ObjectType`
+--   variable type bound to `System.HttpResponse`.
+--
+-- Scope note — known related gap:
+--   The MDL builder for `rest call ... returns response` still creates
+--   a `ResultHandlingString` (not `ResultHandlingHttpResponse`). See
+--   `mdl/executor/cmd_microflows_builder_calls.go` — the
+--   `case ast.RestResultResponse` arm carries the comment
+--   "we would need a different result type, but using String for now".
+--   That means newly authored MDL with `returns response` does NOT yet
+--   trigger the new writer branch. PR #376 fixes the symptom for
+--   existing Studio Pro models that already contain
+--   `ResultHandlingHttpResponse` in their BSON; the BUILDER side is
+--   tracked separately and verified by the SDK-level Go test
+--   `TestSerializeRestResultHandlingHttpResponseUsesObjectType`, which
+--   constructs the AST directly.
+--
+-- Usage (positive control):
+--   mxcli exec mdl-examples/bug-tests/375-rest-httpresponse-result-type.mdl -p app.mpr
+--   `mx check` against the resulting MPR must report 0 errors. The
+--   negative case (a Studio Pro REST call with HttpResponse binding)
+--   needs the existing BSON shape and is covered by the Go test above.
+-- ============================================================================
+
+create module BugTest375;
+
+-- Inline REST call returning a string. Once the builder learns to emit
+-- ResultHandlingHttpResponse for `returns response`, this script can be
+-- updated to use `returns response` and exercise the writer branch end
+-- to end.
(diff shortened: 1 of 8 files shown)


ako (commented):
## PR #378 — fix: emit ResultHandlingHttpResponse for REST 'returns response'

Clean, complete fix. The three layers (builder, formatter, writer) are all addressed, the test coverage directly pins the concrete type and field values at each layer, and the end-to-end `mx check` validation was run. The stacking on #376 is clearly described.

---

### Minor — `375-rest-httpresponse-result-type.mdl` has a stale "known related gap" note

`mdl-examples/bug-tests/375-rest-httpresponse-result-type.mdl` (from PR #376, included here because the branch is stacked) carries two items that this PR closes:

1. The "scope note — known related gap" section explicitly says `case ast.RestResultResponse` is not yet fixed and tracked separately. This PR IS that fix.
2. The microflow uses `returns string` with a comment "Once the builder learns to emit `ResultHandlingHttpResponse` for `returns response`, this script can be updated". The builder now does exactly that.

Before this lands on `main`, the bug-test should be updated:
- Remove the "scope note" and replace with "Fixed by PR #378".
- Update the microflow to use `returns response` end-to-end to confirm the writer branch is now reachable from authored MDL (matching what `377-rest-returns-response-builder.mdl` already does).

---

### Minor — `ResultHandlingHttpResponse.VariableName` is implicitly dual-populated

The builder sets both `ResultHandlingHttpResponse{VariableName: s.OutputVariable}` and `RestCallAction{OutputVariable: s.OutputVariable}`. The writer uses `a.OutputVariable` (the RestCallAction field); the formatter uses `a.OutputVariable` first and falls back to `rh.VariableName` when empty (for describe from BSON where the RestCallAction field may not be populated). The builder comment explains why this concrete type must be used, but not why `VariableName` is set — a one-liner would help: "VariableName is also read by the formatter as a fallback on the describe path when RestCallAction.OutputVariable is empty".

---

### Checklist

- Grammar: `RETURNS RESPONSE` already exists in the parser (line 1663), `RESPONSE` token is case-insensitive. ✓
- Visitor: `ast.RestResultResponse` already wired at visitor line 1025. ✓
- Formatter now emits lowercase `"response"` which round-trips correctly through the case-insensitive parser. ✓
- `getBSONField` test helper already exists in `workflow_write_test.go` (same `package mpr`). ✓
- No grammar/AST/visitor changes needed. ✓

PR #381, opened 2026-04-29
fix: emit list retrieve for owner-both reverse references

### LINKED ISSUE #380: Reverse Reference retrieve with owner-both association is typed as single object
## Problem

A microflow retrieve that navigates a `Reference` association from the child side can be written back as an association retrieve even when the association has explicit ownership metadata and the parent entity is persistable.

For a one-to-many reverse traversal, an association retrieve source represents a single object. The builder should instead emit a database retrieve source with an XPath constraint so the result remains a list of parent objects.

## Expected behavior

For an explicit-owner `Reference` association, including `Owner=Both`, child-to-parent traversal over a persistable parent should be written as a database retrieve with an XPath constraint such as:

```mdl
retrieve $Parents from database Sample.Parent where [Sample.Parent_Child = $Child];
```

Fallback association-source syntax is still acceptable when ownership metadata is missing or the parent entity is not persistable, but the inferred result type should remain `List of Sample.Parent` for child-side traversal.

## Impact

Downstream actions that rely on the retrieved value being a list can become invalid after round-trip when the reverse traversal is typed as a single object.

hjotha:
Closes #380.

## Summary

- Treat explicit-owner reverse `Reference` traversal, including `Owner=Both`, as a database retrieve with XPath when the parent entity is persistable.
- Keep association-source fallback for missing ownership metadata or non-persistable parent entities.
- Preserve list typing for child-side `Reference` traversal even when fallback source syntax is used.

## Validation

- `go test ./mdl/executor -run 'TestAddRetrieveAction_ReverseReference'`
- `make build`
- `make test`

### DIFF
--- a/mdl/executor/cmd_microflows_builder_actions.go
+++ b/mdl/executor/cmd_microflows_builder_actions.go
@@ -303,7 +303,7 @@ func (fb *flowBuilder) addRetrieveAction(s *ast.RetrieveStmt) model.ID {
 		}
 
 		if assocInfo != nil && assocInfo.Type == domainmodel.AssociationTypeReference &&
-			assocInfo.Owner != domainmodel.AssociationOwnerBoth &&
+			assocInfo.Owner != "" &&
 			assocInfo.parentPersistable &&
 			assocInfo.childEntityQN != "" && startVarType == assocInfo.childEntityQN {
 			// Reverse traversal on Reference: child → parent (one-to-many)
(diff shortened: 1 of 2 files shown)


ako (commented):
## PR #381 — fix: emit list retrieve for owner-both reverse references

---

### No Blockers

The core fix is correct. `Owner=Both` (symmetric association) is still a one-to-many relationship when traversed in reverse (child → parent), so using `DatabaseRetrieveSource` with an XPath constraint is the right output. The old `!= AssociationOwnerBoth` condition incorrectly routed it to `AssociationRetrieveSource`.

---

### Moderate Issues

#### 1. Missing `mdl-examples/bug-tests/` file

The checklist requires a bug-test MDL script for every bug fix that is verifiable in Studio Pro:

> "every bug fix should include an MDL test script in `mdl-examples/bug-tests/` that reproduces the issue, so the fix can be verified in Studio Pro if applicable"

The fix changes how `ChangeAction` BSON retrieve actions are built — directly affecting what ends up in the MPR. A short script that creates an entity with a `Both`-owned association and retrieves via reverse traversal, with a note to run `mx check`, would be the natural companion.

---

### Minor Issues

#### 2. `Owner=""` behavioural change is implicit in the condition swap

Changing `!= AssociationOwnerBoth` to `!= ""` has a second effect beyond the stated fix: associations where `Owner` is empty (missing metadata) now fall through to `AssociationRetrieveSource` instead of `DatabaseRetrieveSource`. Previously `"" != "Both"` was `true`, so empty-owner associations would have entered the database-retrieve path.

The new test `TestAddRetrieveAction_ReverseReferenceUnknownOwnerPreservesAssociationSource` correctly documents this new behaviour, and the PR summary mentions "missing ownership metadata" as an intended fallback case. The only gap is that the summary doesn't explicitly note this is also a behaviour change from the previous state (not just a preservation of pre-PR-356 behaviour). A one-line note in the PR description or a code comment at the `!= ""` guard (`// Owner="" means metadata was unavailable; fall back to association source`) would make the intent self-evident for future readers.

---

### Positive Notes

- List typing fix in the `else` branch (`"List of " + otherEntity` for reverse traversal) is correct and appropriately scoped — it applies only when `startVarType == assocInfo.childEntityQN`.
- Three test cases cover the three distinct owner states (`Both` → database source, `""` → association source, non-persistable → association source), and the `NonPersistableParent` test correctly updated its varType assertion to `"List of Sample.Parent"`. ✓
- No grammar, AST, or BSON writer changes needed — the fix is entirely in the routing logic. ✓

---

**Summary:** No blockers. The fix is correct and test coverage is solid. Two gaps: missing `mdl-examples/bug-tests/` file (checklist item), and the implicit `Owner=""` behaviour change warrants a one-line comment at the guard.

PR #384, opened 2026-04-29
fix: preserve object-valued owner-both reverse retrieves

### LINKED ISSUE #383: Object-valued owner-both reverse retrieves can rebuild as lists
## Summary

A compact reverse `Reference` retrieve over an owner-both association can be rebuilt as a database retrieve with list cardinality even when later expressions consume the result as a single object.

## Symptoms

A microflow can contain a compact retrieve like:

```mdl
retrieve $Parent from $Child/Sample.Parent_Child;
call microflow Sample.UseParent(parentName = $Parent/Name);
```

After `describe` + `exec`, the retrieve may be stored as a database retrieve returning a list. The following attribute path then becomes invalid because `$Parent` is no longer object-valued.

## Expected behavior

When the retrieved variable is later used through an object attribute path, the builder should preserve the compact association retrieve and object-valued typing.

When the same owner-both reverse retrieve is later used as a list input, such as in a loop, the builder should still emit the database/list shape needed by list consumers.

## Notes

This is a follow-up to the reverse owner-both list retrieve fix. The missing distinction is not association metadata alone; the builder also needs downstream usage context.

hjotha:
Closes #383.

## Summary

- pre-scans microflow statements for variables later consumed as list inputs or object/attribute inputs
- keeps owner-both reverse `Reference` retrieves as database/list retrieves only for list consumers
- preserves compact association retrieves and object-valued builder typing when the result is dereferenced as an object

## Validation

- `make build`
- `make test`
- `make lint-go`
- targeted Mendix `mx check` repro on a copied MPR returned 0 errors

### DIFF
--- a/mdl/executor/cmd_microflows_builder.go
+++ b/mdl/executor/cmd_microflows_builder.go
@@ -37,11 +37,13 @@ type flowBuilder struct {
 	// continues) so the continuing branch's @anchor survives to the actual
 	// splitID→nextActivity flow — which is emitted one iteration later by the
 	// outer loop, not by addIfStatement.
-	nextFlowAnchor     *ast.FlowAnchors
-	backend            backend.FullBackend          // For looking up page/microflow references
-	hierarchy          *ContainerHierarchy          // For resolving container IDs to module names
-	pendingAnnotations *ast.ActivityAnnotations     // Pending annotations to attach to next activity
-	restServices       []*model.ConsumedRestService // Cached REST services for parameter classification
+	nextFlowAnchor       *ast.FlowAnchors
+	backend              backend.FullBackend          // For looking up page/microflow references
+	hierarchy            *ContainerHierarchy          // For resolving container IDs to module names
+	pendingAnnotations   *ast.ActivityAnnotations     // Pending annotations to attach to next activity
+	restServices         []*model.ConsumedRestService // Cached REST services for parameter classification
+	listInputVariables   map[string]bool              // Variables later consumed by list-only actions
+	objectInputVariables map[string]bool              // Variables later consumed through object/attribute access
 	// previousStmtAnchor holds the Anchor annotation of the statement that
 	// just emitted an activity, so the next flow's OriginConnectionIndex can
 	// be overridden by the user. Cleared after each flow is created.
--- a/mdl/executor/cmd_microflows_builder_actions.go
+++ b/mdl/executor/cmd_microflows_builder_actions.go
@@ -302,10 +302,19 @@ func (fb *flowBuilder) addRetrieveAction(s *ast.RetrieveStmt) model.ID {
 			startVarType = fb.varTypes[s.StartVariable]
 		}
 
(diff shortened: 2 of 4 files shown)


ako (commented):
## PR Review — fix: preserve object-valued owner-both reverse retrieves (#384)

**No blockers.** Two moderate issues, two minor notes.

---

### Moderate

**1. Missing `mdl-examples/bug-tests/` file for issue #383**

The PR checklist requires a bug-test MDL script in `mdl-examples/bug-tests/` that reproduces the issue so the fix can be verified in Studio Pro. None was added. Please add a minimal script that:
- Creates an `Owner=Both` association (e.g., `Customer_Order`)
- Retrieves via reverse reference (`$customer` → `$orders` list, then `$customer` back as object)

**2. `collectListInputVariables` misses `AddToListStmt.List` and `RemoveFromListStmt.List`**

If the output variable of an `Owner=Both` reverse retrieve is immediately used as the *target list* in an `add`/`remove` statement:

```mdl
retrieve $items from $parent via Parent.Child_Parent
add $newItem to $items
```

`$items` won't be marked as a list consumer because `AddToListStmt.List` (and `RemoveFromListStmt.List`) are not visited in `collectListInputVariables`. The variable will be seen as an object consumer instead, and the `!outputUsedAsObject` guard will suppress `AssociationRetrieveSource` — triggering the original bug again for this usage pattern.

Fix: add these two cases to `collectListInputVariables`:

```go
case *ast.AddToListStmt:
    if s.List != "" {
        vars[stripDollar(s.List)] = true
    }
case *ast.RemoveFromListStmt:
    if s.Variable != "" {
        vars[stripDollar(s.Variable)] = true
    }
```

(Check the field names against the AST struct definitions.)

---

### Minor

**3. Implicit operator precedence in `expandReverseReference`**

The final condition:

```go
(assocInfo.Owner != domainmodel.AssociationOwnerBoth || outputUsedAsList && !outputUsedAsObject)
```

`&&` binds tighter than `||` so this is correct, but a reader has to stop and think. Explicit parentheses make the intent obvious:

```go
(assocInfo.Owner != domainmodel.AssociationOwnerBoth || (outputUsedAsList && !outputUsedAsObject))
```

**4. Error-handler sub-builders don't inherit the pre-scan maps**

The `listInputVariables`/`objectInputVariables` maps are populated by a pre-scan of the top-level statement slice. If a retrieval is inside an error-handler body it's built by a child `flowBuilder` that starts fresh — it won't inherit these maps. Retrievals inside error handlers that use `Owner=Both` associations will always take the conservative AssociationRetrieveSource fallback. This is safe (no regression), but worth a comment or a follow-up issue.

---

### Positive

- The 4 new tests are well-structured and cover the key cases (Both+list, Both+object, Default owner, Both+ambiguous fallback).
- `firstRetrieveAction` helper keeps the test assertions readable.
- `sourceAttributeVarRefs` correctly uses the `$Var/` prefix pattern — it won't misfire on bare `$Var` references.
- The conservative fallback (AssociationRetrieveSource when context is unknown) is the right default.

PR #408, opened 2026-04-30
fix: address merged microflow review followups

### LINKED ISSUE #404: Manual while-true detection over-triggers when continue lives only inside nested loop body
## Source

Follow-up from review of merged PR #354 (fix: preserve manual while-true loop roundtrips).

## Problem

`containsContinueStmt` recurses into nested `LoopStmt` / `WhileStmt` bodies, which makes `isManualWhileTrueCandidate` mis-classify an outer `while true` as a manual-back-edge candidate when the only \`continue\` statement targets an inner collection loop:

\`\`\`mdl
while true
begin
  loop \$item in \$items
  begin
    continue;   -- targets the inner loop, NOT the outer while
  end loop;
  -- outer body has no continue / return / raise
end while;
\`\`\`

The outer \`while true\` should be built as a regular \`LoopedActivity\` with \`WhileLoopCondition\`, but is instead built as an \`ExclusiveMerge\` back-edge structure. That creates an unconditional infinite loop in the BSON graph.

This is asymmetric with \`containsBreakForCurrentLoop\`, which correctly stops at nested loop boundaries (\`break\` inside a nested loop exits that nested loop).

## Expected behavior

A \`continue\` that lives only inside a nested loop body should NOT contribute to the outer \`while true\`'s manual-candidate classification. \`isManualWhileTrueCandidate\` should treat outer-scope \`continue\` only.

## Proposed fix

Add a sibling \`containsContinueForCurrentLoop\` that mirrors \`containsBreakForCurrentLoop\` (stops at \`*ast.LoopStmt\` / \`*ast.WhileStmt\` boundaries) and use it in \`isManualWhileTrueCandidate\` instead of the broader \`containsContinueStmt\`. Keep the broader helper if its other call sites still need it; otherwise drop it.

\`\`\`go
func containsContinueForCurrentLoop(stmts []ast.MicroflowStatement) bool {
    for _, stmt := range stmts {
        switch s := stmt.(type) {
        case *ast.ContinueStmt:
            return true
        case *ast.IfStmt:
            if containsContinueForCurrentLoop(s.ThenBody) || containsContinueForCurrentLoop(s.ElseBody) {
                return true
            }
        case *ast.LoopStmt, *ast.WhileStmt:
            // continue inside a nested loop targets that loop, not this one.
        }
    }
    return false
}
\`\`\`

## Tests

- Add a regression test for the false-positive scenario above (outer \`while true\` with continue only inside nested loop) — assert the outer flow is built as \`LoopedActivity\`, not \`ExclusiveMerge\`.
- Strengthen \`TestBuildFlowGraph_ManualWhileTrueTerminalDoesNotAddFallthroughEnd\`: count EndEvent nodes explicitly instead of relying on a guard that always passes when no EndEvent is added.

## Related

- Reviewed in #354 by ako.
- Part of #332 (microflow roundtrip stabilization tracker).

### LINKED ISSUE #405: Reverse retrieve incorrectly classified when output feeds add/remove-from-list statements
## Source

Follow-up from review of merged PR #384 (fix: preserve object-valued owner-both reverse retrieves).

## Problem

\`collectListInputVariables\` in \`mdl/executor/cmd_microflows_builder_actions.go\` decides whether the output of an \`Owner=Both\` reverse retrieve is consumed as a list. It walks all statements that take a list input — \`AggregateListStmt\`, \`ListOperationStmt\`, \`LoopStmt\` — but misses the two statements that read a list as a target:

- \`AddToListStmt.List\` — target of \`add \$item to \$list\`
- \`RemoveFromListStmt.Variable\` (or whichever field holds the list reference) — target of \`remove \$item from \$list\`

If the output of a reverse retrieve is fed straight into \`add\`/\`remove\`, it isn't marked as a list consumer, so the \`!outputUsedAsObject\` guard in \`expandReverseReference\` suppresses \`AssociationRetrieveSource\` and the original bug from #383 reappears for this usage shape.

## Expected behavior

\`add \$X to \$Outputs\` and \`remove \$Y from \$Outputs\` should mark \`\$Outputs\` as a list consumer for the \`Owner=Both\` reverse retrieve classification.

## Proposed fix

Extend \`collectListInputVariables\` with the missing cases:

\`\`\`go
case *ast.AddToListStmt:
    if s.List != \"\" {
        vars[stripDollar(s.List)] = true
    }
case *ast.RemoveFromListStmt:
    // Field name as defined in mdl/ast — adjust if different.
    if s.List != \"\" {
        vars[stripDollar(s.List)] = true
    }
\`\`\`

## Tests

- Add a regression test where a reverse retrieve's output is consumed only by \`add to list\` and assert \`AssociationRetrieveSource\` is suppressed.
- Add a regression test for \`remove from list\`.

## Related

- Reviewed in #384 by ako.
- Part of #332 (microflow roundtrip stabilization tracker).

### LINKED ISSUE #406: describeNanoflow loses void-microflow EndEvent fix
## Source

Follow-up from review of merged PR #371 (fix: preserve void EndEvent returns in describe).

## Problem

PR #371 made \`describeMicroflow\` and \`renderMicroflowMDL\` (the LSP rendering path) emit \`return;\` for empty \`EndEvent\` nodes in void microflows. \`describeNanoflow\` was not updated.

\`describeNanoflow\` wraps the nanoflow's \`ObjectCollection\` in a bare \`&microflows.Microflow{}\` (no \`ReturnType\` set) so it can reuse \`formatMicroflowActivities\`. It does not set \`DescribingMicroflowHasReturnValue\` on the \`ExecContext\`. As a consequence, value-returning nanoflows with early-exit terminal paths still drop the explicit \`EndEvent\` on describe — exactly the symptom #371 fixed for microflows.

## Expected behavior

Value-returning nanoflows preserve their explicit \`return;\` (void) or value-bearing return on describe, and bare \`return;\` is suppressed for value-returning nanoflows that don't have an explicit early-exit.

## Proposed fix

Set \`DescribingMicroflowHasReturnValue\` (and any companion ExecContext fields) inside \`describeNanoflow\` from the nanoflow's own \`ReturnType\`, mirroring \`describeMicroflow\`. Add a \`describeNanoflow\` regression test for both void and value-returning nanoflow shapes.

## Related

- Reviewed in #371 by ako.
- Part of #332.

hjotha:
## Summary

This PR collects follow-ups from reviews on already-merged microflow roundtrip PRs so the actionable feedback does not get lost.

Fixes:
- Stop manual `while true` detection from treating `continue` inside a nested loop as a continue for the outer loop.
- Mark `add $Item to $List` and `remove $Item from $List` target lists as list consumers so owner-both reverse retrieves keep the correct source form.
- Set nanoflow return-value render context on the direct `describe nanoflow` path so value-returning nanoflows do not emit bare `return;` for empty EndEvents.

Review follow-up coverage:
- Add writer coverage that commit actions always serialize default `ErrorHandlingType`.
- Fix the EndEvent empty-return writer test to exercise the actual empty string case.
- Add a negative regression assertion for non-empty change-object refresh inference.
- Add download-file formatter coverage for the no-browser flag case.
- Tighten reverse-association compact XPath name validation and make the owner-both condition explicit.
- Document `MXCLI_EXEC_TIMEOUT` in the quick reference.

Closes #404.
Closes #405.
Closes #406.

## Tests

- `make build`
- `make test`
- `make lint-go`

### DIFF
--- a/docs/01-project/MDL_QUICK_REFERENCE.md
+++ b/docs/01-project/MDL_QUICK_REFERENCE.md
@@ -1054,6 +1054,8 @@ Cross-reference commands require `refresh catalog full` to populate reference da
 | Setup mxcli | `mxcli setup mxcli [--os linux]` | Download platform-specific mxcli binary |
 | LSP server | `mxcli lsp --stdio` | Language server for VS Code |
 
+Set `MXCLI_EXEC_TIMEOUT` to override the per-statement execution timeout used by `mxcli exec` (for example `MXCLI_EXEC_TIMEOUT=12m` or `MXCLI_EXEC_TIMEOUT=900`).
+
 ## ANTLR4 Parser Architecture
 
 The MDL parser uses ANTLR4 for grammar definition, enabling cross-language grammar sharing (Go, TypeScript, Java, Python).
--- a/mdl/executor/bugfix_regression_test.go
+++ b/mdl/executor/bugfix_regression_test.go
@@ -694,6 +694,28 @@ func TestEmptyChangeObjectRefreshesInClient(t *testing.T) {
 	if !action.RefreshInClient {
 		t.Fatal("empty change object must refresh in client to remain valid without member changes or commit")
 	}
+
+	id = fb.addChangeObjectAction(&ast.ChangeObjectStmt{
+		Variable: "Object",
+		Changes: []ast.ChangeItem{{
+			Attribute: "Name",
+			Value:     &ast.LiteralExpr{Kind: ast.LiteralString, Value: "changed"},
+		}},
+	})
+	if id == "" || len(fb.objects) != 2 {
+		t.Fatalf("expected second change object activity, got id=%q objects=%d", id, len(fb.objects))
+	}
+	activity, ok = fb.objects[1].(*microflows.ActionActivity)
+	if !ok {
+		t.Fatalf("object type = %T, want *microflows.ActionActivity", fb.objects[1])
+	}
+	action, ok = activity.Action.(*microflows.ChangeObjectAction)
+	if !ok {
+		t.Fatalf("action type = %T, want *microflows.ChangeObjectAction", activity.Action)
+	}
+	if action.RefreshInClient {
+		t.Fatal("non-empty change object must not infer refresh in client")
+	}
 }
 
 func TestListFindAttributeEqualsExpressionUsesAttributeOperation(t *testing.T) {
--- a/mdl/executor/cmd_javaactions_test.go
+++ b/mdl/executor/cmd_javaactions_test.go
@@ -19,6 +19,7 @@ func TestFormatAction_JavaActionCall_EntityTypeParam(t *testing.T) {
 	action := &microflows.JavaActionCallAction{
 		JavaAction:         "MyModule.Validate",
 		ResultVariableName: "IsValid",
+		UseReturnVariable:  true,
 		ParameterMappings: []*microflows.JavaActionParameterMapping{
 			{
 				Parameter: "MyModule.Validate.InputObject",
@@ -40,6 +41,7 @@ func TestFormatAction_JavaActionCall_MixedParamTypes(t *testing.T) {
 	action := &microflows.JavaActionCallAction{
 		JavaAction:         "MyModule.ProcessEntity",
 		ResultVariableName: "Result",
+		UseReturnVariable:  true,
 		ParameterMappings: []*microflows.JavaActionParameterMapping{
 			{
 				Parameter: "MyModule.ProcessEntity.InputObject",
@@ -67,6 +69,7 @@ func TestFormatAction_JavaActionCall_EntityTypeParam_EmptyEntity(t *testing.T) {
 	action := &microflows.JavaActionCallAction{
 		JavaAction:         "MyModule.Validate",
 		ResultVariableName: "IsValid",
+		UseReturnVariable:  true,
 		ParameterMappings: []*microflows.JavaActionParameterMapping{
 			{
 				Parameter: "MyModule.Validate.InputObject",
@@ -280,6 +283,7 @@ func TestFormatAction_JavaActionCall_EntityTypeAndParameterizedParams(t *testing
 	action := &microflows.JavaActionCallAction{
 		JavaAction:         "MyModule.CopyAttributes",
 		ResultVariableName: "Result",
+		UseReturnVariable:  true,
 		ParameterMappings: []*microflows.JavaActionParameterMapping{
 			{
 				Parameter: "MyModule.CopyAttributes.EntityType",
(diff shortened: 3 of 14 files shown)


hjotha:

ako:
## Review: fix/address-merged-microflow-review-followups

**Result: Approved with one minor nit.**

### What the PR does

Collects five bug fixes and several test improvements from reviews on already-merged microflow PRs.

### Fix-by-fix analysis

**1.  (issue #404)**
Old  recursed into nested / bodies, making it asymmetric with . This caused  to fire on outer whiles whose only `continue` lived inside a nested loop, producing an unconditional back-edge instead of a `LoopedActivity`. Fix correctly stops at nested loop boundaries. New test verifies the outer while is emitted as `LoopedActivity` with a `LoopSource` and zero `ExclusiveMerge` nodes.

**2.  for add/remove targets (issue #405)**
`AddToListStmt` and `RemoveFromListStmt` target lists were not tracked as list consumers, so an owner-both reverse retrieve fed into add/remove was misclassified as object-only and `AssociationRetrieveSource` was suppressed. Fix adds both cases; regression test pins the expected behavior.

**3.  guard in  (issue #406)**
Studio Pro stores a `ResultVariableName`/`OutputVariableName` regardless of whether `UseReturnVariable` is set, so checking only the name was producing spurious `$Result = ...` assignment syntax in DESCRIBE output. All call-action formatters (microflow, nanoflow, java action, external action, javascript action, workflow) now gate on `UseReturnVariable && name != ""`. Existing tests correctly updated to set `UseReturnVariable: true` where assignment output is expected.

**4.  return-value render context**
The wrapper microflow was only populated inside the `ObjectCollection != nil` branch, so `DescribingMicroflowHasReturnValue` was never set for nanoflows. Empty `EndEvent` nodes therefore always emitted `return;` even in void nanoflows that should suppress it. Fix unconditionally creates the wrapper before formatting and restores the flag via defer. New mock test verifies a returning nanoflow no longer emits `return;` for an empty `EndEvent`.

**5.  underscore-at-start fix**
The old loop structure allowed `_` at index 0 (it was ORed with letter/digit checks before the `i > 0` guard). New code explicitly returns false at index 0 if the character is not a letter. Three new rejection cases added to `TestParseReverseAssociationXPathRejectsComplexPredicates`.

**6. Owner-both parentheses clarification**
The added parens around `(outputUsedAsList && !outputUsedAsObject)` are cosmetic only — Go's operator precedence already makes the old expression equivalent. The comment makes the intent unambiguous.

### Minor nit

The `MXCLI_EXEC_TIMEOUT` addition in the quick reference is a free-text paragraph, while the surrounding CLI reference uses table rows. Consider making it a table row under the CLI commands section for visual consistency — but not a blocker.

### Other items
- Commit action ErrorHandlingType writer test: clean, fills genuine coverage gap.
- EndEvent rename to `VoidOrUnknownContext`: accurate, good signal.
- Non-empty change-object negative assertion: closes the regression guard correctly.

PR #444, opened 2026-04-30
fix: emit loop captions during microflow describe

### LINKED ISSUE #443: describe drops captions on loop and while activities
## Symptom

`describe microflow --format mdl` emits `@caption` metadata for action activities and split activities, but captions stored on loop/while activities are omitted from the generated MDL.

That means a microflow containing a captioned loop can lose visible caption metadata after a describe/exec roundtrip.

## Root cause

Loop and while statements are represented by `microflows.LoopedActivity`. The builder already preserves `LoopedActivity.Caption`, but the describer helper only emitted captions for `ActionActivity`, `ExclusiveSplit`, and `InheritanceSplit`.

## Expected behavior

When a `LoopedActivity` has a non-empty caption, the describer should emit it as an `@caption` annotation before the loop/while statement, using the same MDL quoting rules as other activity captions.

## Validation

A focused regression test should cover a loop caption with MDL-sensitive characters so the quoting path stays aligned with existing caption emission.

hjotha:
Closes #443.

## Summary

- Emit `@caption` for `microflows.LoopedActivity` during microflow describe.
- Use the same `mdlQuote` path as other activity captions.
- Add a focused regression test covering apostrophe and newline escaping in loop captions.

## Validation

- `make build`
- `go test ./mdl/executor -run 'TestEmitObjectAnnotations_LoopCaption|TestLoopCaptionPreserved|TestWhileLoopCaptionPreserved' -v`
- `make test`
- `make lint-go`

### DIFF
--- a/mdl/executor/cmd_microflows_show_helpers.go
+++ b/mdl/executor/cmd_microflows_show_helpers.go
@@ -462,6 +462,9 @@ func emitObjectAnnotations(
 	if split, ok := obj.(*microflows.InheritanceSplit); ok && split.Caption != "" {
 		*lines = append(*lines, indentStr+fmt.Sprintf("@caption %s", mdlQuote(split.Caption)))
 	}
+	if loop, ok := obj.(*microflows.LoopedActivity); ok && loop.Caption != "" {
+		*lines = append(*lines, indentStr+fmt.Sprintf("@caption %s", mdlQuote(loop.Caption)))
+	}
 
 	// @annotation (attached Annotation objects)
 	if annotationsByTarget != nil {
--- a/mdl/executor/cmd_microflows_show_helpers_test.go
+++ b/mdl/executor/cmd_microflows_show_helpers_test.go
@@ -160,6 +160,25 @@ func TestEmitObjectAnnotations_EscapesMultilineText(t *testing.T) {
 	}
 }
 
+func TestEmitObjectAnnotations_LoopCaption(t *testing.T) {
+	obj := &microflows.LoopedActivity{
+		BaseMicroflowObject: microflows.BaseMicroflowObject{
+			BaseElement: model.BaseElement{ID: mkID("loop")},
+			Position:    model.Point{X: 100, Y: 200},
+		},
+		Caption: "Loop owner's\ncaption",
+	}
+
+	var lines []string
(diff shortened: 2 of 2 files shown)


ako (commented):
## PR #444 Review — fix: emit loop captions during microflow describe

Clean, focused fix. Three lines of production code that mirror the adjacent `ExclusiveSplit` and `InheritanceSplit` handlers exactly. The write side (`applyAnnotations` in `cmd_microflows_builder_annotations.go`) already set `Caption` on `LoopedActivity` — this completes the roundtrip by emitting it on the describe side.

The unit test is good: covers apostrophe doubling and newline escaping through `mdlQuote`, which are the two failure modes that matter.

### Minor

**No doctype-level example for `@caption` on a loop**

`02-microflow-examples.mdl` has `@caption` examples on activities and splits but nothing on a loop or while-loop statement. `TestMxCheck_DoctypeScripts` would never have caught this regression. Worth adding one microflow in that file with a loop that carries `@caption`, e.g.:

```
@caption 'Process each item'
loop $Item in $Items
  log info $Item/Name;
end loop;
```

This doubles as living documentation that `@caption` is valid on loop statements.

---

Otherwise looks good to merge.

ako:
## PR #444 Review — fix: emit loop captions during microflow describe

Clean, focused fix. Three lines of production code that mirror the adjacent `ExclusiveSplit` and `InheritanceSplit` handlers exactly. The write side (`applyAnnotations` in `cmd_microflows_builder_annotations.go`) already set `Caption` on `LoopedActivity` — this completes the roundtrip by emitting it on the describe side.

The unit test is good: covers apostrophe doubling and newline escaping through `mdlQuote`, which are the two failure modes that matter.

### Minor

**No doctype-level example for `@caption` on a loop**

`02-microflow-examples.mdl` has `@caption` examples on activities and splits but nothing on a loop or while-loop statement. `TestMxCheck_DoctypeScripts` would never have caught this regression. Worth adding one microflow in that file with a loop that carries `@caption`, e.g.:

```
@caption 'Process each item'
loop $Item in $Items
  log info $Item/Name;
end loop;
```

This doubles as living documentation that `@caption` is valid on loop statements.

---

Otherwise looks good to merge.

PR #412, opened 2026-04-30
fix: stabilize integration CI baseline

hjotha:
## Summary

- guard watcher debounce callbacks with a generation counter so stale timers cannot send extra change notifications
- update nanoflow integration fixtures to current MDL syntax
- classify known doctype showcase consistency errors so `make test-integration` fails only on unexpected validation errors

## Validation

```bash
make build
go test ./cmd/mxcli/tui -run TestWatcherDebounce -count=20 -v
./bin/mxcli check mdl-examples/doctype-tests/02b-nanoflow-examples.mdl
CGO_ENABLED=0 go test -tags integration -count=1 -timeout 30m ./mdl/executor -run 'TestRoundtripNanoflow_(Loop|EnumParameter|Annotations)|TestMxCheck_DoctypeScripts/02b-nanoflow-examples.mdl|TestMxCheck_DoctypeScripts/03-page-examples.mdl' -v
make test
make lint-go
make test-integration
```

Closes #411.

PR #448, opened 2026-04-30
test: cover qualified function calls with three segments

### LINKED ISSUE #447: Qualified function call tests should cover three-segment callees
## Symptom

The parser accepts qualified names as function-call targets in microflow expressions, but the regression test only pinned a two-segment callee.

That leaves the three-segment shape emitted by describe for rule-like split conditions less directly covered.

## Expected behavior

The visitor test should assert a synthetic three-segment callee, and a bug-test MDL file should exercise the same IF-condition syntax through `mxcli check`.

## Validation

The change is test-only and should pass the qualified-call visitor tests, the new bug-test syntax check, and the normal Go test suite.

hjotha:
Closes #447.

## Summary

- Update the qualified-call visitor regression to use a synthetic three-segment callee.
- Add a parser-level bug-test MDL file for the same IF-condition form.
- Remove the real project-style name from this test fixture while touching the coverage.

## Validation

- `make build`
- `go test ./mdl/visitor -run 'TestQualifiedCallInIfCondition|TestQualifiedCallPositionalArgs' -v`
- `./bin/mxcli check mdl-examples/bug-tests/qualified-function-call-expression.mdl`
- `make test`
- `make lint-go`

### DIFF
--- a/cmd/mxcli/tui/watcher.go
+++ b/cmd/mxcli/tui/watcher.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -78,6 +79,7 @@ func newWatcher(mprPath, contentsDir string, sender MsgSender) (*Watcher, error)
 
 func (w *Watcher) run(sender MsgSender) {
 	var debounceTimer *time.Timer
+	var debounceSeq atomic.Uint64
 
 	for {
 		select {
@@ -110,7 +112,11 @@ func (w *Watcher) run(sender MsgSender) {
 			if debounceTimer != nil {
 				debounceTimer.Stop()
 			}
+			seq := debounceSeq.Add(1)
 			debounceTimer = time.AfterFunc(watchDebounce, func() {
+				if debounceSeq.Load() != seq {
+					return
+				}
 				sender.Send(MprChangedMsg{})
 			})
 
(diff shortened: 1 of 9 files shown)


hjotha:

ako:
## Review: test: cover qualified function calls with three segments

Clean test-only PR. No blockers. Two minor issues.

---

### Minor 1 — bug test file name doesn't follow the issue-number convention

All existing bug test files use `{issue}-{description}.mdl` naming (e.g. `373-change-action-items-storage-list.mdl`). The new file is named `qualified-function-call-expression.mdl` with no issue prefix. Since this closes #447, it should be `447-qualified-function-call-expression.mdl`.

---

### Minor 2 — bug test MDL "After fix:" header implies this PR introduced a fix

The `bug-tests/` header follows the convention: describe the original symptom and the fix. But the grammar fix (widening `functionCall` to accept `qualifiedName`) happened in an earlier PR — this PR adds the regression test. A reader scanning the file might think the fix landed here. The header would be clearer as:

```
-- Status: working — regression guard added in #447.
```

or move the "After fix" wording to note the historical fix explicitly.

---

### Positives

- Removing `ControlCenterCommons.IsNotEmptyString` and `MxAdmin` — real project-style names shouldn't be in test fixtures (CLAUDE.md recurring finding #5).
- The grammar's `qualifiedName` rule (`identifierOrKeyword (DOT identifierOrKeyword)*`) already handles N segments, so three-segment coverage is correct with no grammar change needed.
- The new test exercises the three-segment form (`SyntheticRules.Strings.IsNotEmpty`) that the describer emits for `RuleSplitCondition` parameters, which is the scenario most likely to show up in real projects.
- `mxcli check` (not `exec`) is the right usage instruction for a parse-only regression test against an unresolvable function name.

PR #456, opened 2026-04-30
fix: stop nested splits at parent merge joins

### LINKED ISSUE #455: Nested split describe can consume parent continuation as a local join
## Problem

When a nested split is described inside a parent branch, the precomputed join for the nested split can point to a node that is only reachable after the parent merge. In that shape, `describe` treats parent continuation as part of the nested branch.

This can move shared continuation into an `else` block and change the apparent control-flow structure. For return-valued microflows, that can also make one path appear to miss its required return after a describe/exec/describe roundtrip.

## Expected

Nested split traversal should stop at the active parent merge when the computed nested join is downstream of that parent merge.

## Reproducer

A synthetic graph with:

- parent split merge A
- nested split computed join B
- A reaches B
- at least one nested branch reaches A

should describe the nested split against A, not consume B as a local nested join.

## Validation

Add regression coverage for resolving downstream nested joins to the parent merge and for keeping independent local nested joins unchanged.

hjotha:
Closes #455.

## Summary

- resolves nested split joins against the active parent merge when the computed nested join is downstream
- prevents the empty-then swap from running against an over-wide nested join
- adds synthetic coverage for downstream parent-merge joins and independent local nested joins

## Validation

- `make build`
- `make test`


hjotha:

hjotha:
Updated this branch with the current upstream main and a follow-up for two additional split-join traversal cases:

- enum splits where only a subset of branches continues into a shared tail now keep that tail outside the case bodies;
- nested branches that can bypass an early join no longer cause that early join to be selected as the parent split join.

Local validation after the merge: `go test ./mdl/executor`, `make build`, `make lint-go`, and `make test`.

ako:
## Review: fix: stop nested splits at parent merge joins

### Summary

Three coordinated fixes for the describe traversal: (1)  now filters out candidate joins that a nested branch can bypass, preventing the wrong (too-early) join from being selected as the parent split's merge point; (2)  detects when a computed nested merge is actually downstream of the parent merge and pins it to the parent instead; (3)  tightened to require right→left anchor indices alongside the Y-position alignment. Good test coverage for all three sub-problems.

**No blockers.**

---

### Moderate

** anchor tightening may not hold for Studio Pro-generated MPR files**

Before:
```go
return dest.GetPosition().Y == split.GetPosition().Y
```
After:
```go
return dest.GetPosition().Y == split.GetPosition().Y &&
    flow.OriginConnectionIndex == AnchorRight &&
    flow.DestinationConnectionIndex == AnchorLeft
```

 always writes , so mxcli-created MPR files satisfy the new constraint. The test was updated to add explicit anchors — confirming it previously passed with Go zero-values (0,0 = /). If Studio Pro writes unset (0,0) or different indices for horizontal boolean split flows, the guard detection silently stops working for described Studio Pro projects: the false branch no longer collapses into a guard continuation and instead emits an explicit . Recommend checking one Studio Pro-generated MPR with a guard-style boolean split before merging.

---

### Minors

**1. No full-traversal integration test for the issue #455 scenario**

The new  tests pin the helper in isolation and  pins the merge-selection logic. But there is no end-to-end  test that asserts the *described MDL output* for a nested IF whose local join is the parent split's join (the exact topology from #455). The unit tests give confidence in the building blocks; a round-trip test would confirm the wiring.

**2.  +  each perform O(N) graph walks per split in the hot traversal path**

For typical microflows this is fine, but for very large flows with many nested IFs the per-split O(N) walks compound to O(N²) per nesting level. Not a blocker — worth a comment in the code noting the complexity assumption.

---

### Positives

- Clean PR — no bundled unrelated changes.
-  /  refactoring turns inline copy-paste into named helpers, making the traversal code easier to follow.
-  directly pins the false-positive category that motivated the anchor tightening.
-  sort key in  is the right primary discriminator (nodes reachable from more branches are better join candidates than closer nodes reached by fewer).

Issue #468, opened 2026-05-01
Verify Forms$FormSettings.TitleOverride: nil against Studio Pro BSON


hjotha:
## Source

Follow-up from ako's review on merged PR #338 (fix: write valid show-page parameter mappings).

## Problem

PR #338 fixed \`Forms\$PageParameterMapping.Variable: nil\` because Studio Pro rejected null embedded objects on project load. \`Forms\$FormSettings.TitleOverride\` is written as \`nil\` in the same block (\`sdk/mpr/writer_microflow_actions.go\` around line 367). If TitleOverride is an embedded object, it needs the same empty-object treatment as Variable. If it is a string/expression scalar, \`nil\` is typically accepted.

Issue #295 (the reported bug) only mentioned Variable, so the current code may or may not cause a load failure depending on the Studio Pro version and the action context.

## Expected behavior

\`TitleOverride\` in an authored MPR loads in Studio Pro without an "invalid project" error across all supported Mendix versions.

## Proposed fix

1. Dump a Studio Pro-generated MPR with a show-page action and inspect the raw BSON for the TitleOverride field. Confirm type (scalar vs embedded object).
2. If embedded object: replace \`nil\` with an empty \`Forms\$TitleOverride\` (or whatever the correct type is) document, mirroring \`emptyPageVariable()\`.
3. If scalar: add a code comment documenting that null is intentional and verified.
4. Add a regression test either way.

## Related

- Reviewed in #338 by ako.

ako:
Resolved. Verified and fixed.

**Type confirmed** (proposed step 1): `Forms$FormSettings.TitleOverride` is an **embedded object** — a `Microflows$TextTemplate`, not a scalar (metamodel: `TitleOverride *MicroflowsTextTemplate`). So `nil` was indeed wrong (Studio Pro rejects null embedded objects), same class as the `Variable` fix in #338/#295.

**Fix** (proposed step 2): both engines now write an empty `Microflows$TextTemplate` object instead of nil —
- legacy: `emptyTextTemplate()` (`writer_microflow_actions.go`, `writer_widgets_action.go`, `writer_navigation.go`);
- modelsdk: `emptyTextTemplateToGen()` (`microflow_write.go`, `widget_write.go`).
Landed in `47821dc9` ("fix(#468): write TitleOverride as empty Microflows$TextTemplate, not nil") + `f9b202ff` (modelsdk).

[...]

Issue #463, opened 2026-05-01
Extract shared walkStatements helper for microflow AST walkers


hjotha:
## Source

Follow-up from ako's review on PR #337 (merged as part of open PR).

## Problem

Several microflow builder passes open-code the same recursive AST walk over MicroflowStatement bodies, each with a nearly identical switch on statement types. Concrete examples:

- \`planMicroflowCallOutputDeclarations\` (inner \`walk\`) in \`cmd_microflows_builder.go\`
- \`countMicroflowCallOutputs\` in the same file
- \`collectListInputVariables\` / \`collectObjectInputVariables\` in \`cmd_microflows_builder_graph.go\`
- \`statementVarRefs\` / \`statementsVarRefs\` in \`cmd_microflows_builder_varutil.go\`
- \`bodyReturns\` in \`validate_microflow.go\`

Each copy is ~40 lines of switch boilerplate with an \`ErrorHandling\` recurse arm. When a new statement type gains an \`ErrorHandling\` field or nested body, it must be added in every place — this is how the bugs in PR #337 (issue #405) and PR #362 surfaced.

## Expected behavior

A single shared helper:

\`\`\`go
func walkStatements(body []ast.MicroflowStatement, fn func(ast.MicroflowStatement))
\`\`\`

does the recursion (including \`IfStmt\`/\`LoopStmt\`/\`WhileStmt\` bodies and every \`ErrorHandling.Body\`). Each caller supplies the per-node callback only. Adding a new statement type with its own body requires exactly one edit in one place.

## Proposed fix

1. Introduce \`walkStatements(body, fn)\` in \`cmd_microflows_builder_varutil.go\` (or a new \`cmd_microflows_builder_walk.go\`).
2. Rewrite the existing walkers to call it with their per-node logic.
3. Keep \`statementVarRefs\` as is — that's the per-node logic, not the walker.

## Related

- Reviewed in #337 by ako.
- Directly related to gaps that produced #405, PR #362 follow-up, and the missing CreateListStmt / SendRestRequestStmt coverage.
- Part of #332 (microflow roundtrip stabilization tracker).

ako:
Status check (still open, not yet done) — leaving notes for whoever picks this up.

**Not implemented:** there is no `walkStatements` helper anywhere in `mdl/executor/`, and the walkers still each open-code their own recursion switch with the `ErrorHandling.Body` arm.

**Current candidates** (the duplication this issue targets):
[...]

PR #469, opened 2026-05-01
fix: address merged-PR ako follow-up polish items

hjotha:
Batches the remaining polish items ako flagged in reviews of PRs that already merged.

### PR #357 — Java return-type inference
- Remove the dead `case javaactions.ListType:` value form from `inferGenericJavaActionReturnType`. The parser always stores `ListType` as a pointer.
- Replace the \`exprToString\` → strip quotes → strip \`\$\` → look up in varTypes string-mangling with a direct `*ast.VariableExpr` type assertion. The old form only worked for simple variable references and silently produced garbage keys for any compound expression.
- Add `TestAddJavaAction_EntityReturnRegistersEntityType` — the `EntityType` case was the only return shape without a unit test.

### PR #363 — change refresh modifier
- Restore the CE0032 reference in the `addChangeObjectAction` comment. Developers searching for the Studio Pro error code no longer hit a dead end.

### PR #364 — enum split
- Collapse the dual `return true` branches of `isTerminalStmt`'s EnumSplitStmt arm into one `return true` and expand the comment documenting the intentional divergence from `bodyReturns` in `validate_microflow.go`.
- Add a doc comment on `hasExplicitFalseBranchAnchor` explaining the `AnchorTop → AnchorBottom` heuristic.
- Add `TestHasExplicitFalseBranchAnchor` covering nil, default right→left, split default bottom→top, authored top→bottom, and both single-sided-customised edge cases.

## Validation

- \`make build\`
- \`go test ./mdl/executor -run 'TestAddJavaAction_EntityReturn|TestHasExplicitFalseBranchAnchor' -count=1\`
- \`make test\`
- \`make lint-go\`

## Related

- Related follow-up issues: #463 (walkStatements refactor), #468 (TitleOverride nil verification).

### DIFF
--- a/cmd/mxcli/tui/watcher.go
+++ b/cmd/mxcli/tui/watcher.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -78,6 +79,7 @@ func newWatcher(mprPath, contentsDir string, sender MsgSender) (*Watcher, error)
 
 func (w *Watcher) run(sender MsgSender) {
 	var debounceTimer *time.Timer
+	var debounceSeq atomic.Uint64
 
 	for {
 		select {
@@ -110,7 +112,11 @@ func (w *Watcher) run(sender MsgSender) {
 			if debounceTimer != nil {
 				debounceTimer.Stop()
 			}
+			seq := debounceSeq.Add(1)
 			debounceTimer = time.AfterFunc(watchDebounce, func() {
+				if debounceSeq.Load() != seq {
+					return
+				}
 				sender.Send(MprChangedMsg{})
 			})
 
--- a/cmd/mxcli/tui/watcher_test.go
+++ b/cmd/mxcli/tui/watcher_test.go
@@ -35,10 +35,11 @@ func TestWatcherDebounce(t *testing.T) {
 	}
 	defer w.Close()
 
-	// Rapidly write 5 times — should debounce into a single message
+	// Rapidly write 5 times — should debounce into a single message.
+	// Keep the burst tighter than the debounce window so slow CI machines do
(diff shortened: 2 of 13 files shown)


hjotha:

ako:
## Review: fix: address merged-PR ako follow-up polish items

### Summary

Three targeted follow-ups for already-merged PRs. All three changes are correct:
-  now uses a direct  type assertion instead of string-mangling through  → strip quotes → strip  — the old path silently produced garbage keys for any compound expression.
- CE0032 reference restored to the  comment — good for searchability.
- Stale "Both paths return true" comment updated to accurately describe the single-path code.

**No blockers.**

---

### Moderate

**Multiple open PRs (#452, #460, #469) share identical bundled-in file changes**

, , , , and  appear in all three currently-open PRs with the same diff. Whichever merges first will create conflicts for the others. The merge order needs to be explicit, or the shared bundle should land in one dedicated PR so the others can be rebased cleanly.

---

### Minors

**1. PR description says "collapse the dual  branches" but the code already has one**

The diff shows the old comment text () alongside a single  context line. The collapse happened before this PR; the actual change is fixing the stale "both paths" comment that was left behind. The description is accurate about the *intent* but inaccurate about what the diff contains.

**2. Test doc comments name a reviewer**

Both new test functions open with . Per CLAUDE.md, comments shouldn't reference callers or reviewers — those belong in the PR description, not in code. Use the issue number alone, or drop the attribution and describe the invariant:

```go
// EntityType return must register the output variable as the entity's
// qualified name so downstream attribute access resolves correctly.
func TestAddJavaAction_EntityReturnRegistersEntityType(t *testing.T) {
```

**3. Bundled unrelated changes (recurring — #366, #370, #474, #448, #450, #452, #460, #469)**

Watcher debounce, nanoflow example cleanup, and Java action  are unrelated to the three follow-up items described in the PR scope. CLAUDE.md: "Each PR is scoped to a single feature or concern."

PR #474, opened 2026-05-01
fix: emit merge flow for empty ELSE bodies (CE0079)

hjotha:
## Summary

- The IF builder skipped the split→merge flow when the ELSE body was empty but the merge was still needed (THEN continues). Studio Pro rejected the graph with **CE0079** ("The 'false' condition value should be configured in properties for an outgoing flow") because the split had no outgoing flow for the `false` case.
- Mirrors the existing empty-THEN handling by emitting a direct split→merge flow with case `false`.
- Builds on the `IfStmt.HasElse` support now present on `main`, so the test distinguishes an omitted ELSE from an explicitly empty ELSE.

## Test plan

- [x] `make build`
- [x] `make test`
- [x] `make lint-go`
- [x] New builder unit test `TestIfEmptyElseBodyWithContinuingThenEmitsFalseFlowToMerge` asserts the split emits a `false` case flow into the merge
- [x] The regression test sets `HasElse: true`, so it exercises the explicit empty-ELSE path instead of the no-ELSE path

### DIFF
--- a/mdl/executor/cmd_microflows_builder_control.go
+++ b/mdl/executor/cmd_microflows_builder_control.go
@@ -272,6 +272,12 @@ func (fb *flowBuilder) addIfStatement(s *ast.IfStmt) model.ID {
 				}
 				applyUserAnchors(flow, originAnchor, destAnchor)
 				fb.flows = append(fb.flows, flow)
+			} else {
+				// Explicit empty ELSE body: the split still needs a configured
+				// false case when both branches converge at the merge.
+				flow := newDownwardFlowWithCase(splitID, mergeID, "false")
+				applyUserAnchors(flow, falseBranchAnchor, falseBranchAnchor)
+				fb.flows = append(fb.flows, flow)
 			}
 		} else if elseReturns && needMerge {
 			fb.addPendingErrorHandlerFlowTo(mergeID)
--- a/mdl/executor/cmd_microflows_builder_empty_else_merge_test.go
+++ b/mdl/executor/cmd_microflows_builder_empty_else_merge_test.go
@@ -1,0 +1,63 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package executor
+
+import (
+	"testing"
+
+	"github.com/mendixlabs/mxcli/mdl/ast"
+	"github.com/mendixlabs/mxcli/sdk/microflows"
+)
+
+// TestIfEmptyElseBodyWithContinuingThenEmitsFalseFlowToMerge is a regression
+// guard for CE0079 ("The 'false' condition value should be configured in
+// properties for an outgoing flow").
+//
+// Pattern: an IF with THEN that continues and an explicitly empty ELSE body.
+// Both branches must feed a merge, but an empty ELSE has no lastElseID to wire.
+func TestIfEmptyElseBodyWithContinuingThenEmitsFalseFlowToMerge(t *testing.T) {
+	fb := &flowBuilder{
+		spacing:  HorizontalSpacing,
+		measurer: &layoutMeasurer{},
+	}
+
+	fb.addIfStatement(&ast.IfStmt{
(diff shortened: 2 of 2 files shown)


hjotha:

ako:
## Review: fix: emit merge flow for empty ELSE bodies (CE0079)

Two blockers before this can merge.

---

### Blocker 1 — test doesn't exercise the new code

The test creates `IfStmt` with no `HasElse` and empty `ElseBody`:

```go
fb.addIfStatement(&ast.IfStmt{
    Condition: ...,
    ThenBody: []ast.MicroflowStatement{&ast.LogStmt{...}},
    // ElseBody deliberately empty
})
```

With the current `hasElseBody := len(s.ElseBody) > 0` (line 34 of `cmd_microflows_builder_control.go`), this evaluates to `false`, so execution takes the **IF-WITHOUT-ELSE** branch at line 304 — not the new code path. The IF-WITHOUT-ELSE branch already emits `newHorizontalFlowWithCase(splitID, mergeID, "false")` when `needMerge = true`. The test was already passing before this PR.

The new `else` block lives inside `if hasElseBody { ... if !elseReturns && needMerge { ... } }`. It can only be reached when `hasElseBody = true` **and** `lastElseID = ""`, which requires an explicitly empty ELSE body that the visitor records with `HasElse: true` — the `HasElse` field added by PR #366.

The test comment acknowledges this ("see below") but there is no second test in the PR.

A test that exercises the new code must set both `HasElse: true` and an empty `ElseBody` in the AST, and **also** requires the `hasElseBody` calculation to be updated to `s.HasElse || len(s.ElseBody) > 0` (see Blocker 2).

---

### Blocker 2 — hidden dependency on PR #366 (`HasElse` + updated `hasElseBody`)

The new `else` block is dead code in the current codebase. Until PR #366 changes `hasElseBody` to `s.HasElse || len(s.ElseBody) > 0`, no MDL input can produce `hasElseBody = true` with an empty ELSE body. The grammar produces `s.ElseBody = nil` or `[]` for `if cond then ... else end if`, so `len(s.ElseBody) = 0` → `hasElseBody = false` → IF-WITHOUT-ELSE path.

This PR should either:
- Be merged after PR #366, with the test using `HasElse: true` directly in the AST
- Include PR #366's `hasElseBody` fix as part of this PR
- At minimum document the dependency explicitly in the PR description

---

### Minor — bundled unrelated changes

Watcher debounce (`sync/atomic`), Java action `RETURNS Void` handling, nanoflow doctype cleanup, enum parameter roundtrip test — none of these relate to the CE0079 fix. Same pattern seen in PRs #366, #370. These belong in separate PRs.

---

### The fix itself is correct

For the scenario it targets (empty ELSE with `hasElseBody = true`), adding `newDownwardFlowWithCase(splitID, mergeID, "false")` inside `if !elseReturns && needMerge` mirrors the existing empty-THEN logic exactly and will produce the correct BSON for Studio Pro. The fix just can't be reached or tested until the `HasElse` + `hasElseBody` prerequisite from PR #366 is in place.

hjotha:
Addressed the review blocker:

- stacked this branch explicitly on #366, because the fix needs `IfStmt.HasElse` to distinguish omitted ELSE from explicitly empty ELSE;
- updated the regression test to set `HasElse: true`, so it now exercises the explicit empty-ELSE path;
- reran `go test ./mdl/executor -run TestIfEmptyElseBodyWithContinuingThenEmitsFalseFlowToMerge -v`, `make build`, `make test`, and `make lint-go`.

hjotha:
Follow-up: after #366 landed on `main`, I rebuilt this branch directly on current `origin/main` and force-pushed a clean single-commit PR. The diff is now limited to the empty-ELSE builder fix plus its regression test; the test still sets `HasElse: true` to exercise the explicit empty-ELSE path.

PR #475, opened 2026-05-01
fix: always emit merge for continuing inheritance split branches (CE0079)

hjotha:
## Summary

- `addStructuredInheritanceSplit` used to take a no-merge shortcut when exactly one non-split branch continued and the rest terminated — it wired the parent's next statement directly to that branch's tail. That caused re-describe to emit the parent continuation **inside** the case body, and Studio Pro raised **CE0079** ("condition value should be configured for an outgoing flow") on terminating branches because their cases had no merge to converge on.
- Fix: whenever any branch survives, emit the `ExclusiveMerge` and wire surviving tails through it. Parent continuation then resumes after the merge as expected.

## Depends on

**Stacked on #365** (`submit/microflow-inheritance-split-cast`). That PR introduces `addStructuredInheritanceSplit`; this PR patches the no-merge shortcut added in the same function. Merge #365 first, then this branch merges cleanly on top. The PR is opened against `main` because `#365` lives on a fork branch.

## Test plan

- [x] `go build ./...` and `go test ./mdl/executor/ -count=1` pass
- [x] `go build -tags integration ./...` builds clean
- [x] New builder unit test `TestInheritanceSplitAlwaysEmitsMergeWhenBranchContinues` asserts the merge is created when one branch continues and the rest terminate

### DIFF
--- a/.claude/skills/mendix/write-microflows.md
+++ b/.claude/skills/mendix/write-microflows.md
@@ -340,6 +340,23 @@ end case;
 
 `(empty)` represents an unset enumeration value. Multiple values can share one `when` branch by separating them with commas. Case values are bare identifiers — do **not** quote them.
 
+### Type Split And Cast Statements
+
+Use `split type` when a microflow branches on an object's runtime specialization.
+Use `cast` inside a type branch to create the specialized variable used by the branch body.
+
+```mdl
+split type $Input
+case Sample.SpecializedInput
+  cast $SpecificInput;
+  return true;
+else
+  return false;
+end split;
+```
+
+`case` values are qualified entity names. The optional `else` branch handles objects that do not match any listed specialization.
+
 ### LOOP Statements
 
 ```mdl
--- a/.claude/skills/mendix/write-microflows.md
+++ b/.claude/skills/mendix/write-microflows.md
@@ -357,6 +357,8 @@ end split;
 
 `case` values are qualified entity names. The optional `else` branch handles objects that do not match any listed specialization.
 
+**`cast` only stores the output variable.** Studio Pro persists Microflows$CastAction with a single `VariableName` field — the source variable is implicit (the type-split's input). Use `cast $SpecificName;` to give the specialized variable its name. The two-variable form `$Output = cast $Source;` parses but `$Source` is dropped on roundtrip; prefer the single-variable form.
+
 ### LOOP Statements
 
 ```mdl
--- a/docs/01-project/MDL_QUICK_REFERENCE.md
+++ b/docs/01-project/MDL_QUICK_REFERENCE.md
@@ -243,6 +243,8 @@ authentication basic, session
 | Free annotation | `@annotation 'text'` before `@position(...)` | Free-floating visual note preserved by order |
 | IF | `if condition then ... [else ...] end if;` | |
 | Enum split | `case $Var when Value then ... end case;` | Enumeration decision branches |
+| Type split | `split type $Var case Module.Entity ... end split;` | Runtime specialization branches |
+| Cast | `cast $SpecificVar;` | Downcast inside a type split branch |
 | LOOP | `loop $item in $list begin ... end loop;` | FOR EACH over list |
 | WHILE | `while condition begin ... end while;` | Condition-based loop |
 | Return | `return $value;` | Required at end of every flow path |
--- a/docs/11-proposals/PROPOSAL_microflow_inheritance_split_statement.md
+++ b/docs/11-proposals/PROPOSAL_microflow_inheritance_split_statement.md
@@ -1,0 +1,35 @@
+# Proposal: Microflow Inheritance Split And Cast Statements
+
+Status: Draft
+
+## Summary
+
+Add round-trip MDL support for type-based microflow decisions and cast actions:
+
+```mdl
+split type $Input
+case Sample.SpecializedInput
+  cast $SpecificInput;
+else
+  return false;
+end split;
+```
+
+## Motivation
+
+Studio Pro represents specialization/type decisions as `InheritanceSplit` objects and stores downcasts as `CastAction` activities. Without first-class MDL statements, `describe` can only emit unsupported comments or incomplete split output, and `exec` cannot rebuild the same graph.
+
+## Semantics
+
+`split type $Var` evaluates the runtime specialization of an object variable. Each `case Module.Entity` branch corresponds to an outgoing sequence flow with an `InheritanceCase`. The optional `else` branch maps to the outgoing flow without an inheritance case.
+
+`cast $Output` emits a `CastAction` that produces the downcast variable. `$Output = cast $Input` is accepted for source-preserving authoring, but current Mendix BSON stores the generated cast variable as the primary persisted field.
+
+## Tests And Examples
+
+`mdl-examples/doctype-tests/inheritance_split_statement.test.mdl` demonstrates the syntax. Go regression tests cover parser construction, builder output, describer output, validation recursion, and BSON writer support for inheritance case values and cast actions.
+
+## Open Questions
+
+- Should `exec` validate `case Module.Entity` against the project's specialization hierarchy when connected?
+- Should the source-preserving `$Output = cast $Input` form round-trip both variable names once the underlying BSON fields are confirmed for all supported Mendix versions?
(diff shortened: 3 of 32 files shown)


hjotha:

ako:
## Review

### Blockers

**B1 — CastAction BSON writer uses wrong field name (`VariableName` → `OutputVariableName`)**

In `sdk/mpr/writer_microflow_actions.go`, the `writeCastAction` function serialises the output variable as:
```go
{Key: "VariableName", Value: a.OutputVariable}
```
Studio Pro writes `OutputVariableName` for `CastAction`. Using `VariableName` will corrupt any project that contains an INHERITANCE SPLIT with `cast` branches — the output variable will be silently lost when the file is re-opened in Studio Pro.

Fix:
```go
{Key: "OutputVariableName", Value: a.OutputVariable}
```
Verify against a Studio Pro-produced MPR: `grep -i OutputVariableName` in the BSON dump of a project containing a CastAction.

---

**B2 — Go type switch regression: `EnumSplitStmt` case body silently stolen by `InheritanceSplitStmt`**

In `mdl/executor/validate.go`, `collectFromStatements` was extended to handle `InheritanceSplitStmt`. The new case was inserted between `case *ast.EnumSplitStmt:` and the loop body that was originally its handler:

```go
case *ast.EnumSplitStmt:          // ← no body — falls through to nothing in Go type switches
case *ast.InheritanceSplitStmt:   // ← gets the loop body
    for _, cse := range s.Cases {
        collectFromStatements(...)
    }
```

In Go, `type switch` cases **do not fall through** — an empty case is simply a no-op. `EnumSplitStmt` is now silently ignored by reference validation, meaning callers of a referenced microflow inside an enum split branch will go unchecked.

Fix — give each case its own complete block:
```go
case *ast.EnumSplitStmt:
    for _, cse := range s.Cases {
        collectFromStatements(ctx, cse.Statements, refs)
    }
case *ast.InheritanceSplitStmt:
    for _, cse := range s.Cases {
        collectFromStatements(ctx, cse.Statements, refs)
    }
```

---

### Moderate

**M1 — `InheritanceCase` `"Value"` field: verify Studio Pro's actual BSON field name**

`writeInheritanceSplitCase` writes the entity reference as `{Key: "Value", Value: ...}`. This should be verified against a Studio Pro-produced MPR. The field is likely `"Entity"` (a binary UUID) rather than a plain string `"Value"`. A wrong field name here produces CE0079 or silent data loss on re-open.

Action: dump a project containing an INHERITANCE SPLIT from Studio Pro (`mx dump-mpr --unit-type=Microflows$InheritanceSplit`) and compare the BSON field name.

**M2 — No DESCRIBE roundtrip test (Recurring Finding #4)**

The PR adds `CREATE INHERITANCE SPLIT` but there is no test that formats an `InheritanceSplitStmt` → MDL string → parses it back and confirms no error. Grammar gaps have gone undetected before until human review (RF #4). Add a roundtrip test in `mdl-examples/doctype-tests/` or a unit test that calls `Describe` on a constructed `InheritanceSplitStmt`.

---

### Minor

**m1 — `qualifiedNameString` duplicates `ast.QualifiedName.String()`**

The helper introduced in this PR (`qualifiedNameString(qn ast.QualifiedName) string`) produces `Module.Name` — identical to what `qn.String()` already does. Remove the helper and call `qn.String()` directly.

---

### Recurring Finding added

**#13 — Go type switch: inserting a new `case TypeB:` between `case TypeA:` and its body silently empties TypeA**

Unlike regular `switch`, Go's `type switch` has no fallthrough — an empty case is a silent no-op. When extending a type switch by inserting a new case between an existing case label and its body, the existing case loses its handler. This pattern looks like code reuse but is actually a regression. Always give each type switch case its own complete block.

hjotha:
Thanks @ako — addressed in `cdd44436`.

**B2 — empty type switch case (real bug, fixed)**

You're absolutely right. `flowRefCollector.collectFromStatements` had `case *ast.EnumSplitStmt:` with an empty body immediately followed by `case *ast.InheritanceSplitStmt:` whose body walked `s.Cases` and `s.ElseBody`. Go type switches don't fall through, so `EnumSplitStmt` was a silent no-op — every microflow / page / java action / entity reference inside an enum-split case body or its else body escaped reference validation.

Fix: gave EnumSplitStmt its own complete walk identical to InheritanceSplitStmt. Two new tests pin this so the regression cannot return:

- `TestValidateMicroflowReferences_DescendsIntoEnumSplitCases` — missing microflow inside a `case ... when ... then` body must error.
- `TestValidateMicroflowReferences_DescendsIntoEnumSplitElse` — missing microflow inside the `else` body of an enum split must error.

**B1 — CastAction `VariableName`: review premise was wrong**

I dumped a real Studio Pro MPR (Control Centre, Mendix 9.24) and Studio Pro stores `CastAction` as:

```
Microflows$CastAction:
  ErrorHandlingType = Rollback
  VariableName = Account
```

`OutputVariableName` is **not** present. Three independent CastActions in the dump (`Administration.ManageMyAccount`, `Cloud.RetrieveEnvironments`, `Groups.ACT_RemoveAccessibleApp`) all use `VariableName`. The writer's current `VariableName` key is correct.

Rather than change correct code, I added `TestSerializeCastAction_UsesVariableNameFieldKey` to pin the field name and prevent a future "fix" based on the same premise. The parser still falls back to `VariableName` if `OutputVariableName` is absent (introduced in PR #365), so any future Mendix version that ever switches naming would still parse.

The corresponding `InheritanceCase.Value` finding is on PR #365 (commit `5f479036`).

ako:
## Code Review

### Context

This PR is stacked on #365 and the diff carries the entire inheritance split feature (1574 additions). The actual CE0079 fix is a small targeted change inside `addStructuredInheritanceSplit` — review the two layers separately.

### CE0079 Fix

Correct and well-explained. The old shortcut wired the parent's next statement directly to the sole continuing branch tail, bypassing a merge node. Studio Pro then had nothing for terminating branches to converge on and rejected their outgoing flows with CE0079. Always emitting `ExclusiveMerge` when `len(branchTails) > 0` mirrors the same invariant `addEnumSplit` already enforces. `TestInheritanceSplitAlwaysEmitsMergeWhenBranchContinues` directly validates the fix.

### Full-stack coverage (from #365)

All layers wired: grammar → visitor → AST → executor → writer. `flowRefCollector`, `walkBody`, `bodyReturns`, `checkBranchScoping`, and `isTerminalStmt` all handle `InheritanceSplitStmt` consistently. The regression test guarding against the switch-fallthrough traversal omission is a good addition.

### Required before merge

- **No `mdl-examples/bug-tests/475-*.mdl`** — the CE0079 scenario (one case returns, one case continues through merge) is covered in Go unit tests but there is no MDL script for running against a real project. Required by checklist.

### Worth discussing

- **`$Output = cast $Source` loses `$Source` on roundtrip** — the writer emits only `VariableName: a.OutputVariable` and discards `ObjectVariable`. The proposal doc acknowledges this: _"current Mendix BSON stores the generated cast variable as the primary persisted field."_ Since the grammar accepts the two-variable form, authors may expect a lossless roundtrip. Either disallow `$Output = cast $Source` in the grammar, or add a note in the skill doc setting the expectation.

### Minor

- `allBranchesReturn := len(s.Cases) > 0 && len(s.ElseBody) > 0` as the initial value is correct but reads as surprising — a comment on why it starts as `true` only when both sides are non-empty would help.
- An empty case body (no return, no actions) produces a `fromSplit: true` tail that wires the split directly to the merge. Valid, but likely an authoring mistake — worth a validator warning.

hjotha:
Thanks @ako — addressed in `c8598ea3`.

**Required**

- **`mdl-examples/bug-tests/475-inheritance-split-continuing-branch-merge.mdl`** — new bug-test scripted as `MF_TypedDispatch`: `split type $obj` with one case that continues into a post-split log activity and one case that terminates with `return`. Before the CE0079 fix this combination tripped CE0079 on the terminating branch; after the fix `mx check` reports 0 errors and the post-split log activity stays outside both case bodies through a describe/exec roundtrip.

**Worth discussing — `$Output = cast $Source` lossiness**

Documented in `.claude/skills/mendix/write-microflows.md`: the Type Split And Cast section now warns that `cast` only persists the output variable in BSON (`Microflows$CastAction.VariableName`) and that the two-variable form `$Output = cast $Source;` drops `$Source` on roundtrip. Authors who expected lossless behaviour from the two-variable form now have the expectation set explicitly.

ako:
## Updated Review

Both required items from the previous review are addressed:

- **Bug test added** — `475-inheritance-split-continuing-branch-merge.mdl` covers the CE0079 scenario with clear symptom, fix description, and usage instructions.
- **Cast roundtrip documented** — `write-microflows.md` now explicitly warns that `$Source` is dropped in the two-variable form and recommends `cast $SpecificName;`. `MDL_QUICK_REFERENCE.md` updated too.

Two minor points remain, neither blocking:

- `allBranchesReturn := len(s.Cases) > 0 && len(s.ElseBody) > 0` still has no comment explaining the pessimistic-start logic.
- An empty case body (direct split→merge wire) has no validator warning — probably an authoring mistake but silently accepted.

Ready to merge once #365 lands.

PR #488, opened 2026-05-02
fix: emit swapped split headers only once

### LINKED ISSUE #487: DESCRIBE emits duplicate IF headers for empty-then split normalization
## Problem

When DESCRIBE normalizes an empty-then split by negating the condition and swapping branches, it can emit the original split header before the normalization decision and then emit the normalized header again.

The generated MDL is syntactically invalid because the first header has no matching body/END IF. EXEC then fails with parse errors such as unmatched IF/ELSE/CASE tokens.

## Expected behavior

DESCRIBE should emit exactly one split header: the final normalized header after branch analysis.

## Fix direction

Delay non-enum split header emission until after empty-then swap analysis in both top-level traversal and nested traversal.

## Validation

Add synthetic tests covering top-level and nested empty-then split normalization, asserting the IF header is emitted exactly once.

hjotha:
Closes #487.

## Summary

- Delay split header emission until after empty-then swap analysis.
- Cover top-level and nested traversal with synthetic regression tests.
- Includes merge of #412 to stabilize CI baseline across active PR branches.

## Validation

- make build
- make lint-go
- make test
- Representative DESCRIBE output was executed back into a copied MPR successfully.

### DIFF
--- a/cmd/mxcli/tui/watcher.go
+++ b/cmd/mxcli/tui/watcher.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -78,6 +79,7 @@ func newWatcher(mprPath, contentsDir string, sender MsgSender) (*Watcher, error)
 
 func (w *Watcher) run(sender MsgSender) {
 	var debounceTimer *time.Timer
+	var debounceSeq atomic.Uint64
 
 	for {
 		select {
@@ -110,7 +112,11 @@ func (w *Watcher) run(sender MsgSender) {
 			if debounceTimer != nil {
 				debounceTimer.Stop()
 			}
+			seq := debounceSeq.Add(1)
 			debounceTimer = time.AfterFunc(watchDebounce, func() {
+				if debounceSeq.Load() != seq {
+					return
+				}
 				sender.Send(MprChangedMsg{})
 			})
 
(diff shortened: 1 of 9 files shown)


hjotha:

ako:
## Review: fix: emit swapped split headers only once

### Summary

Minimal, correct fix. The boolean split path in both `traverseFlow` and `traverseFlowUntilMerge` had two `if stmt != ""` emission blocks: one before `findBranchFlows`/swap analysis and one after. When the empty-then swap fired, the original condition was emitted first and then the negated condition was emitted again, producing duplicate `if ...` headers and invalid MDL. The fix removes the pre-swap block in both functions, leaving only the post-swap emission.

**No blockers.**

---

### Minors

**1. Bundled unrelated changes (recurring — #366, #370, #474, #448, #450, #452, #460, #469, #488)**

Watcher debounce, nanoflow example cleanup, Java action `RETURNS Void`, and roundtrip test changes are all unrelated to the split-header fix. CLAUDE.md: "Each PR is scoped to a single feature or concern."

---

### Positives

- The diff is ten removed lines across two symmetric locations — hard to get wrong.
- `findBranchFlows` explicitly handles `EnumerationCase{Value: "true/false"}` (verified in source), so the new test's use of that variant for a boolean split is valid — the comment in the source confirms it handles every CaseValue variant the parser produces.
- The existing `TestTraverseFlow_EmptyThenSwap` test gained an exact count assertion (`strings.Count(output, "if ") != 1`) that directly pins the regression.
- `TestTraverseFlowUntilMerge_NestedEmptyThenSwapEmitsHeaderOnce` covers the symmetric `traverseFlowUntilMerge` path with a clear topology comment.

PR #490, opened 2026-05-02
fix: scope duplicate output checks by branch path

### LINKED ISSUE #489: Duplicate microflow output checks should be scoped to exclusive control-flow paths
## Summary

`mxcli check` and `describe microflow` duplicate-output diagnostics currently treat output variable names as globally scoped across the entire microflow. That produces false positives for outputs declared in mutually exclusive branches where only one path can execute.

## Expected behavior

Linear duplicate declarations should still fail or warn because both declarations can execute on the same path.

Outputs with the same name in sibling IF/ELSE or enum-split branches should not be reported as CE0111-style duplicates when those branches are mutually exclusive.

## Actual behavior

The validation path accumulates branch declarations into one shared map, and the DESCRIBE warning path counts output names across the full object collection. Both checks can flag branch-local output variables as invalid even when each individual control-flow path is valid.

## Proposed fix

Validate branch bodies with cloned variable scopes and make DESCRIBE duplicate-output warnings walk sequence-flow paths with path-local output state.

hjotha:
Closes #489.

## Summary

- validate IF and enum-split branch bodies with cloned output-variable scopes
- make duplicate-output DESCRIBE warnings track variables per normal sequence-flow path
- keep linear duplicate-output errors/warnings intact while avoiding false positives for mutually exclusive branches

## Validation

- make build
- make test
- make lint-go

### DIFF
--- a/mdl/executor/cmd_microflows_builder_validate.go
+++ b/mdl/executor/cmd_microflows_builder_validate.go
@@ -67,6 +67,14 @@ func (fb *flowBuilder) validateStatements(stmts []ast.MicroflowStatement) {
 	}
 }
 
+func (fb *flowBuilder) validateScopedStatements(stmts []ast.MicroflowStatement) {
+	scoped := *fb
+	scoped.varTypes = cloneStringMap(fb.varTypes)
+	scoped.declaredVars = cloneStringMap(fb.declaredVars)
+	scoped.validateStatements(stmts)
+	fb.errors = scoped.errors
+}
+
 // validateStatement validates a single statement for semantic errors.
 func (fb *flowBuilder) validateStatement(stmt ast.MicroflowStatement) {
 	switch s := stmt.(type) {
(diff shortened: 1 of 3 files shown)


ako:
## Review: fix: scope duplicate output checks by branch path

**Verdict: Approve with minor notes**

### What the PR does
Two coordinated fixes for false-positive duplicate output variable warnings/errors when the same output variable name is used in mutually exclusive branches (IF/ELSE, enum split cases):

1. **`validateScopedStatements`** — clones `varTypes` and `declaredVars` before validating branch bodies so cross-branch variable declarations don't bleed into sibling branches.
2. **`duplicateOutputVariableWarnings` rewrite** — replaces flat occurrence-counting with a path-following graph traversal that clones `seen` and `path` maps at each step, making each execution path independent.

### No blockers

### Minor issues

**O(N²) map cloning in `duplicateOutputVariableWarnings`**
`cloneIDBoolMap(path)` is called at every node in the graph. For a linear flow of N nodes this creates N maps of sizes 1, 2, ..., N — O(N²) total allocations. For typical microflows (tens of nodes) this is fine, but worth a comment if microflows with hundreds of generated nodes become common.

**`starts` fallback uses all objects**
When no `StartEvent` is found, `starts` is seeded from all `NodesByID` values. For a large microflow this triggers O(N) separate DFS traversals from every node, many of which will be immediately terminated by the `path` guard. Correct and safe, just slightly surprising — a brief comment would help future readers.

### Positive notes
- Clean PR: no bundled unrelated changes (unlike several recent PRs on this branch — this is the right scoping)
- Four focused tests that directly reproduce the reported false-positive scenarios
- `validateScopedStatements` is the minimal correct fix (8 lines, no API surface change)
- Path-scoped traversal correctly models exclusive-branch semantics: a variable assigned in the `true` branch cannot be "seen again" in the `false` branch

PR #492, opened 2026-05-02
fix: emit expression case values for IF flows

### LINKED ISSUE #491: IF sequence flows should emit ExpressionCase values
## Summary

IF sequence flows currently encode true/false branches only as enumeration-style case values with the strings `true` and `false`. That matches Mendix 9 storage but conflates IF branch cases with real enum split cases in mxcli's in-memory model.

## Expected behavior

mxcli should distinguish IF branch cases from enum split cases internally, while still writing the storage representation expected by the target Mendix version.

## Actual behavior

mxcli builds IF flows with the same case representation used for enum split values. That makes downstream validation/traversal logic harder to reason about and can contribute to false duplicate-variable consistency handling for mutually exclusive IF paths.

## Proposed fix

Represent `true` and `false` IF flow cases as expression cases internally, keep enumeration case values for real enum split values, and serialize expression cases using the target Mendix version's expected storage shape.

hjotha:
Closes #491.

## Summary

- represent true/false IF sequence-flow cases as expression cases internally
- keep enumeration case values for real enum split values
- serialize expression cases as expression cases for modern MPRs
- downgrade expression cases to enumeration-style values when writing Mendix 9 projects
- update guard and nested enum-split tests to assert the correct case semantics

## Validation

- make build
- make test
- make lint-go

### DIFF
--- a/mdl/executor/cmd_microflows_builder_enum_split_test.go
+++ b/mdl/executor/cmd_microflows_builder_enum_split_test.go
@@ -111,7 +111,7 @@ func TestEnumSplitNestedEmptyThenBranchKeepsContinuationCase(t *testing.T) {
 		if flow.OriginID != nestedSplitID {
 			continue
 		}
-		if value, ok := enumCaseValue(flow); ok && value == "true" {
+		if flowCaseString(flow.CaseValue) == "true" {
 			if _, ok := objects[flow.DestinationID].(*microflows.ExclusiveMerge); ok {
 				return
 			}
--- a/mdl/executor/cmd_microflows_builder_flows.go
+++ b/mdl/executor/cmd_microflows_builder_flows.go
@@ -653,10 +653,7 @@ func newHorizontalFlow(originID, destinationID model.ID) *microflows.SequenceFlo
 // newHorizontalFlowWithCase creates a horizontal SequenceFlow with a boolean case value (for splits)
 func newHorizontalFlowWithCase(originID, destinationID model.ID, caseValue string) *microflows.SequenceFlow {
 	flow := newHorizontalFlow(originID, destinationID)
-	flow.CaseValue = microflows.EnumerationCase{
-		BaseElement: model.BaseElement{ID: model.ID(types.GenerateID())},
-		Value:       caseValue, // "true" or "false" as string
-	}
+	flow.CaseValue = caseValueForFlow(caseValue)
 	return flow
 }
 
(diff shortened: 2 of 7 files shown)


ako:
## Review: fix: emit expression case values for IF flows

**Verdict: Approve with minor notes**

### What the PR does

Fixes issue #491: mxcli was serializing IF-branch sequence flows as `Microflows$EnumerationCase` (string `"true"`/`"false"`) when Studio Pro 10+ expects `Microflows$ExpressionCase` (field `Expression`). The fix introduces `caseValueForFlow` to correctly produce `*ExpressionCase` for boolean branches and `EnumerationCase` for real enum values, with a version-gate in `buildSequenceFlowCase` that downgrades `ExpressionCase → EnumerationCase` when writing Mendix 9 projects.

### No blockers

### Moderate concern

**No DESCRIBE roundtrip test (Recurring Finding #4)**

The PR adds unit tests for `caseValueForFlow` and `buildSequenceFlowCase`, but no test that builds a microflow with an IF split, serializes it, re-parses it, and confirms DESCRIBE produces correct MDL. This is exactly the gap the roundtrip finding covers. A failing Studio Pro project won't fail `make test`, so the regression can resurface silently if the new case type is mishandled anywhere in the show/format path.

Suggested: add a test in `mdl/executor/` that creates a microflow with an IF branch, calls the BSON serializer, re-parses, then calls `showMicroflowActivities` and asserts the output contains `if ... then` rather than garbage.

### Minor issues

**`flowCaseString` handles only pointer `*ExpressionCase`, not value form**

The updated `flowCaseString` test helper only matches `*microflows.ExpressionCase`. If any test creates a value-form `ExpressionCase{}` directly and passes it through `flowCaseString`, it silently returns `""`. In practice all production paths (`caseValueForFlow`, parser) produce pointers, so this is unlikely to bite, but it's inconsistent with how `EnumerationCase` is handled (both forms).

### Positive notes
- Clean PR — no bundled unrelated changes
- The existing `findBranchFlows` already handled `*ExpressionCase` pre-PR; `BooleanCase` handling retained for legacy MPR files
- Version gate (`majorVersion <= 9`) correctly aligns with the `NewCaseValue` vs `CaseValues[]` serialization split already in the writer
- Two new writer tests explicitly cover Mx9 downgrade and Mx10+ path — this is exactly the right level of coverage for version-specific serialization

PR #494, opened 2026-05-02
fix: parse negative annotation coordinates

### LINKED ISSUE #493: Negative @position coordinates are parsed as zero
## Problem

`@position` accepts integer coordinates, including negative canvas coordinates emitted by `describe` for existing microflows. The parser accepts a leading minus syntactically, but the visitor currently reads only direct `NUMBER_LITERAL` nodes when converting annotation parameters to integers.

Because `-150` is parsed as a unary expression, the coordinate silently falls back to `0`. A describe/exec/describe cycle can therefore move activities from negative canvas positions to zero.

## Expected behavior

Negative integer annotation coordinates should round-trip as authored:

```mdl
@position(-150, -210)
log info node 'SyntheticLog' 'message';
```

should keep `X = -150` and `Y = -210` in the AST and generated microflow object.

## Proposed fix

Parse the raw annotation parameter text as an integer before falling back to the existing literal-only path. This preserves signed integer coordinates without evaluating arbitrary annotation expressions.

hjotha:
Closes #493.

## Summary

- preserves signed integer values in annotation parameters such as `@position(-150, -210)`
- keeps the existing literal-only fallback for non-signed numeric annotation parameters
- adds a synthetic microflow visitor regression test for negative coordinates

## Validation

- `go test ./mdl/visitor -run 'NegativeCoordinates|PositionAnnotation' -v`
- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/mdl/visitor/visitor_entity.go
+++ b/mdl/visitor/visitor_entity.go
@@ -497,6 +497,9 @@ func parseAnnotationParamInt(ctx parser.IAnnotationParamContext) int {
 	}
 	paramCtx := ctx.(*parser.AnnotationParamContext)
 	if valueCtx := paramCtx.AnnotationValue(); valueCtx != nil {
+		if val, err := strconv.Atoi(strings.TrimSpace(valueCtx.GetText())); err == nil {
+			return val
+		}
 		annValue := valueCtx.(*parser.AnnotationValueContext)
 		if lit := annValue.Literal(); lit != nil {
 			litCtx := lit.(*parser.LiteralContext)
--- a/mdl/visitor/visitor_test.go
+++ b/mdl/visitor/visitor_test.go
@@ -1804,6 +1804,31 @@ END;`
 	}
 }
 
+func TestMicroflowPositionAnnotationAcceptsNegativeCoordinates(t *testing.T) {
+	input := `CREATE MICROFLOW Synthetic.Check ()
+BEGIN
+  @position(-150, -210)
+  LOG INFO NODE 'SyntheticLog' 'message';
+END;`
+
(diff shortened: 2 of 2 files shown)


ako:
## Review: fix: parse negative annotation coordinates

**Verdict: Approve**

### What the PR does

Fixes issue #493: `@position(-150, -210)` annotations were silently returning `0` for both coordinates. The existing `parseAnnotationParamInt` only followed the `Literal → NUMBER_LITERAL` path, which doesn't match negative numbers. A value like `-150` is parsed by the grammar through the `expression` rule (unary minus + integer), so `annValue.Literal()` returns `nil` and the function fell through to `return 0`.

The fix calls `strconv.Atoi(strings.TrimSpace(valueCtx.GetText()))` first — ANTLR's `GetText()` concatenates all child token texts, producing `"-150"` for the expression case — then keeps the existing literal fallback path for any value `Atoi` can't handle.

### No blockers

### Minor observation

The `TrimSpace` call handles leading/trailing whitespace on the raw text, but not intra-token whitespace like `- 150` (space between minus and digits). That would cause `Atoi` to fail and fall through to the literal path, returning 0. In practice annotation params written with `- 150` are unlikely, and the `expression` grammar rule lexes `-150` without intermediate whitespace. Worth a brief comment if this code is touched again, but not a blocker.

### Positive notes
- Minimal, focused fix — 3 lines of production code, no grammar change required
- New test directly reproduces the bug and asserts exact coordinate values
- The early-return structure is correct: positive integers now also hit the `Atoi` fast path (same result as the literal path), which is harmless

PR #496, opened 2026-05-02
fix: rejoin empty error handlers inside branch bodies

### LINKED ISSUE #495: Empty custom error handlers inside branch bodies can round-trip as return
## Problem

Empty custom error handlers inside structured branch bodies should rejoin the next safe statement in that branch. The top-level flow builder already routes these pending handlers when connecting one statement to the next, but branch builders can connect sequential statements directly.

That means a branch-local empty handler can stay pending until the branch terminates. A describe/exec/describe cycle may then show the empty handler as an explicit `return;` block instead of the authored empty block.

## Expected behavior

For authored MDL like:

```mdl
if $ShouldRefresh then
  call microflow Synthetic.RefreshCache() on error without rollback { };
  $ProcessedCount = call java action Synthetic.CountProcessedItems();
else
  return;
end if;
```

the empty handler should rejoin the Java action inside the branch and continue to describe as an empty handler.

## Proposed fix

Run the same pending empty-handler rejoin step when IF and enum-split branch builders connect sequential branch statements. Keep output-dependent handlers on the existing skip-var path so they still avoid unsafe success tails.

hjotha:
Closes #495.

## Summary

- routes pending empty no-output custom error handlers when IF branch bodies connect one statement to the next
- applies the same fix to enum-split branch bodies
- keeps output-dependent custom handlers on the existing skip-var path
- adds a synthetic IF-branch regression test

## Validation

- `go test ./mdl/executor -run 'EmptyNoOutputHandler.*Branch|EmptyNoOutputHandlerRejoins|OutputHandlerInReturningBranch' -v`
- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/mdl/executor/cmd_microflows_builder_actions.go
+++ b/mdl/executor/cmd_microflows_builder_actions.go
@@ -346,7 +346,7 @@ func (fb *flowBuilder) addEnumSplit(s *ast.EnumSplitStmt) model.ID {
 		lastID := model.ID("")
 		pendingCase := ""
 		var prevAnchor *ast.FlowAnchors
-		for _, stmt := range br.body {
+		for j, stmt := range br.body {
 			thisAnchor := stmtOwnAnchor(stmt)
 			actID := fb.addStatement(stmt)
 			if actID == "" {
@@ -376,6 +376,9 @@ func (fb *flowBuilder) addEnumSplit(s *ast.EnumSplitStmt) model.ID {
 				}
 				applyUserAnchors(flow, prevAnchor, thisAnchor)
 				fb.flows = append(fb.flows, flow)
+				if fb.emptyErrorHandlerFrom == lastID {
+					fb.addPendingErrorHandlerFlowForStatement(lastID, actID, stmt, statementsReferenceVar(br.body[j+1:], fb.errorHandlerSkipVar))
+				}
 			}
 			prevAnchor = thisAnchor
 			if fb.nextConnectionPoint != "" {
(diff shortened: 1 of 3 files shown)


ako:
## Review: fix: rejoin empty error handlers inside branch bodies

### What the PR does

Fixes issue #495: when a `CALL MICROFLOW` with an empty no-output custom error handler appears inside an IF branch body (then or else) or an enum-split branch body, the pending error handler flow wasn't wired to the next statement in that branch. The fix adds the same rejoin call that the top-level flow loop already performs.

### No blockers

### Minor: outer guard is redundant, creates inconsistency with the top-level pattern

The three new blocks read:
```go
if fb.emptyErrorHandlerFrom == lastThenID {
    fb.addPendingErrorHandlerFlowForStatement(...)
}
```

But `addPendingErrorHandlerFlowForStatement` → `addPendingErrorHandlerFlowForState` already has this guard internally:
```go
if state.emptyFrom != "" {
    if state.emptyFrom != originID {
        return state  // no-op
    }
```

Calling it when there's no pending handler, or when the handler is from a different origin, is already a safe no-op. The top-level loop in `cmd_microflows_builder_graph.go:98` calls `addPendingErrorHandlerFlowForStatement` unconditionally — no guard. The branch-body version's outer `if` is functionally correct but redundant, and creates a subtle inconsistency that makes readers wonder why the two call sites look different.

Could drop the outer check to mirror the top-level pattern. Not a blocker — the code is correct either way.

### Positive notes
- Clean, focused fix — 3 lines of logic added in each of 3 places, matching the existing top-level pattern
- `br.body[j+1:]` and `s.ThenBody[i+1:]` are safe when `j`/`i` is the last index (empty slice is valid in Go)
- New test directly reproduces the failure: IF-branch with no-output handler followed by a continuation activity, asserts the error path reaches the continuation
- No bundled unrelated changes

PR #498, opened 2026-05-02
fix: trim empty top-level else branches

### LINKED ISSUE #497: describe can emit empty top-level else after shared continuation traversal
## Problem

Top-level IF traversal can emit an empty `else` branch when the false branch points at a continuation that has already been emitted while traversing the true branch.

The false destination is not necessarily the selected merge, so the existing `falseFlow.DestinationID != mergeID` check can decide that an ELSE body exists. But the actual false-branch traversal then produces no statements because the continuation is already visited.

This creates MDL like:

```mdl
if $UseNestedPath then
  ...
else
end if;
```

A later exec/describe pass removes the empty branch, leaving a roundtrip mismatch.

## Expected behavior

Top-level IF traversal should behave like nested IF traversal: only keep `else` if traversing the false branch actually emits statements.

## Proposed fix

Emit the `else` line tentatively, traverse the false branch with an isolated visited map, and remove the line again if no statements were produced.

hjotha:
Closes #497.

## Summary

- trims a top-level `else` if false-branch traversal emits no statements
- keeps nested IF behavior unchanged, but mirrors its safer tentative-else logic at top level
- adds a synthetic traversal regression test for a visited shared continuation

## Validation

- `go test ./mdl/executor -run 'TopLevelIfRemovesEmptyElse|SequentialIfWithoutElse' -v`
- `make build`
- `make lint-go`
- `make test`

### DIFF
--- a/mdl/executor/cmd_microflows_show_helpers.go
+++ b/mdl/executor/cmd_microflows_show_helpers.go
@@ -724,18 +724,20 @@ func traverseFlow(
 				traverseFlowUntilMerge(ctx, trueFlow.DestinationID, mergeID, activityMap, flowsByOrigin, flowsByDest, splitMergeMap, visited, entityNames, microflowNames, lines, indent+1, sourceMap, headerLineCount, annotationsByTarget)
 			}
 
-			// Emit the ELSE branch only if it has statements. When the false
-			// flow jumps straight to the merge (the MDL was `if X then ... end if`
-			// with no else), emitting `else` with no body produces an empty
-			// branch that normalizes away on re-parse.
-			falseHasBody := falseFlow != nil && falseFlow.DestinationID != mergeID
-			if falseHasBody {
+			if falseFlow != nil {
+				elseLineIdx := len(*lines)
 				*lines = append(*lines, indentStr+"else")
 				visitedFalseBranch := make(map[model.ID]bool)
 				for id := range visited {
 					visitedFalseBranch[id] = true
 				}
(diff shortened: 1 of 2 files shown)


ako:
## Review: fix: trim empty top-level else branches

**Verdict: Approve**

### What the PR does

Fixes issue #497: when a false-branch destination was already visited via the true branch, `traverseFlowUntilMerge` returned immediately at the `visited[currentID]` guard (`cmd_microflows_show_helpers.go:618`) without appending anything. The old check `falseFlow.DestinationID != mergeID` only excluded false branches that pointed directly at the merge — it didn't account for branches pointing at a non-merge node that had already been emitted. The result was an `else` header with an empty body.

The fix switches to a tentative-emit-then-retract pattern: append the `else` header, run the traversal, then check if anything was added. If `len(*lines) == elseLineIdx+1` (only the `else` header itself was appended), trim it back.

### No blockers

### Correctness notes

- `len(*lines) == elseLineIdx+1` is the right sentinel — it fires only when nothing was appended after the `else` header. If the traversal emits any line (including an inner `if ... end if;` with no statements), the outer `else` is correctly kept.
- The previously-working case (`falseFlow.DestinationID == mergeID`) still works identically: `traverseFlowUntilMerge` stops at the merge boundary without appending, then the else is retroactively removed.
- `visitedFalseBranch` is a copy of `visited` at the time of the false branch traversal, so the shared-node scenario the test covers is correctly detected.
- The PR summary says "at top level" but the change is in `traverseFlow` which handles all nesting depths — the description is slightly misleading but the logic is correct at all levels.

### Positive notes
- Tentative-emit-then-retract is cleaner than trying to predict traversal output ahead of time — it handles all "empty false branch" causes with one check
- Test models the exact scenario: outer split → true → inner split → true → shared continuation; outer false branch jumps directly to the same shared node already visited via true

PR #502, opened 2026-05-04
fix: preserve cross-module associations on CREATE object actions

hjotha:
Part of #352.

## Summary

- A `create TargetMod.Entity (OwnerMod.Assoc = \$Ref)` statement round-tripped as an attribute assignment when the association lived in a different module than the create target.
- **Describer side**: `formatAction` on `CreateObjectAction` stripped every association member down to its bare name, losing the cross-module qualifier.
- **Builder side**: `resolveMemberChange` queried the create target's module for the bare name, didn't find the association, and fell through to `AttributeQualifiedName` — triggering Studio Pro **CE1613** ("selected attribute no longer exists") on re-open.

## Fix

1. Describer keeps the `Module.` prefix on any association whose owning module differs from the create target's module.
2. Builder uses the authored module (when the member name is qualified) to look up the association, matching how the describer expresses it.

## Test plan

- [x] New unit test `TestResolveMemberChange_CrossModuleAssociationKeepsAssociationSlot` verifies the association lands on `AssociationQualifiedName` (not the attribute slot) with synthetic module names.
- [x] `go build -tags integration ./...` builds clean.
- [x] `go test ./mdl/executor/ -count=1` passes.
- [ ] Validated in Mendix Studio Pro with \`mx check\` on a local project where a \`CREATE\` assigns a cross-module association — CE1613 no longer reported.

### DIFF
--- a/mdl/executor/cmd_microflows_builder_actions.go
+++ b/mdl/executor/cmd_microflows_builder_actions.go
@@ -1187,20 +1187,29 @@ func (fb *flowBuilder) resolveMemberChange(mc *microflows.MemberChange, memberNa
 	}
 	moduleName := parts[0]
 
-	// If memberName is already qualified (e.g., "MfTest.Order_Customer"),
-	// extract the bare name for association lookup.
+	// If memberName is already qualified (e.g., "Module.Assoc"), the qualifier
+	// is the module that OWNS the association, not the create/change target's
+	// module. Associations can live in any module (see cross-association
+	// lookups below), so prefer the authored module when present.
 	bareName := memberName
 	qualifiedName := memberName
+	lookupModule := moduleName
 	if dot := strings.Index(memberName, "."); dot >= 0 {
+		lookupModule = memberName[:dot]
 		bareName = memberName[dot+1:]
 		// qualifiedName is already set to the full memberName
 	} else {
 		qualifiedName = moduleName + "." + memberName
 	}
 
-	// Query domain model to check if this member is an association
+	// Query the authored (or target) module's domain model first. When the
+	// association actually lives in a different module — the common case for
+	// cross-module associations like `OtherModule.Assoc_Name` on a
+	// `TargetModule.Entity` entity — keep the qualified name so the writer
+	// serialises `Association` correctly instead of falling back to
+	// `Attribute` and triggering Studio Pro CE1613 on re-open.
 	if fb.backend != nil {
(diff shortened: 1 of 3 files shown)


ako:
## Review: fix: preserve cross-module associations on CREATE object actions

### What the PR does

Fixes a describe → exec → describe roundtrip breakage for cross-module associations on `CREATE`/`CHANGE` actions (part of #352):

**Describer side** (`formatAction`): Previously all association names were unconditionally stripped to bare names (`OwnerMod.Assoc → Assoc`). The fix compares the association's module against the create target's module and keeps the prefix only when they differ.

**Builder side** (`resolveMemberChange`): Previously the domain model lookup always used the create target's module (`moduleName`), so a qualified name like `OwnerMod.Assoc` looked in the wrong module and fell through to `AttributeQualifiedName`, triggering Studio Pro CE1613. The fix extracts `lookupModule` from the qualifier when the member name is dotted.

### No blockers

### Minor: formatter change is untested

`cmd_microflows_create_cross_module_assoc_test.go` covers the builder path correctly. The formatter change in `formatAction` — which decides whether to emit `Module.Assoc` or bare `Assoc` — has no new test. The existing `TestFormatAction_CreateObject_WithAssociationMember` only covers the same-module case (`"MyModule.Order_Customer"` on `"MyModule.Order"`), which still strips correctly after the fix.

A test for the cross-module path would be:
```go
action := &microflows.CreateObjectAction{
    EntityQualifiedName: "TargetMod.Entity",
    OutputVariable:      "NewObj",
    InitialMembers: []*microflows.MemberChange{
        {AssociationQualifiedName: "OwnerMod.Entity_Other", Value: "$Ref"},
    },
}
want := "$NewObj = create TargetMod.Entity (OwnerMod.Entity_Other = $Ref);"
```

Not a blocker — the builder test catches the end-to-end regression, and the formatter logic is straightforward — but worth adding for roundtrip completeness (Recurring Finding #4).

### Positive notes
- `strings.SplitN(..., 2)` is the right choice over `strings.Split` — Mendix qualified names are always `Module.Name`, and N=2 avoids surprises with any hypothetical extra dots
- The `lookupModule` variable makes the intent clear: prefer the authored module qualifier over the entity's module for association lookups
- Same-module associations are unaffected — the existing test `TestFormatAction_CreateObject_WithAssociationMember` confirms the prefix is still stripped when `assocModule == entityModule`

PR #503, opened 2026-05-04
fix: rejoin empty error handler at IF split when next IF reads skipVar

hjotha:
## Summary

- When a call action with an output variable had \`on error without rollback { }\` and the following statement was an IF whose condition read that output variable, the builder queued the handler as pending and never rejoined it.
- If both IF branches terminated (both RETURN), the microflow-end terminator then created a **phantom EndEvent** for the un-rejoined handler. Re-describe then surfaced an extra \`return;\` inside the empty handler body, producing a semantic difference between the original and re-executed MPR.
- The authored Mendix pattern wires the error edge into the IF's split: the IF's own \`empty\` / status-code guard in the condition is what handles the error case, and the ELSE branch already contains the fallback path.

## Fix

Narrow change in \`addPendingErrorHandlerFlowForState\`: when \`state.skipVar\` is set and the next statement references it **and** is an \`*ast.IfStmt\`, rejoin at the IF split instead of deferring the handler. Non-IF statements keep the existing deferral semantics.

## Test plan

- [x] \`go test ./mdl/executor/ -count=1\` passes
- [x] \`go build -tags integration ./...\` builds clean
- [x] New unit test \`TestEmptyErrorHandlerFollowedByBothReturnIfRejoinsAtIfSplit\` asserts the error edge from the call reaches the IF split and the microflow ends with exactly 2 EndEvents (no phantom).
- [ ] Validated in Mendix Studio Pro against a local project — \`mx check\` clean on the microflow that originally surfaced this.

### DIFF
--- a/mdl/executor/cmd_microflows_builder_flows.go
+++ b/mdl/executor/cmd_microflows_builder_flows.go
@@ -184,6 +184,18 @@ func (fb *flowBuilder) addPendingErrorHandlerFlowForState(state pendingErrorHand
 	}
 	if state.skipVar != "" {
 		if statementReferencesVar(stmt, state.skipVar) {
+			// An IF whose condition reads the skipVar itself handles the
+			// empty/error case in its own branches (e.g. `if $Response !=
+			// empty and $Response/StatusCode < 400 then … else <fallback>`).
+			// The original Mendix flow in these cases routes the error-
+			// handler edge directly into the IF's split, reusing the
+			// fallback path. Re-using the split as the rejoin point avoids
+			// the phantom EndEvent that terminatePendingErrorHandlersAtEnd
+			// would otherwise create when both IF branches terminate.
+			if _, ok := stmt.(*ast.IfStmt); ok {
+				fb.addErrorHandlerRejoinFlowForState(state, originID, destinationID)
+				return pendingErrorHandlerState{}
+			}
 			if !fb.hasDeclaredReturnValue() {
 				if derivedVar := outputDerivedVariable(stmt, state.skipVar); derivedVar != "" {
 					state.skipVar = derivedVar
--- a/mdl/executor/cmd_microflows_empty_handler_if_rejoin_test.go
+++ b/mdl/executor/cmd_microflows_empty_handler_if_rejoin_test.go
@@ -1,0 +1,106 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package executor
+
+import (
+	"testing"
+
+	"github.com/mendixlabs/mxcli/mdl/ast"
+	"github.com/mendixlabs/mxcli/model"
+	"github.com/mendixlabs/mxcli/sdk/microflows"
+)
+
+// TestEmptyErrorHandlerFollowedByBothReturnIfRejoinsAtIfSplit reproduces the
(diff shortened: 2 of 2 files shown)


ako:
## Review: fix: rejoin empty error handler at IF split when next IF reads skipVar

**Verdict: Approve**

### What the PR does

Fixes a phantom `EndEvent` produced when a `CALL MICROFLOW … on error without rollback {}` is immediately followed by an `IF` that tests the output variable. The pending error-handler state machine deferred indefinitely (because `statementReferencesVar` kept returning true for every IF branch statement), so `terminatePendingErrorHandlersAtEnd` synthesised an orphaned EndEvent — which then surfaced as a spurious `return;` on re-describe.

The fix inserts a single check before the existing deferral logic: if the statement that references `skipVar` is itself an `*ast.IfStmt`, rejoin the error-handler edge immediately at the IF split. The IF's own condition already guards the error scenario in its fallback branch, matching Studio Pro's authored flow graph.

### No blockers

### Correctness notes

The new branch fires only inside the `state.skipVar != ""` && `statementReferencesVar` path — a narrow condition. Non-IF statements that reference `skipVar` keep the existing deferral semantics unchanged.

The `state.emptyFrom` path is processed above this block and unaffected. `addErrorHandlerRejoinFlowForState` is the same function used by the existing non-`skipVar` rejoin path — correct call site.

One inherent limitation: the fix is a heuristic. It assumes that any IF testing `skipVar` handles the error case in one of its branches (the Mendix idiom for `if $Response != empty and … else <fallback>`). If an IF references `skipVar` incidentally without actually guarding the error path, the error-handler edge would still be wired to the IF split — but this matches the Studio Pro graph for this common pattern, and the alternative (phantom EndEvent) is strictly worse.

### Test quality

The test verifies both invariants from the fix:
- Exactly 2 `EndEvent` objects (not 3 — no phantom from `terminatePendingErrorHandlersAtEnd`)
- The error-handler edge from the call activity reaches the IF split

The `depth > 5` cap in the walk is fine for this graph shape. Clean PR, no bundled unrelated changes.

hjotha:
Closing this PR — the fix introduces a **worse regression** than the original bug.

## What happened

After merging this fix locally and running the roundtrip audit, the microflow that originally surfaced the phantom-EndEvent bug went from \`mismatch\` (cosmetic re-describe diff) to **\`mpr_check_failed\` with CE0108** ("Variable 'X' is defined but not in scope at this location").

## Root cause

The authored Mendix pattern routes the error-handler flow to a **merge in the ELSE branch** of the downstream IF — not to the IF's split. That merge sits after the point where the output variable has already been read, so variable scope is preserved. Rejoining at the IF split (as this PR does) skips past the variable declaration, causing Studio Pro to reject subsequent activities that reference it.

The correct fix requires understanding where the IF's branches would otherwise converge and wiring the error flow to that post-IF merge — which the builder does not currently track. That's a larger change than this PR attempted.

## Next steps

- Reverting the fix keeps the bug at its original cosmetic-only severity (extra \`return;\` on re-describe, no \`mx check\` failure).
- A proper fix needs to track the IF's post-convergence merge point and rejoin the error flow there. Deferring to a future PR with a more complete builder model.

Closing without merging.

PR #504, opened 2026-05-04
fix: rejoin empty custom error handler at IF ELSE first activity

hjotha:
## Summary

- When a call activity carries an empty custom error handler (`on error without rollback { }`) and is followed by an IF that tests the call's output variable with both branches terminating, Studio Pro authors the error edge into the fallback path — into a merge shared with the split's `false` branch that feeds the first ELSE activity. The describer/builder now emits that topology instead of leaving the handler dangling.
- Previous attempt (#503, closed) wired the error edge straight to the IF split and triggered CE0108 "variable not in scope" on activities that referenced the call's output variable. This PR places the rejoin at the first ELSE activity via `addEmptyErrorHandlerRejoinFlowFrom(splitID, errorOrigin, firstElseID)`, preserving scope through the split→merge edge.
- Falls back to the pre-existing phantom-EndEvent terminator when the ELSE body is empty, starts with a compound statement, or the follow-up IF does not both-return. No behavioural change outside the narrow capture condition.

## Validation

- Unit test `TestEmptyErrorHandlerRejoinsAtIfElseFirstActivity` verifies: 2 EndEvents (no phantom), error flow from call reaches a merge, merge receives the split's false branch AND flows to the first ELSE activity.
- `go test ./mdl/executor/` passes; `go test -tags integration ./...` passes.
- Targeted audit against the authored reproduction confirms `mx check` errors = 0 and `mpr check` errors = 0; residual roundtrip diff is cosmetic (10 lines, unrelated to this fix).

## Test plan

- [x] Unit test `TestEmptyErrorHandlerRejoinsAtIfElseFirstActivity`
- [x] Full executor test suite
- [x] Integration test suite (`-tags integration`)
- [x] Targeted roundtrip audit on the authored reproduction — MPR validates (0 errors)

### DIFF
--- a/mdl/executor/cmd_microflows_builder_control.go
+++ b/mdl/executor/cmd_microflows_builder_control.go
@@ -41,6 +41,36 @@ func (fb *flowBuilder) addIfStatement(s *ast.IfStmt) model.ID {
 	// a branch's RETURN affecting the parent flow state prematurely
 	savedEndsWithReturn := fb.endsWithReturn
 
+	// Capture an incoming empty custom error handler whose source is the
+	// previous call activity (set by finishCustomErrorHandler via
+	// registerEmptyCustomErrorHandlerWithSkip) when the call has a
+	// non-empty output variable (skipVar). The Studio Pro authored pattern
+	// wires the error edge into the IF's ELSE branch — at the first
+	// activity in the ELSE body — so the fallback path handles both "no
+	// value" and "error" cases. Left un-rejoined,
+	// terminatePendingErrorHandlersAtEnd would synthesise a phantom
+	// EndEvent; naively rejoining at the split would bypass the output
+	// variable declaration and trigger CE0108 "variable not in scope" on
+	// subsequent activities.
+	//
+	// We consume the pending state here so the rejoin is emitted in the
+	// `lastElseID == ""` block below (when the first ELSE activity is
+	// created); the corresponding builder fields are cleared so downstream
+	// logic does not see the handler twice.
+	pendingElseErrorOrigin := model.ID("")
+	if fb.errorHandlerSource != "" && fb.errorHandlerTailFrom == fb.errorHandlerSource &&
+		fb.errorHandlerSkipVar != "" && fb.errorHandlerTailIsSource &&
+		hasElseBody && len(s.ElseBody) > 0 &&
+		bothReturn && // only when both branches terminate — a continuation in THEN would need a different rejoin target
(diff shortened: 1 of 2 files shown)


ako:
## Review: fix: rejoin empty custom error handler at IF ELSE first activity

### What the PR does

Follow-on to #503. That PR wired the error edge directly to the IF split, which caused CE0108 "variable not in scope" — because the error path bypasses the call activity that declares `$R`, and downstream activities reference `$R`. This PR routes the error edge to the first ELSE activity via a shared merge, matching the Studio Pro-authored topology:

```
Call → (normal) → IF Split → true: ThenBody → EndEvent
                           → false: Merge → firstElseActivity → ... → EndEvent
Call → (error) ────────────→ same Merge
```

The ELSE body is the fallback/error path that handles both "no value" and "error" cases without needing `$R`, so scope is preserved.

### No blockers

### Minor: bypasses the state abstraction rather than using `setActivePendingErrorHandler`

The capture block directly clears four individual fields:
```go
fb.errorHandlerSource = ""
fb.errorHandlerTailFrom = ""
fb.errorHandlerSkipVar = ""
fb.errorHandlerTailIsSource = false
```

This is functionally equivalent to `fb.setActivePendingErrorHandler(pendingErrorHandlerState{})` in this case — `registerEmptyCustomErrorHandlerWithSkip` for the `skipVar != ""` path sets exactly these four fields and leaves `emptyErrorHandlerFrom` untouched, so clearing just these four is correct right now. But direct field clearing is slightly fragile: if `pendingErrorHandlerState` gains a new field in the future, this block would miss it. Using `setActivePendingErrorHandler(pendingErrorHandlerState{})` or at minimum a comment saying "emptyErrorHandlerFrom is not set by the skipVar path so it does not need clearing here" would be more robust.

### Correctness notes

- The condition is appropriately strict: `tailFrom == source && tailIsSource` is the exact shape set by `registerEmptyCustomErrorHandlerWithSkip` for the `skipVar != ""` case — no other code path produces this combination
- `bothReturn` guard is important: without it, a continuation in THEN would create a dangling handler in the happy path
- `firstElseIsLeafActivity` correctly excludes compound-statement heads (IF, LOOP, WHILE, ENUM SPLIT) where rejoin plumbing would be more complex
- Fallback to phantom-EndEvent for all unmatched cases is explicit and preserves existing behaviour

### Test

The test correctly verifies the three-edge topology: error flow → merge, split-false → merge, merge → firstElseActivity. The "second log in visitation order = else body log" heuristic is fine for this synthetic test shape. Clean PR, no bundled unrelated changes.

Issue #506, opened 2026-05-04
Microflow roundtrip: REST call with retry-loop error handler loses main-path variable scope (CE0108)


hjotha:
## Summary

When a `rest call` (or other call activity) has a custom error handler of the shape `on error without rollback { ... ; if <retry> then ... else raise end if }` and is followed by activities that reference the call's output variable, the roundtrip produces an MPR that fails `mx check` with `CE0108 "Variable X is defined but not in scope"`. The authored (Studio Pro) topology uses a loop-back edge from the error handler's non-terminating branch back to a merge placed before the source activity to implement retry semantics; MDL does not have syntax to express this, so the builder emits a linear topology that changes variable scope.

## Reproduction

```bash
# base MPR: any project containing the microflow below
mxcli exec -p app.mpr retry_microflow.mdl
mx check app.mpr   # [error] [CE0108] "Variable 'X' is defined but not in scope at this location."
```

Minimal reproducing MDL:

```mdl
create or modify microflow M.Retry ($In: String) returns String begin
  declare $RetryCount Integer = 0;
  $R = call microflow M.Helper(...) on error without rollback {
    log error node @M.LogNode 'retry';
    if $RetryCount<1 then
      set $RetryCount = $RetryCount+1;
    else
      raise error;
    end if;
  };
  retrieve $Items from $R/M.Items_R;   -- CE0108: $R out of scope
  return $In;
end;
/
```

## Root cause

The authored BSON has a merge node placed **before** the source call, with the retry branch of the error handler flowing back to it:

```
prev → MERGE ────┐
                 ↓
              sourceCall
                 ↓
              (happy-path continues)
                 ↑
sourceCall ─err→ Log → Split ─true→ Log → Java → ChangeVar ─┘ (loops back to MERGE)
                          └─false→ ErrorEvent
```

The builder currently reads MDL and produces a linear topology that places the merge **between the two downstream activities**, moving the call's output variable out of scope for subsequent activities.

## Why MDL can't express this today

MDL assumes the error handler is a forward flow: after the handler runs, execution continues to whatever activity follows the source. There is no syntax for "re-execute activity N" or "jump back to merge point M". Studio Pro authors draw the loop-back as a free-form sequence flow that has no surface syntax in MDL.

## Scope

- Reproducible against `origin/main` with no open PRs applied — not a regression introduced by recent PRs.
- Hits any microflow with a call activity carrying a custom error handler of the retry-loop shape described above.

## Possible approaches

[...]

PR #507, opened 2026-05-04
fix: loop-back retry error handler to merge before source activity (#506)

hjotha:
## Summary

- When a call activity's custom error handler ends with an IF where exactly one branch terminates (via `raise error` or `return`) and the other performs retry actions, the builder now reproduces the Studio Pro authored topology: a merge is inserted just before the source activity, and the non-terminating branch's tail loops back to that merge. Previously the builder emitted a linear topology that placed the merge after the source, moving the call's output variable out of scope for downstream activities and triggering `CE0108 "Variable X is defined but not in scope"` on `mx check`.
- `isRetryLoopErrorHandler` detects the pattern (trailing IF with exactly-one-terminating branch) and routes through a new `buildRetryLoopErrorHandler` helper. Non-retry error handlers still use the existing forward-merge path.
- New `flowBuilder.incomingRedirect` field lets the outer loop in `buildFlowGraph` redirect the next inbound flow to the pre-source merge instead of the source activity itself.

## Reproduction & validation

Minimal MDL:
```mdl
create or modify microflow M.Retry ($In: String) returns String begin
  declare $RetryCount Integer = 0;
  $R = call microflow M.Helper(...) on error without rollback {
    log error node @M.LogNode 'retry';
    if $RetryCount<1 then
      set $RetryCount = $RetryCount+1;
    else
      raise error;
    end if;
  };
  return $In;
end;
/
```

Before this PR: `mxcli exec` followed by `mx check` reports `CE0108`. After this PR: `mx check` reports 0 errors. Additional downstream `mx check` diagnostics that surface after the scope fix are pre-existing describer/builder issues in retrieve-chain-over-mapping and are independent of this change.

## Test plan

- [x] Unit test `TestRetryLoopErrorHandlerLoopsBackToSource` — verifies the retry shape produces `prev->merge->source`, `tail->merge`, and no direct `prev->source` edge.
- [x] Unit test `TestNonRetryErrorHandlerUsesForwardMerge` — guards the opposite path (non-retry handlers must not introduce a merge).
- [x] `go test ./...` passes.
- [x] Reproduction MDL yields `mx check` errors = 0; before this PR it reported `CE0108`.

### DIFF
--- a/mdl/executor/cmd_microflows_builder.go
+++ b/mdl/executor/cmd_microflows_builder.go
@@ -32,6 +32,13 @@ type flowBuilder struct {
 	errors              []string          // Validation errors collected during build
 	measurer            *layoutMeasurer   // For measuring statement dimensions
 	nextConnectionPoint model.ID          // For compound statements: the exit point differs from entry point
+	// incomingRedirect, when set, instructs the next enclosing flow emission to
+	// terminate at this merge/activity ID instead of the most recently emitted
+	// activity. Used by retry-loop error handlers (where a merge is inserted
+	// before the source activity so the normal inbound flow enters the merge
+	// and the error-handler tail loops back to the same merge). Cleared after
+	// the outer loop consumes it for the next flow.
+	incomingRedirect model.ID
 	nextFlowCase        string            // If set, next connecting flow uses this case value (for merge-less splits)
 	// nextFlowAnchor carries the branch-specific FlowAnchors that should be
 	// applied to the flow created by the NEXT iteration of buildFlowGraph.
--- a/mdl/executor/cmd_microflows_builder_flows.go
+++ b/mdl/executor/cmd_microflows_builder_flows.go
@@ -51,11 +51,113 @@ func (fb *flowBuilder) finishCustomErrorHandler(activityID model.ID, activityX i
 		return
 	}
 	if len(eh.Body) > 0 {
+		// Retry-loop pattern: error body ends with an IF whose non-terminating
+		// branch should loop back to the source activity (Studio Pro authors
+		// this as a merge placed before the source, with the handler tail
+		// returning to that merge). MDL cannot express the loop-back directly,
+		// so we detect the shape and wire the topology ourselves.
+		if isRetryLoopErrorHandler(eh.Body) {
+			fb.buildRetryLoopErrorHandler(activityID, activityX, eh.Body)
+			return
+		}
 		mergeID := fb.addErrorHandlerFlow(activityID, activityX, eh.Body)
 		fb.handleErrorHandlerMergeWithSkip(mergeID, activityID, outputVar)
 		return
 	}
 	fb.registerEmptyCustomErrorHandlerWithSkip(activityID, eh, outputVar)
+}
+
+// isRetryLoopErrorHandler reports whether the error-handler body looks like
+// a retry loop: the last statement is an IF whose else branch terminates
+// (via RAISE ERROR or RETURN) and whose then branch continues. That shape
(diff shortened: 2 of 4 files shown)


ako:
## Review: fix: loop-back retry error handler to merge before source activity

**Verdict: Approve with moderate note**

### What the PR does

Fixes CE0108 for the retry-loop error handler pattern: when a call activity's custom error handler ends with an IF where exactly one branch terminates (`raise`/`return`) and the other performs retry actions, Studio Pro places a merge **before** the source call, wiring normal inbound flow through `prev → merge → source` and the handler's non-terminating tail back to `merge`. The previous builder placed the merge after the source, moving the call's output variable out of scope for downstream activities.

The fix introduces:
- `isRetryLoopErrorHandler` + `bodyTerminates` — detect the pattern
- `buildRetryLoopErrorHandler` — builds the pre-source merge and loop-back flows, sets `fb.incomingRedirect`
- `incomingRedirect` field — the outer `buildFlowGraph` loop redirects the next inbound flow to the merge instead of the source

### Moderate: `bodyTerminates` is inconsistent with the existing `isTerminalStmt`

`isTerminalStmt` (line 771) handles four terminal types: `ReturnStmt`, `RaiseErrorStmt`, `BreakStmt`, `ContinueStmt`. The new `bodyTerminates` only handles the first two. If an error handler's "error branch" ends with `BreakStmt` or `ContinueStmt`, `bodyTerminates` returns `false`, `isRetryLoopErrorHandler` returns `false`, and the fix doesn't fire — producing a CE0108 regression for that shape.

`BreakStmt`/`ContinueStmt` in an error handler body are unusual (error handlers don't run inside a loop scope directly), so this is low-risk in practice. But it is a divergence from the existing convention. The simplest fix is to make `bodyTerminates` delegate to `isTerminalStmt`:
```go
func bodyTerminates(body []ast.MicroflowStatement) bool {
    if len(body) == 0 {
        return false
    }
    return isTerminalStmt(body[len(body)-1])
}
```
That would collapse two functions into one and eliminate the inconsistency.

### Minor: `incomingRedirect` is a temporal side-channel

The field is set inside `addStatement` (deep in `finishCustomErrorHandler`) and consumed in the same `buildFlowGraph` iteration. This coupling works correctly today but is implicit — if any code between `addStatement` returning and the `incomingRedirect` consumption ever calls another statement-building path, the redirect could be consumed by the wrong flow. The inline comment documents it well; a `// consumed in the same iteration` note at the field definition would reinforce the contract.

### Correctness notes

- `buildRetryLoopErrorHandler` calls `addErrorHandlerFlow` first (emitting `source → (error) → first-handler-activity`), then inserts the merge. The merge is strictly on the INPUT side of the source — it does not affect the error flow edge. ✓
- The loop-back flow is correctly emitted as a plain `SequenceFlow` (not `IsErrorHandler`), matching Studio Pro's topology and avoiding CE0136/CE0019 on retrieve chains. ✓
- `fb.baseY` field exists (line 24 of `cmd_microflows_builder.go`). ✓
- `isRetryLoopErrorHandler` gating on `thenTerminates != elseTerminates` correctly excludes both-terminate and neither-terminate IFs. ✓

### Tests

Both tests are well-structured. `TestRetryLoopErrorHandlerLoopsBackToSource` verifies all three topology edges (`prev→merge`, `merge→call`, `tail→merge`) plus the negative guard (no direct `non-merge→call` normal flow). `TestNonRetryErrorHandlerUsesForwardMerge` confirms no merge is introduced for a non-retry body. Clean PR, no bundled unrelated changes.

PR #508, opened 2026-05-04
fix: resolve JSON structure qualified-name lookup through folder hierarchy

hjotha:
## Summary

- `GetJsonStructureByQualifiedName` matched the stored `ContainerID` directly against the module ID, so JSON structures nested in module subfolders (e.g. `Module/Subfolder/MyJson`) were reported as "not found".
- That failure silently degraded REST-call roundtrip fidelity: `addRestCallAction` uses the JSON structure's root element type to decide whether the import mapping returns a single Object (`SingleObject=true`) or a List (`SingleObject=false`). A failed lookup defaulted to `SingleObject=false`, so any REST call whose import mapping referenced a folder-nested JSON structure produced List-typed output when the authored mapping was Object-typed.
- Downstream activities that used the call's output variable then failed `mx check` with cascading diagnostics — `CE0019` (variable as start point must be Object), `CE0097` (list operation needs a List), `CE0136` (retrieve missing Entity), `CE0243` (mapping type mismatch) — all rooted in the single misclassified `SingleObject` flag.

## Fix

Use `buildContainerModuleNameMap` — the same helper `GetImportMappingByQualifiedName` already uses — to resolve container IDs through the folder hierarchy up to the owning module, then match on module name.

```go
// before
moduleID := ""
for _, m := range modules {
    if m.Name == moduleName {
        moduleID = string(m.ID); break
    }
}
for _, js := range all {
    if js.Name == name && (moduleID == "" || string(js.ContainerID) == moduleID) {
        return js, nil
    }
}

// after
moduleMap, _ := r.buildContainerModuleNameMap()
for _, js := range all {
    if js.Name == name && moduleMap[js.ContainerID] == moduleName {
        return js, nil
    }
}
```

No schema change; lookup behaviour only. Mirrors existing `GetImportMappingByQualifiedName` / `GetExportMappingByQualifiedName` patterns.

## Test plan

- [x] `TestGetJsonStructureByQualifiedName_ResolvesThroughFolders` — synthetic MPR with a JSON structure placed inside a module subfolder; lookup by module name must find it.
- [x] Same test exercises the negative case (cross-module lookup must still fail).
- [x] `go test ./...` passes.
- [x] Manual reproduction: exec MDL containing `returns mapping M.IMM_X as M.Entity` where `M.JSON_X` is nested under `M/Folder/` now produces `Range.SingleObject=True` + `VariableType=ObjectType` matching the authored BSON.

### DIFF
--- a/sdk/mpr/jsonstructure_folder_lookup_test.go
+++ b/sdk/mpr/jsonstructure_folder_lookup_test.go
@@ -1,0 +1,124 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package mpr
+
+import (
+	"database/sql"
+	"path/filepath"
+	"testing"
+
+	"github.com/mendixlabs/mxcli/model"
+	"go.mongodb.org/mongo-driver/bson"
+	_ "modernc.org/sqlite"
+)
+
+// JSON structures live inside a module but may be nested in subfolders.
+// GetJsonStructureByQualifiedName must resolve container IDs through the
+// folder hierarchy up to the owning module; otherwise addRestCallAction
+// silently defaults SingleObject=false and produces invalid REST-call
+// roundtrips on projects that organise their JSON structures in
+// folders.
+func TestGetJsonStructureByQualifiedName_ResolvesThroughFolders(t *testing.T) {
+	dbPath := filepath.Join(t.TempDir(), "test.mpr")
+	db, err := sql.Open("sqlite", dbPath)
+	if err != nil {
+		t.Fatalf("open: %v", err)
+	}
+	t.Cleanup(func() { _ = db.Close() })
+
+	if _, err := db.Exec(`
+		CREATE TABLE Unit (
+			UnitID BLOB PRIMARY KEY NOT NULL,
+			ContainerID BLOB,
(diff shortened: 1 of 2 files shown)


ako:
## Review: fix/resolve-json-structure-qualified-name-lookup

**Result: Approved — no blockers.**

### What the PR does

Fixes  in  to resolve the module name through the full folder hierarchy (using ) instead of doing a direct  comparison, which failed for structures nested inside folders.

### Checklist

- [x] **Mirrors established pattern** —  is already used by  and ; this PR brings JSON structures into line with that pattern
- [x] **Secondary bug fixed** — the old  fallback (triggered when the module name wasn't found) would silently return any structure with a matching name from *any* module; new code correctly requires an exact module-name match via the map
- [x] **Test coverage** —  creates a synthetic MPR with a Module → Folder → JsonStructure hierarchy and verifies both the positive (folder-nested lookup succeeds) and negative (wrong-module lookup returns error) cases
- [x] **No executor changes needed** — pure SDK layer fix; callers are unaffected

### Notes

The structure of the fix is clean and the test is appropriately targeted. No concerns.

Issue #514, opened 2026-05-04
Remove generated ANTLR parser files from git


ako:
## Problem

Every MDL feature that adds syntax triggers a full regeneration of `mdl/grammar/parser/mdl_parser.go` (119,737 lines). When two branches both run `make grammar`, this file produces an unresolvable conflict. The file is deterministically derived from the `.g4` source — there is no information in it that isn't already in the grammar.

## Proposed solution

Add `mdl/grammar/parser/*.go` (and `.interp`/`.tokens` files) to `.gitignore` and stop tracking them. Regenerate with `make grammar` locally and in CI.

**Prerequisites:** All grammar contributors already have ANTLR4 installed — this has been verified in practice.

## Implementation steps

1. Add to `.gitignore`:
   ```
   # Generated ANTLR4 parser (regenerate with: make grammar)
   mdl/grammar/parser/*.go
   mdl/grammar/parser/*.interp
   mdl/grammar/parser/*.tokens
   ```
2. Run `git rm --cached` on the tracked generated files
3. Add ANTLR4 install + `make grammar` step to CI workflows (`push-test.yml`, `release.yml`) before `make build`/`make test`
4. Add a `make grammar-check` target that regenerates and diffs, to catch un-regenerated grammar changes in CI

## Expected outcome

Grammar-related merge conflicts are eliminated. Two branches adding syntax to different parts of the grammar can no longer conflict on the generated file.

See proposal: `docs/11-proposals/PROPOSAL_agentic_architecture_improvements.md` (Change 1)

PR #519, opened 2026-05-05
fix: roundtrip REST mapping cardinality via `as list of` syntax

hjotha:
## Summary

Restore correct cardinality (single object vs list) for REST-call import mappings, especially those backed by XML schemas or message definitions. The fix has three layers: (1) make the JSON-structure lookup work through folder hierarchies, (2) fix the parser so `MinOccurs`/`MaxOccurs` survive the BSON read, and (3) propagate the call-site cardinality through MDL using new `as list of Module.Entity` syntax instead of guessing from the import mapping shape.

### Why the call site needs explicit syntax

Studio Pro stores the cardinality on the microflow's `ImportMappingCall`, not on the import mapping itself:

| Microflow | `Range.SingleObject` | `ForceSingleOccurrence` | Authored result |
|-----------|----------------------|-------------------------|-----------------|
| `PrivateCloudData.REST_GetEnvironmentByUUID` | `false` | `true`  | single Object |
| `MendixSSO.RetrieveUserRoles`                | `false` | `false` | list          |

Both bind the same shape of mapping element (`Kind: "Object"`, `MaxOccurs: -1`). The user's choice is at the call site. Earlier iterations of this PR tried to derive cardinality from `MaxOccurs` alone (`-1` ⇒ list); that classified PCD as a list and produced `CE0117 "Error(s) in expression."` at the End event, where the microflow returns a single `PrivateCloudData.Environment`.

### Fixes

1. **Folder-nested JsonStructure lookup** — `GetJsonStructureByQualifiedName` matched the stored `ContainerID` directly against the module ID, so structures inside module subfolders (`Module/Sub/Json`) were reported as not found. Walk the container parent chain via `buildContainerModuleNameMap`, the same helper `GetImportMappingByQualifiedName` already uses.
2. **Fall back to import mapping root** — when `JsonStructure` is empty (XML schema / message definition), use the mapping's own root element kind as the cardinality hint when MDL omits the explicit `as list of` form.
3. **Parse `MinOccurs` / `MaxOccurs`** — the MPR reader silently dropped these on `ImportMappings$ObjectMappingElement` / `ImportMappings$ValueMappingElement` because Mendix authors them as `int64` while the existing JsonElement parser only handles `int32`. Add a small `bsonInt` helper that accepts both.
4. **`as list of Module.Entity` syntax** — new MDL form to mark a list result. Pipeline:
   - **Grammar** (`MDLMicroflow.g4`): adds `RETURNS MAPPING qn AS LIST_OF qn` ahead of the bare `AS qn` form.
   - **AST** (`RestResult.IsList`): records the cardinality.
   - **Visitor**: sets `IsList` when `LIST_OF()` is present in the parse tree.
   - **Describer** (`formatRestCallAction`): emits `as list of Module.Entity` when `rh.SingleObject` is false; `as Module.Entity` otherwise.
   - **Builder** (`addRestCallAction`): trusts `s.Result.IsList` to set both `SingleObject` and `ForceSingleOccurrence` (mirrored), so the writer reproduces the BSON shape Studio Pro emits.

### MDL syntax

```mdl
-- Single object — first item from a list-typed mapping (PCD pattern)
$Item = rest call get $URL
  returns mapping Module.IMM as Module.Item;

-- List result (MendixSSO pattern)
$Items = rest call get $URL
  returns mapping Module.IMM as list of Module.Item;
```

## Reproduction

A REST call binding a single object via a list-typed mapping:

```mdl
$Environment = rest call get $URL
  returns mapping PrivateCloudData.IMM_EnvironmentByUUID as PrivateCloudData.Environment;
return $Environment;
```

Before this PR: `mx check` reports `CE0117 "Error(s) in expression." at End event` because the result was forced to a `ListType` while the microflow returns a single Object. After this PR: 0 errors; the result is correctly typed and the BSON roundtrip preserves both `Range.SingleObject` and `ForceSingleOccurrence`.

## Test plan

- [x] `TestGetJsonStructureByQualifiedName_ResolvesThroughFolders` — folder-nested JSON structure lookup.
- [x] `TestAddRestCallAction_MappingAsEntityProducesSingleObject` — `as Module.Entity` (no `list of`) ⇒ `SingleObject=true`, `ForceSingleOccurrence=true`.
- [x] `TestAddRestCallAction_MappingAsListOfEntityProducesListResult` — `as list of Module.Entity` ⇒ `SingleObject=false`, `ForceSingleOccurrence=false`.
- [x] `TestFormatRestCallAction_MappingSingleObject` — describer emits `as` for `SingleObject=true`.
- [x] `TestFormatRestCallAction_MappingListOf` — describer emits `as list of` for `SingleObject=false`.
- [x] `TestAddImportFromMappingFallsBackToImportMappingRootForListResult` / `…ForSingleObject` — heuristic fallback for `IMPORT FROM MAPPING` (no MDL syntax change there).
- [x] `go test ./...` passes.
- [x] Roundtrip audit on the seven microflows that exercise the touched paths (PCD `REST_GetEnvironmentByUUID`, MendixSSO `RetrieveUserRoles`, plus five others): six match with zero `mx check` errors; PCD now matches where it previously failed `CE0117`. The seventh (`SBOMData.REST_GetSBOMByComponentId`) was a pre-existing whitespace-only cosmetic mismatch unrelated to this change.

### DIFF
--- a/.claude/skills/mendix/write-microflows.md
+++ b/.claude/skills/mendix/write-microflows.md
@@ -882,7 +882,10 @@ rest call delete 'https://api.example.com/items/{1}' with (
 - `returns string` — response body as string variable
 - `returns nothing` / `returns none` — ignore response
 - `returns response` — returns `System.HttpResponse` object
-- `returns mapping Module.ImportMapping as Module.Entity` — import mapping
+- `returns mapping Module.ImportMapping as Module.Entity` — single object result
+- `returns mapping Module.ImportMapping as list of Module.Entity` — list result
+
+**Pick `as` vs `as list of` based on the call site, not the mapping shape.** The same import mapping can yield either a single object or a list — Studio Pro stores the cardinality on the microflow's `ImportMappingCall` (`Range.SingleObject` + `ForceSingleOccurrence`). Use `as Module.Entity` when the response is a single object (the mapping may still be list-typed; Studio Pro binds the first item). Use `as list of Module.Entity` when the response should bind a list. Mismatching the cardinality with the surrounding code produces `mx check` `CE0117` at the End event or `CE0013` / `CE0100` on downstream loop / aggregate / list-operation activities.
 
 **REST CALL supports full error handling** (`on error continue`, `on error rollback`, custom error handlers).
 
--- a/docs/01-project/MDL_QUICK_REFERENCE.md
+++ b/docs/01-project/MDL_QUICK_REFERENCE.md
@@ -231,6 +231,11 @@ authentication basic, session
 | Call Java action | `$Result = call java action Module.Name (Param = $value);` | Java action (microflow only) |
 | Call web service | `$Result = call web service Module.Service operation OperationName;` | Legacy SOAP; quoted refs are fallback for dangling raw IDs |
 | Call web service raw | `$Result = call web service raw 'base64-bson';` | Escape hatch for byte-for-byte legacy SOAP round-trip |
+| REST call (string) | `$Var = rest call get '<url>' returns string;` | Body as string |
+| REST call (response) | `$Var = rest call get '<url>' returns response;` | `System.HttpResponse` object |
+| REST call (mapping single) | `$Var = rest call get '<url>' returns mapping Module.IMM as Module.Entity;` | Single object — Studio Pro emits `ForceSingleOccurrence=true` |
+| REST call (mapping list) | `$Var = rest call get '<url>' returns mapping Module.IMM as list of Module.Entity;` | List result |
+| REST call (none) | `rest call get '<url>' returns nothing;` | Discard response |
 | Show page | `show page Module.PageName ($Param = $value);` | Also accepts `(Param: $value)` |
 | Close page | `close page;` | |
 | Download file | `download file $FileDocument [show in browser];` | Streams a `System.FileDocument` |
--- a/mdl-examples/bug-tests/519-rest-mapping-as-list-of.mdl
+++ b/mdl-examples/bug-tests/519-rest-mapping-as-list-of.mdl
@@ -1,0 +1,90 @@
+-- ============================================================================
+-- Bug #519: REST mapping result cardinality must be authored at the call site
+-- ============================================================================
+--
+-- Symptom (before fix):
+--   `returns mapping Module.IMM as Module.Entity` was the only form available.
+--   The same import mapping can yield either a single object or a list
+--   depending on the call site — Studio Pro stores the cardinality on the
+--   microflow's ImportMappingCall (Range.SingleObject + ForceSingleOccurrence),
+--   not on the import mapping itself. mxcli's builder had to guess from the
+--   import-mapping shape, which was lossy. With message-definition / XML-schema
+--   backed mappings (no JsonStructure), the guess flipped between releases —
+--   one version forced single, another forced list — producing CE0117 at the
+--   End event or CE0013 / CE0100 on downstream loop / aggregate / list-op
+--   activities depending on which way the heuristic went.
+--
+-- After fix:
+--   New MDL syntax `returns mapping Module.IMM as list of Module.Entity` makes
+--   the cardinality explicit at the call site. The describer emits `as list of`
+--   when SingleObject=false and `as` otherwise, so the roundtrip preserves the
+--   exact cardinality Studio Pro authored. The builder mirrors SingleObject
+--   into ForceSingleOccurrence so the writer reproduces the BSON layout
+--   Studio Pro emits.
+--
+-- Validation:
+--   `mxcli check 519-rest-mapping-as-list-of.mdl` parses both forms.
+--   Roundtrip (describe → exec → describe) preserves both cardinality forms
+--   byte-for-byte. `mx check` against the resulting MPR reports 0 errors for
+--   the single-object case (consumed via `return $Item`) and the list case
+--   (consumed via `count($Items)` and `loop`).
+--
+-- Usage:
+--   mxcli exec mdl-examples/bug-tests/519-rest-mapping-as-list-of.mdl -p app.mpr
+--   mxcli -p app.mpr -c "describe microflow BugTest519.MF_GetSinglePet"
+--   mxcli -p app.mpr -c "describe microflow BugTest519.MF_GetPetList"
+-- ============================================================================
+
+create module BugTest519;
+
+create json structure BugTest519.JSON_Pets
+snippet '[{"id": 1, "name": "Fido", "status": "available"}]';
+
+create non-persistent entity BugTest519.PetResponse (
+  PetId : integer,
+  Name : string,
+  Status : string
+);
+/
+
+create import mapping BugTest519.IMM_Pet
+  with json structure BugTest519.JSON_Pets
+{
+  create BugTest519.PetResponse {
+    PetId = id,
+    Name = name,
+    Status = status
+  }
+};
+
+-- Single-object form: pull the first element of a list-typed mapping into a
+-- single result. Mirrors the PrivateCloudData.REST_GetEnvironmentByUUID
+-- pattern (ForceSingleOccurrence=true, Range.SingleObject=false).
+create microflow BugTest519.MF_GetSinglePet (
+  $Url : string
+)
+returns BugTest519.PetResponse
+begin
+  $Pet = rest call get $Url
+    timeout 30
+    returns mapping BugTest519.IMM_Pet as BugTest519.PetResponse
+    on error rollback;
+  return $Pet;
+end;
+/
+
+-- List form: bind a list result. Mirrors the MendixSSO.RetrieveUserRoles
+-- pattern (ForceSingleOccurrence=false, Range.SingleObject=false).
+create microflow BugTest519.MF_GetPetList (
+  $Url : string
+)
+returns Integer
+begin
+  $Pets = rest call get $Url
+    timeout 30
+    returns mapping BugTest519.IMM_Pet as list of BugTest519.PetResponse
+    on error rollback;
+  declare $Count Integer = count($Pets);
+  return $Count;
+end;
+/
--- a/mdl-examples/doctype-tests/06-rest-client-examples.mdl
+++ b/mdl-examples/doctype-tests/06-rest-client-examples.mdl
@@ -1457,6 +1457,33 @@ begin
 end;
 /
 
+-- ============================================================================
+-- Level 15.2: GET single object via `as Module.Entity`
+-- ============================================================================
+--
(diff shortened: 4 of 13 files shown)


ako:
## Review

The core fix is sound: falling back to `ImportMappingElement.Kind` for message-definition- and XML-schema-backed mappings (no `JsonStructure`) correctly resolves the CE0117/CE0019/CE0136 cascade. The Object-fallback path in `addRestCallAction` is correct and addresses the real bug.

---

### Moderate

**M1 — Array fallback in `addImportFromMappingAction` is dead code**

`parseImportMappingElement` only ever sets `Kind` to `"Object"` (`ImportMappings$ObjectMappingElement`) or `"Value"` (`ImportMappings$ValueMappingElement`). There is no `ImportMappings$ArrayMappingElement` branch — unrecognised `$Type` values fall through to `default: return nil`, and nil elements are never appended to `im.Elements`. So `im.Elements[0].Kind == "Array"` can never be `true` from real MPR data.

```go
// sdk/mpr/parser_import_mapping.go
switch typeName {
case "ImportMappings$ObjectMappingElement":   // → Kind: "Object"
    return parseImportObjectMappingElement(raw)
case "ImportMappings$ValueMappingElement":    // → Kind: "Value"
    return parseImportValueMappingElement(raw)
default:
    return nil  // any Array element type falls here and is dropped
}
```

`TestAddRestCallAction_MappingFallsBackToArrayKindWhenJsonStructureMissing` constructs `Kind: "Array"` in a mock — a state the parser cannot produce — so it provides false assurance for a dead branch.

**Fix options (pick one):**
- Verify with a real Studio Pro BSON dump of an XML-schema-backed array-rooted import mapping. If `ImportMappings$ArrayMappingElement` exists as a `$Type`, add a parser case; the Array fallback and test then become meaningful.
- If no such `$Type` exists in real MPR files, remove the Array fallback from `addImportFromMappingAction` and drop the Array-kind test.

**M2 — No test for `addImportFromMappingAction` with the new message-definition fallback**

Two new tests cover the REST call path but none exercise `addImportFromMappingAction` with `JsonStructure = ""` and an Object-rooted element. The `resolved` flag + Object fallback in that function is equally new and untested.

---

### Minor

**m3 — Nil guard inconsistency**

The new fallback guards `im.Elements[0] != nil`, but the immediately following pre-existing entity lookup does not:

```go
if len(im.Elements) > 0 && im.Elements[0].Entity != "" {  // no nil guard
```

Low risk in production (parser never appends nil), but inconsistent with the guard added two lines above.

---

### Verdict

Approve after M1 and M2. M1 in particular could mislead a future contributor into believing array-rooted message-definition mappings work end-to-end when the parser silently drops them.

hjotha:
Thanks @ako — addressed in `c393ddee`:

**M1 — dead Array branch removed.** You're right: `parseImportMappingElement` only ever produces `"Object"` or `"Value"`, so the `Kind == "Array"` check was unreachable from real MPR data. Both fallback sites (`addRestCallAction` and `addImportFromMappingAction`) now key on `MaxOccurs == -1 || MaxOccurs > 1` only — repetition is the actual discriminator for repeating-Object roots in real authored BSON. The `TestAddRestCallAction_MappingFallsBackToArrayKindWhenJsonStructureMissing` test is removed alongside the dead branch (mock-only state).

**M2 — `addImportFromMappingAction` fallback now has tests.** Two new cases in `cmd_microflows_builder_import_mapping_test.go`:
- `TestAddImportFromMappingFallsBackToImportMappingRootForListResult` — Object root with `MaxOccurs: -1` → `List of <Entity>`.
- `TestAddImportFromMappingFallsBackToImportMappingRootForSingleObject` — Object root with `MaxOccurs: 1` → singleton.

**m3 — guard parity.** Added the `im.Elements[0] != nil` guard before the pre-existing entity lookup so it matches the guard above. (Low-risk in practice since the parser never appends nil, but eliminates the inconsistency.)

If we ever discover an `ImportMappings$ArrayMappingElement` (or similar) `$Type` in a real MPR dump, the parser dispatcher in `parser_import_mapping.go` is the single place that needs a new case — the fallback no longer needs to special-case it.

ako:
## Review (updated)

All three items from the first round have been addressed or superseded. Two things left.

---

### Moderate

**M1 — `bsonInt` duplicates `extractInt` in the same package**

`parser.go` already defines:

```go
func extractInt(v any) int {
    if v == nil { return 0 }
    switch val := v.(type) {
    case int32:  return int(val)
    case int64:  return int(val)
    case int:    return val
    case float64: return int(val)
    }
    return 0
}
```

The new `bsonInt` in `parser_import_mapping.go` is identical except it omits the `float64` case and the `nil` guard. Both are in `package mpr`. Replace `bsonInt` with `extractInt` throughout `parser_import_mapping.go` and delete the new helper.

---

### Minor

**m2 — `TestAddRestCallAction_MappingFallsBackToArrayKindWhenJsonStructureMissing` tests an unreachable state**

The test comment says "an Array root on the mapping element must yield a list-typed result handling" but exercises `Kind: "Array"` — a value `parseImportMappingElement` never produces (`ImportMappings$ObjectMappingElement` → "Object", `ImportMappings$ValueMappingElement` → "Value"; no Array branch). The practical list-from-root path is `MaxOccurs > 1 / -1`, which is already well-covered by the Unbounded and Bounded repeat tests. This test doesn't cause a bug but documents a non-existent input shape. Low priority — rename it or add a comment that it represents a hypothetical future parser extension.

**m3 — Nil guard inconsistency (pre-existing, now more visible)** `im.Elements[0] != nil` is checked inside the new fallback, but the immediately following entity-lookup `im.Elements[0].Entity != ""` (unchanged from before this PR) has no nil guard.

---

### What's improved vs round 1

- Folder-hierarchy fix for `GetJsonStructureByQualifiedName` with its own parser-level test ✓  
- `MaxOccurs > 1 / -1` reliably detects list regardless of Kind, covering the real Studio Pro data ✓  
- `addImportFromMappingAction` fallback now tested (Object singleton + unbounded list) ✓  
- `MinOccurs` / `MaxOccurs` are now parsed from BSON ✓  

Approve after the `bsonInt` → `extractInt` substitution.

hjotha:
Thanks @ako — addressed in `dcf5b7c3`.

**M1 — `bsonInt` → `extractInt`**

You're right, `extractInt` in `parser.go` already handles `int32` / `int64` / `int` / `float64` with a nil guard, which is a strict superset of `bsonInt`. Replaced both `MinOccurs` / `MaxOccurs` call sites and removed the helper.

**m2 — unreachable Array-kind test**

Already removed in the previous fixup `c393ddee` (the dead `Kind == "Array"` branch and its test). No leftover Array-kind code or tests.

**m3 — nil guard inconsistency**

Fixed in `c393ddee` as well — `im.Elements[0]` is now nil-guarded before the entity-name lookup.

The PR now also adds new MDL syntax `as list of Module.Entity` (commit `c4febc1d`) so the cardinality round-trips correctly even for cases like PCD's `REST_GetEnvironmentByUUID`, where the underlying mapping is list-typed (`MaxOccurs=-1`) but the call site binds a single Object via `ForceSingleOccurrence=true`. The MaxOccurs heuristic alone couldn't tell those apart.

ako:
## Code Review

The core design (trusting call-site MDL for cardinality instead of inferring from the mapping shape) is correct and well-justified. The `ForceSingleOccurrence` + `SingleObject` mirroring fix is solid, and the folder-nested JSON structure lookup is a clean fix.

### Required before merge

- **`MDL_QUICK_REFERENCE.md`** — `returns mapping` table needs the `as list of` variant added.
- **`.claude/skills/write-microflows.md`** — new syntax not documented; authors writing REST-call MDL won't know the form exists.
- **Missing doctype test** — `mdl-examples/doctype-tests/06-rest-client-examples.mdl` has no example using `returns mapping ... as list of`. Required per checklist for syntax changes.

### Minor

- `369-rest-mapping-result-cardinality.mdl` doesn't exercise the new `as list of` path, and its comment says "There is no new MDL syntax" (now stale). A `519-*.mdl` bug test covering the `as list of` roundtrip would be cleaner.
- In the `addImportFromMappingAction` fallback, `MaxOccurs == 0` (Go zero value / unset) stays `SingleObject=true`. That's the right default but a short comment on the ambiguity would help future readers.

### Note on generated parser files

`make grammar` regenerates `mdl/grammar/parser/` at build time — no need to commit those files. The grammar change in `domains/MDLMicroflow.g4` is sufficient.

hjotha:
Thanks @ako — addressed in `8e0a8884`.

**Required**

- **`docs/01-project/MDL_QUICK_REFERENCE.md`** — REST CALL row added under Microflow statements with both `as Module.Entity` (single object) and `as list of Module.Entity` (list result) variants. Earlier table had no rest-call entry at all; this also fills that gap.
- **`.claude/skills/mendix/write-microflows.md`** — REST CALL response types now lists both forms with a paragraph explaining the call-site cardinality rule (Studio Pro stores `ForceSingleOccurrence` + `Range.SingleObject` on the `ImportMappingCall`, not on the import mapping itself; mismatching the cardinality with the surrounding code triggers CE0117 / CE0013 / CE0100).
- **`mdl-examples/doctype-tests/06-rest-client-examples.mdl`** — Levels 15.2 and 15.3 added. 15.2 covers single-object via `as Module.Entity` (`return $Pet`); 15.3 covers list via `as list of Module.Entity` (`count($Pets)`).

**Minor**

- **`mdl-examples/bug-tests/519-rest-mapping-as-list-of.mdl`** — new bug-test mirroring the two real patterns: `MF_GetSinglePet` (PCD `REST_GetEnvironmentByUUID` shape, `ForceSingleOccurrence=true`) and `MF_GetPetList` (MendixSSO `RetrieveUserRoles` shape, `ForceSingleOccurrence=false`). Leaving #369 in place — it covers the import-from-mapping path which is unchanged here.

The note about generated parser files: agreed, `mdl/grammar/parser/` is gitignored on this branch (added in #514). The grammar change in `domains/MDLMicroflow.g4` is the only source-of-truth edit.

ako:
## Updated Review

All required items from the previous review are addressed:

- **Bug test** — `519-rest-mapping-as-list-of.mdl` added with clear symptom, fix description, and validation steps.
- **Doctype test** — `06-rest-client-examples.mdl` updated with both `as Module.Entity` and `as list of Module.Entity` examples.
- **`write-microflows.md`** — updated with the two forms and a clear explanation of call-site semantics, including the CE0117/CE0013/CE0100 consequences of mismatching cardinality.
- **`MDL_QUICK_REFERENCE.md`** — updated.

The skill doc explanation is particularly good — it explicitly calls out that the same mapping can yield either a single object or a list depending on the call site, which is the non-obvious part of this fix.

Ready to merge.

PR #521, opened 2026-05-05
fix: emit "empty" keyword for typed Java action arguments

hjotha:
## Summary

When a Java action parameter is **typed as a `BasicParameterType`** (any inner type other than entity-type or microflow-type — `String`, `Integer`, `Boolean`, `ListType`, `ParameterizedEntityType`, `DateTime`, `Decimal`, …) and the MDL author binds it to `empty`, Studio Pro authors the BSON `BasicCodeActionParameterValue` with `Argument: "empty"` — the literal MDL keyword. The current builder collapses every empty primitive binding to `Argument: ""` (the unbound marker), regardless of parameter type. For typed parameters that substitution triggers `mx check` `CE0126 "Missing value for parameter X"` because the model treats the parameter as missing rather than explicitly empty.

## Fix

When the Java action definition is resolvable via the backend, classify each parameter:
- `EntityTypeParameterType` → already handled separately (`EntityTypeCodeActionParameterValue`).
- `MicroflowType` → already handled separately (`MicroflowParameterValue` with empty `Microflow`).
- Anything else (BasicParameterType — primitives, lists, parameterized-entity types) → emit `Argument: "empty"` for the MDL `empty` literal.

When `jaDef == nil` (no backend or action not resolvable) the builder keeps the prior `Argument: ""` behaviour to preserve the documented "intentionally unbound" semantics from `PROPOSAL_microflow_empty_java_action_argument.md`. The describer already maps both shapes to the MDL `empty` keyword so round-trip stays symmetric.

## Reproduction

A Java action call that binds a typed parameter to `empty`:

```mdl
call java action M.MyAction(items = empty);          -- ListType
call java action M.MyAction(key = empty);            -- StringType
call java action M.MyAction(microflowArgument = empty); -- ParameterizedEntityType
```

Before this PR: `mx check` reports `CE0126 "Missing value for parameter X"`. After this PR: `mx check` reports 0 errors.

## Test plan

- [x] Existing test `TestBuildJavaAction_EmptyArgumentPreservesEmptyBasicValue` continues to pass — without backend (jaDef == nil) primitives still emit `Argument: ""`.
- [x] New parametrized test `TestBuildJavaAction_EmptyResolvedBasicArgumentEmitsEmptyKeyword` covers ListType and StringType — both must emit `Argument: "empty"` once the parameter type is resolved.
- [x] Existing test `TestBuildJavaAction_EmptyMicroflowArgumentUsesMicroflowParameterValue` continues to pass.
- [x] `go test ./...` passes.
- [x] Roundtrip audit on a 20-microflow random sample: previously failing `DataLake.X` (ParameterizedEntityType) and `MxKafka.X` (StringType) microflows now match, 18/18 unaffected microflows held their prior status — no regressions.


ako:
## Code Review

Correct and clean fix. The `default:` branch in the switch is the right approach — future `BasicParameterType` subtypes will automatically get the right behaviour without needing an explicit list. The `jaDef == nil` fallback to `""` is well-reasoned.

### Required before merge

- **No `mdl-examples/bug-tests/521-*.mdl`** — checklist requires a bug-test script for every bug fix. A short script binding a ListType or String parameter to `empty` with a `-- @expected: 0 errors` annotation covers it.
- **`.claude/skills/fix-issue.md` symptom table not updated** — CE0126 triggered by `empty` on a typed Java action parameter is a new entry for the symptom-to-layer-to-file mapping.

### Minor

- `ParameterizedEntityType` is called out in the PR description and roundtrip audit but not in the parametrized test. Since it falls into `default:` the logic is identical — a third case would complete the coverage for the types explicitly mentioned.
- The comment block above `resolvedBasicParams` is thorough but could be trimmed to 2–3 lines; the full rationale is already in the proposal doc it cites.

hjotha:
Thanks @ako — addressed in `faee3363`.

**Required**

- **`mdl-examples/bug-tests/521-empty-java-action-typed-arg.mdl`** — new bug-test calls `BugTest521.JA_ProcessBatch(Items = empty, Tag = empty)` exercising both list-typed and string-typed `BasicParameterType` slots bound to MDL `empty`. Before the fix this trips CE0126 "Missing value for parameter X"; after the fix `mx check` reports 0 errors and both arguments serialise as `Argument: "empty"`.
- **`.claude/skills/fix-issue.md`** — new symptom row mapping CE0126 on `call java action ... ($Param = empty)` to `addCallJavaActionAction` and the `resolvedBasicParams` fix pattern. Future contributors hitting the same symptom can land on the right file directly.

ako:
## Updated Review

Both required items from the previous review are addressed:

- **Bug test** — `521-empty-java-action-typed-arg.mdl` added, covering ListType, StringType, and ParameterizedEntityType cases with clear symptom and validation steps.
- **`fix-issue.md` symptom table** — CE0126 / typed Java action `empty` entry added with the full layer / file / remedy columns.

`ParameterizedEntityType` is still not in the Go parametrized test (still just `list` and `string`), but since the fix is handled by the `default:` branch the behaviour is identical — not a blocker.

Ready to merge.

PR #522, opened 2026-05-05
fix: classify reverse-Reference traversal through entity inheritance

hjotha:
## Summary

`addRetrieveAction` decided whether `retrieve $X from $V/M.Assoc` was a reverse traversal (child-side → parent-side, returning a list) by comparing the start variable's entity type to the association's child entity name **exactly**. Inheritance broke that check: when the start variable was a subclass of the association's child entity, the comparison missed and the result variable was registered as the parent entity *singleton* instead of `List of <parent>`.

Downstream `find($List, attr = value)` then dispatched to `FindByExpression` instead of `Find` because `resolveListOperationMember` read the variable type as a non-list. Studio Pro authors `Find` for the simple equality shape; emitting `FindByExpression` triggers `mx check` `CE0117 "Error in expression at List operation activity 'Find by expression'"` because the runtime can no longer see the qualified attribute path the model expects.

## Reproduction

A Reference association with `Parent=Header / Child=Message`. A microflow holds a `$Request` variable typed as a `Request` entity that **extends `Message`**. Traversing `$Request/Module.Headers` should yield `List of Header`:

```mdl
$HeaderList = retrieve $Request/Module.Headers;
$Match = find($HeaderList, Key = 'Authorization');
```

Before this PR: `mx check` reports `CE0117` at the `Find by expression` activity. After this PR: `mx check` reports 0 errors and the result variable is correctly typed as `List of Module.Header`.

## Fix

New `entityIsSubtypeOf` helper performs an inheritance-aware walk (mirrors the existing `resolveAttributeInEntityHierarchy` traversal). The exact-match `startVarType == childEntityQN` checks in `addRetrieveAction` are replaced with `entityIsSubtypeOf(startVarType, childEntityQN)`. The replacement covers both the `expandReverseReference` `DatabaseRetrieveSource` branch and the `AssociationRetrieveSource` fallback that registers the result type. Direct (non-inheriting) traversal still works the same way; the helper short-circuits on identity.

## Test plan

- [x] New `TestAddRetrieveAction_ReverseRefThroughInheritedChild` — reverse Reference through an inherited child entity registers `List of <parent>`.
- [x] New `TestAddRetrieveAction_ReverseRefDirectChild` — guards the non-inheritance path: exact child match continues to register `List of <parent>`.
- [x] Existing reverse-Reference / owner-both / non-persistable-parent tests continue to pass (`go test ./mdl/executor/`).
- [x] `go test ./...` passes.
- [x] Roundtrip audit on the 20-microflow random sample that surfaced the bug: previously failing microflow now matches (mpr_check_failed → match), 19/19 unaffected microflows held their prior status — no regressions.

### DIFF
--- a/mdl/executor/cmd_microflows_builder_actions.go
+++ b/mdl/executor/cmd_microflows_builder_actions.go
@@ -541,6 +541,14 @@ func (fb *flowBuilder) addRetrieveAction(s *ast.RetrieveStmt) model.ID {
 
 		outputUsedAsList := fb.listInputVariables != nil && fb.listInputVariables[s.Variable]
 		outputUsedAsObject := fb.objectInputVariables != nil && fb.objectInputVariables[s.Variable]
+		// startsFromChildSide is true when the retrieve's start variable is the
+		// child side of the association (or a subclass of it). Inheritance has
+		// to be honoured so traversals like `$httpRequest/System.HttpHeaders`
+		// — where HttpRequest extends HttpMessage and HttpHeaders has child
+		// HttpMessage — are still classified as reverse traversal.
+		startsFromChildSide := assocInfo != nil &&
+			assocInfo.childEntityQN != "" &&
+			fb.entityIsSubtypeOf(startVarType, assocInfo.childEntityQN)
 		// Owner-both Reference associations need later usage context: the same
 		// compact retrieve can be consumed as either a list or a single object.
 		// Owner="" means metadata was unavailable, so keep the association source.
@@ -549,7 +557,7 @@ func (fb *flowBuilder) addRetrieveAction(s *ast.RetrieveStmt) model.ID {
 			assocInfo.Owner != "" &&
 			assocInfo.parentPersistable &&
 			assocInfo.childEntityQN != "" &&
-			startVarType == assocInfo.childEntityQN &&
+			startsFromChildSide &&
 			(assocInfo.Owner != domainmodel.AssociationOwnerBoth || (outputUsedAsList && !outputUsedAsObject))
 
 		if expandReverseReference {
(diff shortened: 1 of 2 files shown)


ako:
## Code Review

**Verdict: approve with one follow-up item**

The fix is correct and well-scoped. `entityIsSubtypeOf` is a faithful mirror of the existing `resolveAttributeInEntityHierarchy` sibling — same `seen` cycle guard, same `GetModuleByName` + `GetDomainModel` + `GeneralizationRef` chain walk. All 4 `startVarType == assocInfo.childEntityQN` sites are replaced consistently, and `startsFromChildSide` short-circuits on `assocInfo == nil` so there's no backend cost for non-association retrieves.

### Moderate — missing `mdl-examples/bug-tests/` entry

The checklist requires a bug-test MDL script alongside every fix. The PR description already contains the exact reproducer — please commit it as `mdl-examples/bug-tests/522-reverse-reference-inheritance.mdl`:

```mdl
-- Reproducer for #522: reverse Reference traversal through entity inheritance.
-- Before fix: find() emitted FindByExpression → CE0117 in mx check.
create module InhTest;
create persistent entity InhTest.Header (Key: string);
create persistent entity InhTest.Message ();
create persistent entity InhTest.Request extends InhTest.Message ();
create association InhTest.Header_Message from InhTest.Header to InhTest.Message;
create microflow InhTest.MF_FindHeader ($Request: InhTest.Request) begin
  $HeaderList = retrieve $Request/InhTest.Header_Message;
  $Match = find($HeaderList, Key = 'Authorization');
  return $Match;
end;
```

### Minor

- **`fix-issue.md` symptom table**: add a row mapping `CE0117 at Find-by-expression on a find() result` → `addRetrieveAction / entityIsSubtypeOf` in `cmd_microflows_builder_actions.go`.
- **Redundant guard**: `assocInfo.childEntityQN != ""` stays inside `expandReverseReference` even though `startsFromChildSide` already implies it. Harmless — leave or remove at your discretion.

PR #533, opened 2026-05-08
fix: prevent catalog crash on duplicate business event channels

dhruvbehl:
## Summary

- `INSERT OR IGNORE INTO contract_messages` to tolerate duplicate rows from Mendix Portal's AsyncAPI documents

## Problem

Mendix-portal's business event module has duplicate channel references across services. This caused a UNIQUE constraint violation that crashed the entire catalog build, making `REFRESH CATALOG FULL` fail — which blocks all cross-reference queries (callers, callees, impact analysis).

image

## Test plan

- [x] `go build ./mdl/catalog/` passes
- [x] `go test ./mdl/catalog/` passes (190 tests)
- [x] Run `REFRESH CATALOG FULL` against a Mendix Portal MPR to confirm no crash
- [x] Verify `SHOW CONTRACT MESSAGES` returns expected rows

image

image

### DIFF
--- a/mdl/catalog/builder_contract.go
+++ b/mdl/catalog/builder_contract.go
@@ -138,7 +138,7 @@ func (b *Builder) buildContractMessages() error {
 	}
 
 	stmt, err := b.tx.Prepare(`
-		INSERT INTO contract_messages (Id, ServiceId, ServiceQualifiedName,
+		INSERT OR IGNORE INTO contract_messages (Id, ServiceId, ServiceQualifiedName,
 			ChannelName, OperationType, MessageName, Title, ContentType, PropertyCount,
 			ModuleName, ProjectId, SnapshotId, SnapshotDate, SnapshotSource)
 		VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)


ako (commented):
## Code Review

**Fix is correct and well-scoped.** One-line change with a clear, reproducible crash behind it.

### What the fix does

The synthetic  for a  row is a SHA-256 of `svcQN + "|channel|" + ch.Name + "|" + ch.MessageRef`. When Mendix Portal's AsyncAPI document repeats the same channel+messageRef combination, the same hash is generated twice, hitting the `Id TEXT PRIMARY KEY` constraint. `INSERT OR IGNORE` is the right idiom here — duplicate IDs mean genuinely identical data (same hash inputs), so silently dropping the second row loses nothing.

### Minor notes

**Potential gap in sibling tables** — `contract_entities` and `contract_actions` use plain `INSERT INTO` with the same SHA-256 synthetic-ID approach. A malformed `$metadata` document with duplicate entity or action names would hit the same crash. Low priority since OData documents tend to be well-formed, but worth a follow-up.

**Missing bug-test script** — per project checklist, bug fixes should include an `mdl-examples/bug-tests/` MDL script that reproduces the symptom. The manual test plan with screenshots is a good substitute for Studio Pro validation, but a short `mdl-examples/bug-tests/533-duplicate-contract-messages.mdl` would let the regression be caught automatically in CI.

**`fix-issue.md` not updated** — the symptom (`REFRESH CATALOG FULL` crash → UNIQUE constraint on `contract_messages`) is not in `.claude/skills/fix-issue.md`. New contributors tracing a similar crash in the future won't find it in the map.

### Verdict

The fix is safe and correct. The missing bug-test and fix-issue.md entry are process gaps rather than blockers — given that this crashed a production catalog build, landing the fix quickly is the right call.

ako (commented):
## Code Review

**Fix is correct and well-scoped.** One-line change with a clear, reproducible crash behind it.

### What the fix does

The synthetic `Id` for a `contract_messages` row is a SHA-256 of `svcQN + "|channel|" + ch.Name + "|" + ch.MessageRef`. When Mendix Portal's AsyncAPI document repeats the same channel+messageRef combination, the same hash is generated twice, hitting the `Id TEXT PRIMARY KEY` constraint. `INSERT OR IGNORE` is the right idiom here — duplicate IDs mean genuinely identical data (same hash inputs), so silently dropping the second row loses nothing.

### Minor notes

**Potential gap in sibling tables** — `contract_entities` and `contract_actions` use plain `INSERT INTO` with the same SHA-256 synthetic-ID approach. A malformed `$metadata` document with duplicate entity or action names would hit the same crash. Low priority since OData documents tend to be well-formed, but worth a follow-up.

**Missing bug-test script** — per project checklist, bug fixes should include an `mdl-examples/bug-tests/` MDL script that reproduces the symptom. The manual test plan with screenshots is a good substitute for Studio Pro validation, but a short `mdl-examples/bug-tests/533-duplicate-contract-messages.mdl` would let the regression be caught automatically in CI.

**`fix-issue.md` not updated** — the symptom (`REFRESH CATALOG FULL` crash → UNIQUE constraint on `contract_messages`) is not in `.claude/skills/fix-issue.md`. New contributors tracing a similar crash in the future won't find it in the map.

### Verdict

The fix is safe and correct. The missing bug-test and `fix-issue.md` entry are process gaps rather than blockers — given that this crashed a production catalog build, landing the fix quickly is the right call.

PR #596, opened 2026-05-23
fix #595: PublishedODataService EntityTypePointer wired by qualified entity name

ako:
## Summary

Fixes #595. `serializePublishedODataService` keyed `entityTypeIDMap` by `et.ExposedName` (e.g. `"Customers"`) but looked it up by `es.EntityTypeName` (e.g. `"OdTest.Customer"`). The lookup always returned an empty string, so `EntityTypePointer` was never written on `ODataPublish$EntitySet` BSON. Studio Pro's `EntitySet.Check` then NREs dereferencing the missing pointer and aborts the entire background project checker.

One-line fix in `sdk/mpr/writer_odata.go`: key the map by `et.Entity` to match what `astEntityDefToModel` stores on the entity set.

## Why the test missed this

`TestSerializePublishedODataService` checked all the EntitySet's own fields (`ExposedName`, `UsePaging`, `ReadMode`, etc.) but never the cross-reference back to the EntityType. The PR adds that assertion. It also switches the test entity-type `BaseElement.ID` from the placeholder `"et-1"` to a real UUID so `idToBsonBinary` doesn't silently fall back to random IDs, making the pointer comparison meaningful.

## Test plan

- [x] New unit assertion fails without the fix:
  ```
  writer_odata_test.go:294: EntityTypePointer: expected primitive.Binary, got <nil> (<nil>)
  ```
- [x] After the fix:
  ```
  --- PASS: TestSerializePublishedODataService (0.00s)
  ```
- [x] `make test` — all packages pass
- [x] `make lint` — Go + TypeScript pass
- [x] `./bin/mxcli check mdl-examples/bug-tests/595-published-odata-entitytypepointer.mdl` → `✓ Syntax OK (3 statements)`
- [ ] **Manual validation in Studio Pro** (user): apply the bug-test MDL to a project, open in Studio Pro, confirm no `NullReferenceException` in `EntitySet.Check` and the project checker emits real problems normally

## Related

- #590 — sibling Studio Pro NRE (`CallExternalAction.SchemaAction`), fixed in PR #592 as an MDL example-file change
- #594 — sibling Studio Pro NRE (`ODataRemoteEntitySource.RemoteId`), example-file workaround in PR #592; executor fix pending
- PR #592 — MDL example-file workarounds for #590 and #594

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #595, opened 2026-05-23
PublishedODataService EntityTypePointer never set — Studio Pro NRE in EntitySet.Check


ako:
## Summary

`serializePublishedODataService` (`sdk/mpr/writer_odata.go:248`) tries to resolve each EntitySet's owning EntityType through an internal map keyed by the **wrong field**. The lookup always returns empty, so the `EntityTypePointer` field is never written on `ODataPublish\$EntitySet` sub-documents. Studio Pro's `EntitySet.Check` (`Mendix.Modeler.Integration.ODataServices.Publish.EntitySet.Check`, line 53) then dereferences `null` and aborts the entire project checker with a NullReferenceException — hiding all other problems in the project, not just the published OData service.

## Stack trace

```
System.AggregateException: One or more errors occurred. (Object reference not set to an instance of an object.)
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Mendix.Modeler.Integration.ODataServices.Publish.EntitySet.Check(IProblems problems)
      in Mendix.Modeler.Integration/ODataServices/Publish/EntitySet.cs:line 53
   at Mendix.Modeler.Storage.TypeWalks.TypeWalk.ForEachObjectDo[TObject](...)
   ...
   at Mendix.Modeler.Core.View.Problems.BackgroundCheckRunner.Check(IProject project, CancellationToken cancellationToken)
```

## Root cause

In `sdk/mpr/writer_odata.go`:

```go
// Line 235 — map keyed by ExposedName
entityTypeIDMap[et.ExposedName] = etID

// Line 248 — lookup by EntityTypeName (which holds the qualified entity name)
entityTypeID := entityTypeIDMap[es.EntityTypeName]
```

`PublishedEntityType.ExposedName` is the OData-facing name (e.g. `\"Customers\"`).
`PublishedEntitySet.EntityTypeName` is the qualified entity name (e.g. `\"OdTest.Customer\"`, set in `astEntityDefToModel` at `mdl/executor/cmd_odata.go:1544`).

These are different values, so the lookup returns `\"\"`. In `serializePublishedEntitySet` (line 305):

```go
if entityTypeID != \"\" {
    doc[\"EntityTypePointer\"] = idToBsonBinary(entityTypeID)
}
```

The condition is always false → `EntityTypePointer` is never written. Studio Pro then can't navigate from EntitySet to its EntityType and NREs.

## Reproduction

[...]

Issue #590, opened 2026-05-23
10-odata-examples.mdl Level 9.1 crashes Studio Pro project check (NRE in CallExternalAction.SchemaAction)


ako:
## Summary

Running `mdl-examples/doctype-tests/10-odata-examples.mdl` against a project leaves the project in a state that crashes Studio Pro's background project checker with a `System.NullReferenceException` in `Mendix.Modeler.Integration.CallExternalAction.get_SchemaAction()`.

The crash blocks all in-IDE error reporting (the checker dies before reporting any problems), so the project appears 'silent' even though Studio Pro is actually unable to evaluate it.

## Stack trace

```
System.AggregateException: One or more errors occurred. (Object reference not set to an instance of an object.)
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Mendix.Modeler.Integration.CallExternalAction.get_SchemaAction()  CallExternalAction.cs:35
   at Mendix.Modeler.Integration.ExternalActionParameterMapping.get_SchemaParameter()  ExternalActionParameterMapping.cs:23
   at Mendix.Modeler.Integration.ExternalActionParameterMapping.Check(IProblems problems)  ExternalActionParameterMapping.cs:40
   at Mendix.Modeler.Storage.TypeWalks.TypeWalk.ForEachObjectDo[TObject](...)
   ...
   at Mendix.Modeler.Core.View.Problems.BackgroundCheckRunner.Check(IProject project, CancellationToken)
```

## Root cause

The example file's Level 9.1 generates a microflow `OdTest.ACT_CallExternalAction` containing two `call external action` activities against `OdTest.SalesforceAPI`:

```mdl
$Result = call external action OdTest.SalesforceAPI.GetAccounts(status = 'active', Region = $SelectedRegion);
call external action OdTest.SalesforceAPI.NotifyUpdate(AccountId = $AccountId) on error continue;
```

`OdTest.SalesforceAPI` is created in Level 8.1 with `MetadataUrl: 'https://api.salesforce.com/odata/v4/$metadata'` — a placeholder URL that doesn't actually return an OData `$metadata` document. After `CREATE ODATA CLIENT`, the resulting `Rest\$ConsumedODataService` BSON has:

```json
{
  \"Metadata\": \"\",
  \"MetadataHash\": \"\",
  \"Validated\": false,
  \"MetadataReferences\": [0],
  \"ValidatedEntities\": [0],
  ...
}
```

[...]

PR #598, opened 2026-05-24
fix #594: preserve omitted scalar fields on `create or modify external entity`

### LINKED ISSUE #594: create or modify external entity wipes omitted RemoteName/EntityTypeName (Studio Pro NRE in ODataRemoteEntitySource.RemoteId)
## Summary

`create or modify external entity` overwrites `RemoteEntityName` (the BSON `RemoteName` field on `Rest\$ODataRemoteEntitySource`) with the empty string when the `RemoteName` property is omitted from the statement. The resulting `Rest\$ODataRemoteEntitySource` BSON has `\"RemoteName\": \"\"`, which Studio Pro's Integration Pane visualiser cannot render — `ODataRemoteEntitySource.get_RemoteId()` throws `ArgumentNullException(\"EntityTypeName\")` (Studio Pro internally aliases the BSON `RemoteName` field to the C# property `EntityTypeName`).

Studio Pro should also null-check here (separate upstream robustness gap), but the root cause on our side is that omitted properties in `or modify` are being treated as \"reset to empty\" rather than \"preserve existing value\".

## Reproduction

```mdl
-- First create with RemoteName set
create external entity OdTest.RemoteAccount
from odata client OdTest.SalesforceAPI
(
  EntitySet: 'Accounts',
  RemoteName: 'Account',
  Countable: Yes
)
( AccountId: string(200) );

-- Then `or modify` it WITHOUT RemoteName
create or modify external entity OdTest.RemoteAccount
from odata client OdTest.SalesforceAPI
(
  EntitySet: 'Accounts',
  Countable: Yes,
  AllowCreateChangeLocally: Yes
)
( AccountId: string(200) );
```

Resulting BSON (from a real project, observed after running `mdl-examples/doctype-tests/10-odata-examples.mdl`):

```json
{
  \"\$Type\": \"Rest\$ODataRemoteEntitySource\",
  \"EntitySet\": \"Accounts\",
  \"RemoteName\": \"\",
  \"SourceDocument\": \"OdTest.SalesforceAPI\",
  ...
}
```

Then opening the project in Studio Pro and switching to the Integration view triggers:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'EntityTypeName')
   at Mendix.Modeler.Integration.ODataRemoteEntitySource.get_RemoteId() in
      Mendix.Modeler.Integration/ODataServices/Consume/ODataRemoteEntitySource.cs:line 51
   at Mendix.Modeler.Integration.View.ODataServices.Consume.ODataApiVisualizer ...
   at Mendix.Modeler.IntegrationOverview.View.LandscapeSearch.GetUsedInApp()
   at Mendix.Modeler.IntegrationOverview.View.Web.IntegrationPaneWebApiHandler.HandleUsedInAppRequest()
   ...
```

## Root cause

`mdl/executor/cmd_odata.go:818` (the `or modify` branch of the create-external-entity handler):

```go
if existingEntity != nil {
    existingEntity.Source = \"Rest\$ODataRemoteEntitySource\"
    existingEntity.RemoteServiceName = serviceRef
    existingEntity.RemoteEntitySet = s.EntitySet
    existingEntity.RemoteEntityName = s.RemoteName  // <-- overwrites with \"\" when omitted
    existingEntity.Countable = s.Countable
    ...
}
```

When the MDL parser sees no `RemoteName` property, `s.RemoteName` is `\"\"`. The handler unconditionally assigns it, wiping the prior non-empty value.

The non-modify branch (line 849 — initial create) has the same shape but is correct: the BSON for a fresh create is well-formed even with empty `RemoteName` (Studio Pro probably defaults it). The bug is specifically in the modify path overwriting a pre-existing valid value.

## Proposed fix

Guard the assignment so omitted properties preserve existing values:

```go
if s.RemoteName != \"\" {
    existingEntity.RemoteEntityName = s.RemoteName
}
```

The same review should be applied to other fields in this block — `EntitySet`, `Countable`, `Creatable`, `Deletable`, `Updatable`, `CreateChangeLocally`, `Documentation`. The current handler already guards `Documentation` and `Attributes` (lines 824-829) but treats the boolean/scalar fields as required-on-every-modify.

Two design options:

1. **Preserve when omitted** — matches SQL `UPDATE ... SET column = X` semantics. Requires the parser to distinguish \"omitted\" from \"explicitly empty\" — currently the AST collapses both to `\"\"`. Boolean fields would need a `*bool` AST type (or a separate \"was set\" bitmap).
2. **Require explicit RemoteName on `or modify`** — simpler but breaks the upsert pattern.

Option 1 is the right long-term fix; option 2 is a stopgap that gives a clear error rather than silent corruption.

## Workaround (example file)

`mdl-examples/doctype-tests/10-odata-examples.mdl` Level 8.6c is being patched in PR #592 to explicitly set `RemoteName: 'Account'`, sidestepping the bug for the example file. The underlying executor bug remains.

## Related

- Issue #590 — sibling Studio Pro NRE (CallExternalAction.SchemaAction) from the same example file
- PR #592 — fixes the example file for both NREs; this issue tracks the executor fix

ako:
## Summary

Fixes #594. `execCreateExternalEntity` unconditionally wrote `existingEntity.X = s.X` for every property in its `or modify` branch. When the MDL statement omitted `RemoteName`, `s.RemoteName` was the empty string and the pre-existing value was wiped. The resulting `Rest$ODataRemoteEntitySource` BSON had `RemoteName: ""`, and Studio Pro's Integration Pane visualiser then NREs in `ODataRemoteEntitySource.get_RemoteId()` with `ArgumentNullException("EntityTypeName")` (Studio Pro aliases the BSON `RemoteName` field as the C# property `EntityTypeName`).

## Approach

Convert the scalar fields on `CreateExternalEntityStmt` (`EntitySet`, `RemoteName`, `Countable`, `Creatable`, `Deletable`, `Updatable`, `AllowCreateChangeLocally`) to pointers so the executor can distinguish:

- **nil** → field omitted from the statement → preserve existing value on `or modify`
- **non-nil** → explicitly set by the user (including to `false` or `""`) → assign

This matches the existing tri-state convention already used for `AlterProjectSecurityStmt.DemoUsersEnabled *bool`. The visitor allocates pointers only inside each `case` of the property switch. The executor gates every assignment on non-nil on the modify branch; the initial-create branch still requires `EntitySet` (no prior value to preserve) and returns a validation error if it's omitted.

## Test plan

- [x] New mock-backend test (`TestCreateOrModifyExternalEntity_PreservesOmittedFields_Issue594`) constructs an existing entity with non-default values for every clobberable field, runs `or modify` with only `EntitySet` + `AllowCreateChangeLocally` set, and asserts all other fields retain their pre-existing values.
- [x] Verified the test catches the bug — temporarily reverting just the `RemoteName` branch to the old unconditional assignment reports `RemoteEntityName = "", want "Account"` and the test fails.
- [x] `make build` — passes
- [x] `make test` — all packages green
- [x] `make lint` — Go + TypeScript pass
- [x] `./bin/mxcli check mdl-examples/bug-tests/594-or-modify-external-entity-preserve-omitted.mdl` → `✓ Syntax OK (5 statements)`
- [ ] **Manual validation in Studio Pro** (user): apply the bug-test MDL to a project, confirm `RemoteAccount` still has `RemoteName: "Account"` after the `or modify`, and the Integration Pane renders without NRE.

## Compatibility

The AST change is package-internal — `CreateExternalEntityStmt` lives in `mdl/ast` and no external code depends on its fields. Visitor, executor, validate, validate_duplicates, and stmt_summary are all updated. The only existing call sites in `cmd_odata_mock_test.go` were migrated to local `strPtr`/`boolPtr` helpers.

## Related

- #594 — the underlying executor bug (this PR)
- #590, #595 — sibling Studio Pro NREs from the same example file (PRs #592, #596)
- PR #592 — example-file workaround that re-stated `RemoteName: 'Account'` in Level 8.6c; still defensive but no longer required after this fix lands

### DIFF
--- a/mdl-examples/bug-tests/594-or-modify-external-entity-preserve-omitted.mdl
+++ b/mdl-examples/bug-tests/594-or-modify-external-entity-preserve-omitted.mdl
@@ -1,0 +1,61 @@
+-- Bug test: `create or modify external entity` must preserve fields the
+-- statement omits rather than overwriting them with zero values.
+--
+-- Issue #594: the executor (mdl/executor/cmd_odata.go) unconditionally wrote
+-- `existingEntity.X = s.X` for every property. When the MDL statement omitted
+-- `RemoteName`, `s.RemoteName` was the empty string and the existing value
+-- was wiped. The resulting `Rest$ODataRemoteEntitySource` BSON had
+-- `RemoteName: ""`, which Studio Pro's Integration Pane visualiser then
+-- dereferences in `ODataRemoteEntitySource.get_RemoteId()` and NREs with
+-- ArgumentNullException("EntityTypeName") — Studio Pro aliases the BSON
+-- `RemoteName` field as the C# property `EntityTypeName`.
+--
+-- The Go unit test in mdl/executor/cmd_odata_mock_test.go
+-- (TestCreateOrModifyExternalEntity_PreservesOmittedFields_Issue594) covers
+-- the executor logic with a mock backend. This MDL script is the end-to-end
+-- regression test: applying it to a real project must leave a
+-- bug594.RemoteAccount with RemoteName = "Account" after the modify, and
+-- the project must open in Studio Pro without the Integration Pane NRE.
+--
+-- Run with:
+--   mxcli exec mdl-examples/bug-tests/594-or-modify-external-entity-preserve-omitted.mdl -p <project>.mpr
+--   mxcli describe entity bug594.RemoteAccount -p <project>.mpr   # RemoteName must still be 'Account'
+--   then open <project>.mpr in Studio Pro and confirm the Integration Pane renders.
+create module bug594;
+
+create odata client bug594.SalesforceAPI (
+  ODataVersion: OData4,
+  MetadataUrl: 'https://api.salesforce.com/odata/v4/$metadata',
+  timeout: 300
+);
+
+create external entity bug594.RemoteAccount
+from odata client bug594.SalesforceAPI
+(
+  EntitySet: 'Accounts',
+  RemoteName: 'Account',
+  Countable: Yes,
+  Creatable: Yes,
+  Deletable: No,
+  Updatable: No
+)
+(
+  AccountId: string(200),
+  AccountName: string(255)
+);
+
+-- Modify without restating RemoteName / Countable / Creatable. Before the
+-- fix, every omitted property was overwritten with its zero value and the
+-- entity ended up unusable. After the fix, all of them are preserved.
+create or modify external entity bug594.RemoteAccount
+from odata client bug594.SalesforceAPI
+(
+  EntitySet: 'Accounts',
+  AllowCreateChangeLocally: Yes
+)
+(
+  AccountId: string(200),
+  AccountName: string(255)
+);
+
+describe entity bug594.RemoteAccount;
--- a/mdl/ast/ast_odata.go
+++ b/mdl/ast/ast_odata.go
@@ -121,16 +121,21 @@ type DropODataServiceStmt struct {
 func (s *DropODataServiceStmt) isStatement() {}
 
 // CreateExternalEntityStmt represents: CREATE [OR MODIFY] EXTERNAL ENTITY Module.Name FROM ODATA CLIENT Module.Service (...) (attrs);
+//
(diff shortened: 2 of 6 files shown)


ako:
Already landed in main

PR #683, opened 2026-06-23
feat(lint): add --rules / -r flag to isolate rule execution

Andries-Smit:
## Summary

- Adds `-r` / `--rules` flag to `mxcli lint` that restricts execution to the specified rule IDs
- When the flag is omitted all rules run as before — no behaviour change
- `--list-rules` is unaffected and still shows the full rule catalogue

## Motivation

During lint rule development, running all ~40 rules on every iteration is slow and noisy. This flag lets you pin execution to the rule(s) under development:

```bash
mxcli lint -p app.mpr -r MPR001
mxcli lint -p app.mpr -r MPR001 -r SEC001
mxcli lint -p app.mpr -r MPR001,SEC001   # comma form also works
```

## Implementation

Reuses the existing `Linter.ConfigureRule(id, RuleConfig{Enabled: false})` mechanism — no changes to `linter.go` or any rule files. After all rules (including Starlark) are registered, any rule whose ID is not in the allowlist is disabled before `lint.Run()` is called.

## Test plan

- [ ] `mxcli lint -p app.mpr -r MPR001` — only MPR001 violations appear
- [ ] `mxcli lint -p app.mpr -r MPR001 -r SEC001` — only those two rules run
- [ ] `mxcli lint -p app.mpr` — all rules still run (no regression)
- [ ] `mxcli lint -p app.mpr -r MPR001 --list-rules` — full rule catalogue shown

https://claude.ai/code/session_01KeDQV2xS3xL16vEZWEDmaw

### DIFF
--- a/cmd/mxcli/cmd_lint.go
+++ b/cmd/mxcli/cmd_lint.go
@@ -77,6 +77,9 @@ Examples:
   mxcli lint -p app.mpr --format json
   mxcli lint -p app.mpr --format sarif > results.sarif
   mxcli lint -p app.mpr --list-rules
+  mxcli lint -p app.mpr -r MPR001
+  mxcli lint -p app.mpr -r MPR001 -r SEC001
+  mxcli lint -p app.mpr -r MPR001,SEC001
 `,
 	Run: func(cmd *cobra.Command, args []string) {
 		projectPath, _ := cmd.Flags().GetString("project")
--- a/cmd/mxcli/cmd_lint.go
+++ b/cmd/mxcli/cmd_lint.go
@@ -84,6 +84,7 @@ Examples:
 		useColor, _ := cmd.Flags().GetBool("color")
 		listRules, _ := cmd.Flags().GetBool("list-rules")
 		excludeModules, _ := cmd.Flags().GetStringSlice("exclude")
+		onlyRules, _ := cmd.Flags().GetStringSlice("rules")
 
 		if projectPath == "" {
 			fmt.Fprintln(os.Stderr, "Error: --project (-p) is required")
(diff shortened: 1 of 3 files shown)


Andries-Smit:
Thanks for the review. Addressed in the latest push:

**Critical — comma-separated parsing**: This is actually a non-issue. `StringSliceP` in pflag/cobra [splits on commas by design](https://pkg.go.dev/github.com/spf13/pflag#StringSlice) — `-r MPR001,SEC001` produces `[]string{"MPR001", "SEC001"}` as expected. Verified with a quick unit test.

**Minor — missing test coverage**: Added `mdl/linter/linter_test.go` with three tests covering the filter logic:
- single rule allowlist
- multi-rule allowlist  
- empty allowlist (all rules run, no regression)

**Minor — help text**: Added comma-form and multi-flag examples to the `mxcli lint` help text.

Andries-Smit:
Manual tested and can confirm, the Comma-separated rule IDs do work. Its a false positive

ako:
Reviewed the implementation — change is small, correct, and well-scoped. Confirmed the disputed point and verified the rest:

**Comma-separated parsing is a non-issue (agree with @Andries-Smit).** `pflag.StringSlice` parses through a CSV reader and splits on commas by design, so `-r MPR001,SEC001` yields `["MPR001","SEC001"]`. The bot's "critical" finding is a false positive.

Also verified:
- `--list-rules` is unaffected — `Linter.Rules()` returns the full slice; `Enabled:false` only gates `Run()` (`linter.go:141`).
- No regression when the flag is omitted — the whole block is guarded by `len(onlyRules) > 0`.
- New tests pass (`ok github.com/mendixlabs/mxcli/mdl/linter`).

Two minor, non-blocking gaps:

1. **Tests don't exercise the real code path.** `linter_test.go` re-implements the allowlist loop inline rather than driving `cmd_lint.go` / cobra flag parsing — so the comma-form (the one thing actually questioned) is only verified manually, not by a test. A test that parses `-r MPR001,SEC001` through the command would close that loop.
2. **Silent on typos.** A misspelled ID (`-r MPRO01`) disables every rule and reports zero violations with no warning — the "no silent side effects on typos" checklist item. A warning when an allowlisted ID matches no registered rule would help.

Neither blocks merge.

PR #684, opened 2026-06-23
feat(lint): add --modules/-m flag to scope lint to specific modules

Andries-Smit:
## Summary

- Adds `--modules / -m` flag to `mxcli lint` to filter linting to a specific set of modules (inclusion/allowlist), complementing the existing `--exclude` flag
- Useful during rule development to iterate quickly without noise from unrelated modules
- Accepts repeated flags (`-m A -m B`) or comma-separated values (`-m A,B`)

## Changes

**`mdl/linter/context.go`**
- Added `included map[string]bool` field to `LintContext`
- Added `SetIncludedModules([]string)` method
- Extended `IsExcluded()` to return `true` when an inclusion filter is active and the module is not in it
- Updated all 7 catalog iterators (`Entities`, `Microflows`, `Pages`, `Enumerations`, `Widgets`, `Snippets`, `DatabaseConnections`) to call `ctx.IsExcluded()` instead of reading `ctx.excluded` map directly — ensures the new filter is respected automatically everywhere

**`cmd/mxcli/main.go`**
- Registered `--modules / -m` (`StringSliceP`) flag for `lintCmd`

**`cmd/mxcli/cmd_lint.go`**
- Reads the new flag and calls `ctx.SetIncludedModules()` when set
- Added usage examples to the command `Long` description

## Usage

```bash
# Scope to a single module
mxcli lint -p app.mpr -m MyModule

# Scope to multiple modules
mxcli lint -p app.mpr -m MyModule -m AnotherModule
mxcli lint -p app.mpr -m MyModule,AnotherModule

# Combined with --exclude (exclude takes priority within the included set)
mxcli lint -p app.mpr -m MyModule --exclude SubModule
```

## Test plan

- [x] `go build ./...` — clean build
- [x] `go test ./mdl/linter/...` — all linter tests pass
- [ ] Manual: `mxcli lint -p app.mpr -m <module>` — violations only reference the specified module
- [ ] Manual: `mxcli lint --list-rules` — no regression

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014ZV9A7j6FFSjPnYJQjNJ5v

PR #685, opened 2026-06-23
feat(lint): expose constants() query function to Starlark lint rules

Andries-Smit:
## Summary

- Adds `constants()` as a new query function available in Starlark custom lint rules
- Exposes 8 properties per constant: `id`, `name`, `qualified_name`, `module_name`, `folder`, `description`, `default_value`, `exposed_to_client`
- Updates the `write-lint-rules.md` skill with the function signature and full property table

## Implementation

The catalog already populated `constants_data` / `constants` view from the MPR file. This PR wires that data through two layers that were missing:

1. **`mdl/linter/context.go`** — `LintConstant` struct + `Constants() iter.Seq[LintConstant]` method, querying the `constants` catalog view and respecting excluded-module filtering
2. **`mdl/linter/starlark.go`** — `builtinConstants()` method, `constantToStarlark()` converter, and registration in `buildPredeclared()`
3. **`.claude/skills/mendix/write-lint-rules.md`** — documents `constants()` in the Available Query Functions table and adds a `### constant` object properties section

## Example rule

```python
def check():
    return [
        violation(
            message = "Exposed constant has no default value: " + c.qualified_name,
            location = c.module_name,
        )
        for c in constants()
        if c.exposed_to_client and c.default_value == ""
    ]

RULES = [rule(id = "CONV001", name = "Exposed constants must have a default value", fn = check)]
```

## Test plan

- [x] `go build ./mdl/linter/...` passes
- [x] `go test ./mdl/linter/...` passes
- [ ] Manual: run `mxcli lint -p app.mpr --rules my_rules/` with a rule that calls `constants()` and confirm results

https://claude.ai/code/session_01Nq5xEDmnhVMd2HKVt1hiN3

### DIFF
--- a/.claude/skills/mendix/write-lint-rules.md
+++ b/.claude/skills/mendix/write-lint-rules.md
@@ -27,6 +27,7 @@ def check():
 | `microflows()` | list of microflow | All non-system microflows |
 | `pages()` | list of page | All non-system pages |
 | `enumerations()` | list of enumeration | All non-system enumerations |
+| `constants()` | list of constant | All non-system constants |
 | `widgets()` | list of widget | All non-system widgets |
 | `snippets()` | list of snippet | All non-system snippets |
 | `attributes_for(entity_qualified_name)` | list of attribute | Attributes for a specific entity |
@@ -145,6 +146,18 @@ def check():
 | `folder` | string | `"enumerations"` — folder path within module |
 | `description` | string | Documentation text |
 | `value_count` | int | Number of enum values |
+
+### constant
+| Property | Type | Example |
+|----------|------|---------|
+| `id` | string | Document UUID |
+| `name` | string | `"AppBaseUrl"` |
+| `qualified_name` | string | `"MyModule.AppBaseUrl"` |
+| `module_name` | string | `"MyModule"` |
+| `folder` | string | `"constants"` — folder path within module |
+| `description` | string | Documentation text |
+| `default_value` | string | `"https://example.com"` |
+| `exposed_to_client` | bool | `true` if constant is exposed to client |
 
 ### widget
 | Property | Type | Example |
--- a/mdl/linter/context.go
+++ b/mdl/linter/context.go
@@ -559,6 +559,59 @@ func (ctx *LintContext) Enumerations() iter.Seq[Enumeration] {
 	}
 }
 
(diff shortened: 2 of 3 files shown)


ako:
The implementation is clean and follows the established pattern (struct + iterator + Starlark builtin + converter + registration + docs). I verified all 8 selected columns (`Id, Name, QualifiedName, ModuleName, Folder, Description, DefaultValue, ExposedToClient`) exist in `constants_data` (`mdl/catalog/tables.go:755`), and the system-module filter matches the sibling iterators. It also merges cleanly into `main`.

**🔴 One real bug — `Constants()` won't respect the `--modules` filter (merge-base skew).** This branch was cut before #683/#684 and never rebased. Since then **#684 merged**, refactoring all 7 catalog iterators from `ctx.excluded[…]` to the centralized `ctx.IsExcluded(…)` so they honor the new `--modules/-m` include filter. The new iterator here still uses the old pattern:

```go
if ctx.excluded[c.ModuleName] {   // context.go, Constants()
```

The merge is *clean* (brand-new method in a fresh region, no textual conflict) — which is what makes this easy to miss. I materialized the merged tree to confirm: post-merge, `Constants()` is the **lone iterator of 8** still reading the map directly while the other 7 call `ctx.IsExcluded()`. Result: `mxcli lint -p app.mpr -m ModA` with a rule calling `constants()` would still see constants from *every* module. **Fix is one line:** `if ctx.IsExcluded(c.ModuleName)`.

**Non-blocking:**
- **No test added** — the test plan claims `go test ./mdl/linter/...` passes but the diff adds none. `main` now has `mdl/linter/context_test.go` (from #684) with `setupModuleFilterDB` / `NewLintContextFromDB` helpers; a `Constants()` inclusion/exclusion test is a few lines and would have caught the bug above.
- **Example rule reuses `id = "CONV001"`** in the description, which collides with the existing built-in CONV001 (Boolean naming) — use a fresh ID in the docs example.
- **`DataType` not exposed** — the catalog stores each constant's `DataType` but it isn't surfaced to Starlark. Optional, but plausibly useful (e.g. "secret-typed constants must not be exposed to client").

**Recommendation: request changes** for the one-line `ctx.IsExcluded()` fix (correctness with `--modules`), ideally plus a `Constants()` test using the now-available helpers. Everything else is solid.

PR #688, opened 2026-06-24
feat(lint): expose XPath/expression AST to Starlark rules

Andries-Smit:
Add two new builtins available to all `.star` lint rules:

- `xpath_expressions()` — iterates XPathExpressionEntry records from the catalog, covering RETRIEVE, SECURITY, and DATASOURCE usage types across microflows, access rules, and pages.

- `parse_xpath(expr)` — parses a raw XPath string into a structured Starlark tree using the existing exprcheck parser. Returns a recursive struct tree with a `kind` discriminator field and type-specific child fields (left/right for bin, operand for unary, args for call, etc.).

Internal additions:
- `starlark_xpath.go` — robustExprToStarlark() converter and xpathExpressionEntryToStarlark() mapper
- `starlark_xpath_test.go` — unit tests for the converter
- `LintContext.XPathExpressions()` iterator (context.go)

These builtins make it possible to write performance and security lint rules that inspect the parsed expression tree rather than raw strings — for example, flagging negative enumeration XPath conditions that prevent database index usage.

**Example Rule**
``` star
# PERF001: Prefer Positive XPath Conditions
#
# Negative conditions (!=, not()) in database retrieve constraints prevent the
# database from using indices, since it cannot index "what is not there".
# This specifically applies to enumeration attribute comparisons.
#
# Noncompliant: [EnumAttribute != Module.Enumeration.C]
# Compliant:    [EnumAttribute = Module.Enumeration.A or EnumAttribute = Module.Enumeration.B]
#
# Reference: Mendix performance best practices, negative XPath constraints.

RULE_ID = "PERF001"
RULE_NAME = "PositiveXPathCondition"
DESCRIPTION = "Flags negative enumeration XPath conditions (!=, not) in retrieve/security constraints — negative conditions prevent index usage"
CATEGORY = "performance"
SEVERITY = "warning"

def check():
    violations = []

    for e in xpath_expressions():
        if e.usage_type not in ("RETRIEVE", "SECURITY"):
            continue
        if not e.xpath_expression:
            continue

        ast = parse_xpath(e.xpath_expression)
        if _has_enum_negation(ast):
            violations.append(violation(
                message="Negative XPath condition in {} '{}' prevents index usage on '{}'".format(
                    e.document_type.lower(), e.document_qualified_name, e.target_entity
                ),
                location=location(
                    module=e.module_name,
                    document_type=e.document_type.lower(),
                    document_name=e.document_qualified_name,
                    document_id=e.document_id,
                ),
                suggestion="Rewrite '{}' using positive conditions (= and or) so the database can use indices".format(
                    e.xpath_expression
                ),
            ))

    return violations

# ---------------------------------------------------------------------------
# AST helpers
# ---------------------------------------------------------------------------

def _is_attr(node):
    """True for an attribute or association path: bare identifier (variable), attr_path, or 2-part qualified name."""
    return node.kind == "variable" or node.kind == "attr_path" or (node.kind == "qname" and node.sub == "")

def _is_enum_or_string(node):
    """True for an enumeration value (3-part qname) or a string literal."""
    return node.kind == "string" or (node.kind == "qname" and node.sub != "")

def _is_attr_enum_pair(left, right):
    """True when one side is an attribute path and the other is an enum/string literal."""
    return (_is_attr(left) and _is_enum_or_string(right)) or \
           (_is_enum_or_string(left) and _is_attr(right))

def _is_neq_pattern(node):
    """Matches: attr != Enum.Value"""
    return (node.kind == "bin" and node.op == "!=" and
            _is_attr_enum_pair(node.left, node.right))

def _is_not_eq_pattern(node):
    """Matches: not(attr = Enum.Value)"""
    if node.kind != "call" or node.name.lower() != "not":
        return False
    for arg in node.args:
        if (arg.kind == "bin" and arg.op == "=" and
                _is_attr_enum_pair(arg.left, arg.right)):
            return True
    return False

def _has_enum_negation(root):
    """Walks the AST using an explicit stack (no recursion, no while — Starlark restrictions).
    Uses for _ in range(N) as a bounded loop over a mutable work list.
    """
    if root == None:
        return False
    stack = [root]
    for _ in range(256):  # XPath ASTs are shallow in practice; 256 is a safe upper bound
        if not stack:
            break
        node = stack.pop()
        if node == None:
            continue
        k = node.kind

        # Leaf nodes — no children to inspect
        if k in ("null", "string", "number", "bool", "empty", "variable",
                 "attr_path", "qname", "constant", "token", "recovered", "unknown"):
            continue

        # Pattern check at this node
        if _is_neq_pattern(node) or _is_not_eq_pattern(node):
            return True

        # Push children onto the stack
        if k == "bin":
            stack.append(node.left)
            stack.append(node.right)
        elif k == "unary":
            stack.append(node.operand)
        elif k == "call":
            for a in node.args:
                stack.append(a)
        elif k == "paren":
            stack.append(node.inner)
        elif k == "if":
            stack.append(node.cond)
            stack.append(node.then)
            stack.append(node.else_)
    return False

```

### DIFF
--- a/.claude/skills/mendix/write-lint-rules.md
+++ b/.claude/skills/mendix/write-lint-rules.md
@@ -41,6 +41,7 @@ def check():
 | `module_roles()` | list of module_role | All module roles (deduplicated from role mappings) |
 | `role_mappings()` | list of role_mapping | User role to module role assignments |
 | `project_security()` | project_security or None | Project-level security settings (requires MPR reader) |
+| `xpath_expressions()` | list of xpath_expression | All XPath constraint expressions in the catalog (access rules, retrieve actions, widgets) |
 
 ### Graph-analysis functions (architecture rules)
 
@@ -193,6 +194,76 @@ def check():
 | `interval_seconds` | int | `86400` — `0` for unrecognised interval type |
 | `enabled` | bool | `True` if the event is active |
 
+### xpath_expression
+
+Returned by `xpath_expressions()`. Each row represents one XPath constraint used in a retrieve action, access rule, or widget data source.
+
+| Property | Type | Example |
+|----------|------|---------|
+| `id` | string | Row UUID |
+| `document_type` | string | `"MICROFLOW"`, `"NANOFLOW"`, `"DOMAIN_MODEL"`, `"PAGE"`, `"SNIPPET"` |
+| `document_id` | string | Owning document UUID |
+| `document_qualified_name` | string | `"MyApp.GetActiveItems"` |
+| `component_type` | string | `"RETRIEVE_ACTION"`, `"ACCESS_RULE"`, `"WIDGET"` |
+| `component_id` | string | Component UUID |
+| `component_name` | string | Activity/rule name (may be empty) |
+| `xpath_expression` | string | Raw XPath string, may include outer `[ ]` |
+| `target_entity` | string | Qualified name of entity being queried, e.g. `"MyApp.Order"` |
+| `referenced_entities` | string | Comma-separated qualified names of entities referenced by the XPath |
+| `is_parameterized` | bool | True when the XPath contains `$variable` references |
+| `usage_type` | string | `"RETRIEVE"`, `"SECURITY"`, `"DATASOURCE"` |
+| `module_name` | string | `"MyApp"` |
+
+### expr
+
+Returned by `parse_xpath(s)`. Every node has a `kind` field; additional fields depend on the kind.
+
+| `kind` | Additional fields | Description |
+|--------|-------------------|-------------|
+| `"bin"` | `op` (string), `left` (expr), `right` (expr) | Binary operator: `=`, `!=`, `<`, `>`, `<=`, `>=`, `and`, `or` |
+| `"unary"` | `op` (string), `operand` (expr) | Unary operator: `not`, `-` |
+| `"call"` | `name` (string), `args` (list of expr) | Function call, e.g. `contains(…)`, `length(…)` |
+| `"string"` | `value` (string) | String literal |
+| `"number"` | `value` (string) | Numeric literal (kept as string to preserve precision) |
+| `"bool"` | `value` (bool) | `true` or `false` |
+| `"empty"` | — | Mendix `empty` keyword |
+| `"variable"` | `name` (string) | `$ParameterName` |
+| `"attr_path"` | `variable` (string), `path` (list of string) | `$Obj/Association/Attribute` |
+| `"qname"` | `module` (string), `name` (string), `sub` (string) | Qualified name, e.g. `MyApp.Status.Active` |
+| `"paren"` | `inner` (expr) | Parenthesised expression |
+| `"if"` | `cond` (expr), `then` (expr), `else_` (expr) | If-then-else expression |
+| `"constant"` | `qname` (string) | Mendix constant reference, e.g. `[%MyConst%]` |
+| `"token"` | `token` (string), `arg` (string) | Mendix token expression, e.g. `[%CurrentUser%]` |
+| `"recovered"` | `source` (string), `reason` (string) | Parse failure — node carries the raw source fragment |
+| `"null"` | — | Nil / missing node |
+| `"unknown"` | — | Unrecognised AST node type |
+
+**Walking an expr tree:** check `node.kind` and recurse into child fields. Leaf kinds (no child nodes) are: `string`, `number`, `bool`, `empty`, `variable`, `qname`, `constant`, `token`, `recovered`, `null`, `unknown`.
+
+Example — count `not(…)` calls in an XPath (using `parse_xpath`):
+
+```python
+def count_not(node):
+    if node.kind in ("null", "unknown", "recovered", "string", "number",
+                     "bool", "empty", "variable", "qname", "constant", "token"):
+        return 0
+    if node.kind == "call" and node.name == "not":
+        return 1 + sum([count_not(a) for a in node.args])
+    if node.kind == "call":
+        return sum([count_not(a) for a in node.args])
+    if node.kind == "bin":
+        return count_not(node.left) + count_not(node.right)
+    if node.kind == "unary":
+        return count_not(node.operand)
+    if node.kind == "paren":
+        return count_not(node.inner)
+    if node.kind == "if":
+        return count_not(node.cond) + count_not(node.then) + count_not(node.else_)
+    if node.kind == "attr_path":
+        return 0
+    return 0
+```
+
 ### attribute
 | Property | Type | Example |
 |----------|------|---------|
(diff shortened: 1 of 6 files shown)


ako:
Strong PR — verified end-to-end on the branch.

**What's solid:**
- **Real, populated data path.** The query targets the `xpath_expressions` view, which exists on `main` (`tables.go:385`) and is filled by `builder_xpath.go`. All 13 selected columns exist, and the builder emits the `RETRIEVE`/`SECURITY`/`DATASOURCE` usage types the example rule filters on.
- **No merge-base skew.** Unlike #685, the new `XPathExpressions()` iterator correctly uses `ctx.IsExcluded(e.ModuleName)`, so it honors `--modules` and stays consistent with the other 7 iterators.
- **Tests included and passing.** Unit tests cover the AST converter + bracket stripping; the integration test drives a real Starlark rule through `xpath_expressions()` + `parse_xpath()` against a file-based catalog. `go vet` clean, `go test ./mdl/linter/` green (incl. `TestStarlarkXPathBuiltins`); the `exprcheck` field mappings all compile.
- Graceful failure: `parse_xpath` ignores the parse error and the converter maps `nil → kind:"null"`, so malformed input degrades to `recovered`/`null` rather than panicking.

**🟠 Main gap — no skill documentation (should block):** all 5 changed files are under `mdl/linter/`; nothing updates `.claude/skills/mendix/write-lint-rules.md`. That skill has the canonical "Available Query Functions" table + per-object property tables, and #685 set the precedent of documenting each new builtin there. This PR adds *two* builtins (`xpath_expressions()`, `parse_xpath()`), a new `xpath_expression` object (13 fields), **and** an `expr` AST node schema with 14 `kind` variants and type-specific fields (`left/right/op/operand/args/name/inner/cond/then/else_/path/variable/module/sub/value`). Without docs, rule authors can't discover these or learn the field names — and since this skill is synced into user projects via `make sync-skills`, it's invisible to end users entirely (CLAUDE.md "Skills — new features documented" checklist item).

**Minor:**
- **`stripXPathBrackets` mishandles chained predicates.** Mendix constraints can be `[a = 1][b = 2]`; the function strips only the outer pair when both ends are brackets, yielding `a = 1][b = 2` → parser recovers, silently skipping a valid constraint. Correct for the common single-predicate case; worth a guard or a noted limitation.
- **Observation (not a defect):** the catalog already stores a precomputed `XPathAST TEXT` column the entry struct doesn't expose; `parse_xpath` re-parses from the raw string. Fine since `parse_xpath` accepts arbitrary strings, but worth being aware of.
- **Nit:** the example rule guards with `if node == None`, but the converter never returns Starlark `None` (it returns `kind:"null"` structs), so those checks are dead — the walker relies on `"null"` being in its leaf-kind set instead. Harmless; example only.

**Recommendation: approve once the skill docs are added** for the two builtins, the `xpath_expression` object, and the `expr` AST schema. Implementation, data wiring, and tests are all in good shape — the doc gap is the one thing that shouldn't merge without. The `stripXPathBrackets` chained-predicate case is a good small follow-up.

mdl/linter/context.go
```diff
@@ -823,6 +823,69 @@ func (ctx *LintContext) ScheduledEvents() iter.Seq[ScheduledEvent] {
...
+	ReferencedEntities    string // comma-separated qualified names
+	IsParameterized       bool   // true when XPath contains $variable references
+	UsageType             string // RETRIEVE, SECURITY, DATASOURCE
+	ModuleName            string
+}
+
+// XPathExpressions returns an iterator over all XPath expression entries in the catalog.
+func (ctx *LintContext) XPathExpressions() iter.Seq[XPathExpressionEntry] {
+	return func(yield func(XPathExpressionEntry) bool) {
+		rows, err := ctx.db.Query(`
```
ako, line 846: Minor: this iterator goes straight to `ctx.db.Query` without the `if ctx.db == nil { return }` guard that three other iterators have. It matches the dominant pattern here (16/19 iterators skip it) and `builtinXPathExpressions` already guards `r.ctx == nil`, so this is not a regression — just a latent nil-panic risk if `ctx.db` could ever be nil at this call site. Non-blocking.

mdl/linter/starlark.go
```diff
@@ -815,6 +821,34 @@ func scheduledEventToStarlark(se ScheduledEvent) starlark.Value {
 	})
 }
 
+// builtinXPathExpressions returns all XPath expression entries from the catalog.
+func (r *StarlarkRule) builtinXPathExpressions(_ *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
ako, line 825: Nit: `args`/`kwargs` are accepted but never validated, so `xpath_expressions("oops")` is silently accepted. Consistent with the other no-arg builtins, so cosmetic — an `UnpackArgs("xpath_expressions", args, kwargs)` with no params would reject stray args if you want strictness.

.claude/skills/mendix/write-lint-rules.md
```diff
@@ -298,6 +369,7 @@ Returned by `project_security()`. Returns `none` if no MPR reader is available.
 |----------|-------------|
 | `violation(message, location?, suggestion?)` | Create a violation to return |
 | `location(module, document_type, document_name, document_id?)` | Create a location for a violation |
+| `parse_xpath(s)` | Parse a raw XPath/expression string and return its AST as an `expr` struct tree. Outer `[ ]` are stripped automatically. Parse failures produce a `recovered` root node rather than raising. |
```
ako, line 372: Doc nuance: "Parse failures produce a `recovered` root node rather than raising" is accurate for the common case (unrecognized tokens at the primary position yield a `RecoveredExpr`, and it never raises). But a *partial* parse with trailing garbage returns the partial tree (trailing tokens are dropped and the hints discarded via `ast, _ :=`), and a nil child surfaces as `kind="null"`. Might be worth softening to "rather than raising; malformed input may yield a partial tree or a `null`/`recovered` node."

ako (commented):
**Review summary — approve, no blockers.**

Clean, well-scoped feature. Validated on the branch: `go build ./...`, `go vet ./mdl/linter/...`, and `gofmt -l` are all clean; the new unit + integration tests pass. Verified that the catalog `xpath_expressions` view/`xpath_expressions_data` table and all referenced `exprcheck` AST types/fields exist.

A few minor, non-blocking observations posted inline. One that isn't tied to a line:

- **Commit message inaccuracy** (`1b96f60`): the body claims it adds a `"PERF001": "Performance"` mapping in `report.go`, but `report.go` isn't in the diff and `PERF001` isn't present. `PERF001` is only an *example* rule in the PR description, not shipped — so no mapping is actually needed, but the commit body implies a shipped rule. Suggest correcting the body (or dropping that bullet).

None of these block merge.

PR #692, opened 2026-06-24
feat(lint): expose scheduled_events() query function in Starlark rules

Andries-Smit:
Adds a new `scheduled_events()` built-in to the linter's Starlark API so star rules can inspect scheduled event configuration (name, microflow, interval, enabled flag).

- Extends `LintReader` with `ListScheduledEvents()` (already implemented by MprBackend and MockBackend — no concrete changes needed there)
- Adds `ScheduledEvent` struct and `ScheduledEvents() iter.Seq` on `LintContext`; resolves module names from the catalog `modules` table and converts interval+type to seconds via `intervalToSeconds()`
- Registers `scheduled_events` builtin and adds `scheduledEventToStarlark` conversion in `starlark.go`

### DIFF
--- a/.claude/skills/mendix/write-lint-rules.md
+++ b/.claude/skills/mendix/write-lint-rules.md
@@ -30,6 +30,7 @@ def check():
 | `constants()` | list of constant | All non-system constants |
 | `widgets()` | list of widget | All non-system widgets |
 | `snippets()` | list of snippet | All non-system snippets |
+| `scheduled_events()` | list of scheduled_event | All non-system scheduled events (requires MPR reader) |
 | `attributes_for(entity_qualified_name)` | list of attribute | Attributes for a specific entity |
 | `activities_for(microflow_qualified_name)` | list of activity | Activities for a microflow (requires FULL catalog) |
 | `permissions()` | list of permission | All permissions across all element types |
@@ -181,6 +182,16 @@ def check():
 | `module_name` | string | `"Sales"` |
 | `folder` | string | `"snippets"` — folder path within module |
 | `widget_count` | int | Number of widgets |
+
+### scheduled_event
+| Property | Type | Example |
+|----------|------|---------|
+| `name` | string | `"SE_NightlyCleanup"` |
+| `qualified_name` | string | `"MyModule.SE_NightlyCleanup"` |
+| `module_name` | string | `"MyModule"` |
+| `microflow_name` | string | `"MyModule.MF_NightlyCleanup"` — resolved from catalog; raw UUID when catalog not built |
+| `interval_seconds` | int | `86400` — `0` for unrecognised interval type |
+| `enabled` | bool | `True` if the event is active |
 
 ### attribute
 | Property | Type | Example |
--- a/mdl/linter/context.go
+++ b/mdl/linter/context.go
@@ -24,6 +24,7 @@ type LintReader interface {
 	ListModules() ([]*model.Module, error)
 	ListFolders() ([]*types.FolderInfo, error)
 	GetRawUnit(id model.ID) (map[string]any, error)
+	ListScheduledEvents() ([]*model.ScheduledEvent, error)
 }
 
 // LintContext wraps a catalog and provides rule-friendly APIs.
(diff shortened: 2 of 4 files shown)


ako:
Verified end-to-end on the branch.

**What's solid:**
- **Correct interval mapping.** The `intervalToSeconds` keys match the real metamodel enum `ScheduledEventsIntervalType` (`generated/metamodel/enums.go:1764`) exactly — `Second/Minute/Hour/Day/Week/Month/Year` — so there's no silent `interval_seconds: 0` bug. (`Month`=30d, `Year`=365d are calendar approximations, inherent and fine for a lint heuristic.)
- **Data path fully wired.** All fields used (`ContainerID/Name/MicroflowID/Interval/IntervalType/Enabled`) exist on `model.ScheduledEvent`; backends populate `IntervalType` from BSON (`parser_enumeration.go:205`, `modelsdk/scheduledevent_read.go:58`); and `ListScheduledEvents()` is implemented across MPR/modelsdk/mock, so the new `LintReader` method is satisfied everywhere.
- **Excellent tests.** Covers all 7 interval units + 2 unknown-type fallbacks, microflow-name resolution (catalog hit *and* raw-UUID fallback), excluded modules, nil reader, and a full Starlark integration test. All pass.

**🔴 Blockers / should-fix:**
1. **Merge conflict (CONFLICTING).** Simulated the merge — the only conflict is in `mdl/linter/starlark.go`: adding `scheduled_events` widened the longest key, so gofmt re-aligned the whole "Query functions" map, colliding with #685's `constants` insertion. Mechanical to resolve (keep all builtins, re-gofmt), but needs a rebase.
2. **On rebase, switch to `ctx.IsExcluded(moduleName)`.** `ScheduledEvents()` uses the old `ctx.excluded[moduleName]` (`context.go:733`). The branch predates #684, so it looks consistent here — but `main` now has all 8 iterators on the centralized `ctx.IsExcluded(...)`, which also honors the `--modules` filter. After rebase, `ScheduledEvents()` would be the lone iterator still reading the map directly, silently ignoring `-m` (same skew that bit #685). `TestScheduledEvents_ExcludedModules` only exercises the *exclude* path, so it won't catch this — worth an include-filter (`SetIncludedModules`) assertion too.
3. **No skill docs.** All 3 changed files are under `mdl/linter/`; `.claude/skills/mendix/write-lint-rules.md` isn't updated (same gap as #688). The new `scheduled_events()` builtin and the `scheduled_event` object (6 fields) should go in the "Available Query Functions" table + a property table, especially since that skill is synced into user projects.

**Minor:**
- Both `defer rows.Close()` defer to the end of the iterator closure, so the `modules` result set stays open while the `microflows` query runs. Works here (pooled file-based catalog, tests pass), but closing each promptly would be cleaner.
- `intervalToSeconds` returns `0` for an unrecognized type with no signal; harmless given the keys are exhaustive against the enum, but a rule could misread `interval_seconds == 0`.

**Recommendation: request changes** — rebase to clear the conflict, and while resolving, flip `ScheduledEvents()` to `ctx.IsExcluded()` (+ an include-filter test) and add the skill docs. The core implementation and tests are genuinely strong; these are integration/consistency items, not logic defects.

ako:
Re-reviewed the update — all three blockers from my earlier review are resolved, verified on the rebased branch:

1. **Merge conflict resolved.** Properly rebased onto current `main` (merge-base == main head); the `starlark.go` builtins block re-gofmt'd cleanly, keeping `constants` and adding `scheduled_events`. CLEAN/MERGEABLE.
2. **`ctx.IsExcluded(moduleName)`** now used in `ScheduledEvents()` (was the direct `ctx.excluded[...]`), so it's consistent with the other iterators and honors `--modules`.
3. **Include-filter test added** — `TestScheduledEvents_IncludedModules` seeds two modules, calls `SetIncludedModules(["ModA"])`, and asserts exactly `[ModA]`, exercising the path the old exclude-only test missed.
4. **Skill docs added** — `write-lint-rules.md` now lists `scheduled_events()` in the Available Query Functions table plus a `### scheduled_event` property table. Checked against the implementation: all 6 fields documented, correctly noting the `microflow_name` catalog/raw-UUID fallback and `interval_seconds: 0` for unrecognised interval types.

**Verification:** `go vet` clean, full `./mdl/linter` suite green, all 6 scheduled-event tests pass (incl. the new include-filter one). Interval mapping still matches the real `ScheduledEventsIntervalType` enum.

**Remaining (minor, non-blocking, unchanged):** the two `defer rows.Close()` in `ScheduledEvents()` still defer to the end of the iterator closure, so the `modules` result set stays open during the `microflows` query. Works (tests pass on the pooled file-based catalog) — purely a cleanliness nit.

**Recommendation: approve.** The review items (conflict, `--modules` consistency + test, docs) are all properly addressed and verified.

PR #694, opened 2026-06-25
docs(microflows): add batch-commit-after-loop performance pattern

daming-bbc:
## Summary

Adds a **Performance: Batch Commit After Loop** section to `.claude/skills/mendix/write-microflows.md`, immediately after the existing `### LOOP Statements` notes.

## Problem

Committing inside a loop issues one database transaction per record — N round-trips for N objects. This is a common AI-generated antipattern that silently degrades performance at scale.

## Solution documented

Three-step pattern:

1. **Before loop**: `$XxxList = create list of Module.Entity;`
2. **Inside loop**: `add $NewXxx to $XxxList;` (replaces the per-iteration `commit`)
3. **After loop**: `commit $XxxList with events on error rollback;` — single transaction

Includes a clear ❌ / ✅ code example and a note that the same pattern applies when looping over *changes* to existing objects.

## Motivation

Discovered while fixing three import microflows (`ACT_ImportMaterialBatchesFromGraphStudio`, `ACT_ImportVehiclesFromGraphStudio`, `ACT_ImportWorkOrdersFromGraphStudio`) that each committed inside a SPARQL-result loop. Adding this to the skill file prevents the same antipattern from being generated in future projects.

## Diff

Pure addition — no existing content modified.

### DIFF
--- a/.claude/skills/mendix/write-microflows.md
+++ b/.claude/skills/mendix/write-microflows.md
@@ -435,6 +435,38 @@ end loop;
 - The loop variable type is **automatically derived** from the list type (e.g., `list of Test.Product` → `Test.Product`)
 - CHANGE statements inside loops use the derived type to resolve attribute names
 
+### Performance: Batch Commit After Loop
+
+**CRITICAL**: Do NOT commit inside a loop. Each `commit` inside a loop issues a separate database transaction, which causes N round-trips for N records and degrades performance significantly.
+
+❌ **INCORRECT — commit inside loop (N transactions):**
+```mdl
+loop $Binding in $BindingsList
+begin
+  $NewBatch = create BatteryOntology.MaterialBatch (BatchNo = $BatchNoObj/Value);
+  commit $NewBatch with events;  -- ❌ one DB transaction per record
+end loop;
+```
+
+✅ **CORRECT — create list before loop, commit once after:**
+```mdl
+$BatchList = create list of BatteryOntology.MaterialBatch;
+loop $Binding in $BindingsList
+begin
+  $NewBatch = create BatteryOntology.MaterialBatch (BatchNo = $BatchNoObj/Value);
+  add $NewBatch to $BatchList;   -- accumulate in memory
+end loop;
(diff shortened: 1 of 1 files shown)


ako:
Clean, accurate docs PR — I'd approve it.

Since this edits a synced skill, the thing that matters most is whether the documented MDL is valid, so I verified each construct against the grammar **and** ran `mxcli check` on both code blocks wrapped in a proper microflow:

- `$List = create list of Module.Entity;` ✅ (`createListStatement`)
- `add $X to $List;` ✅ (`addToListStatement`; the skill already uses this at line 499)
- `commit $List with events on error rollback;` ✅ (`commitStatement: COMMIT VARIABLE (WITH EVENTS)? REFRESH? onErrorClause?`, and `ON ERROR ROLLBACK` is a valid `onErrorClause`)
- trailing `-- …` comments ✅ (`LINE_COMMENT`)

**Both the ❌ and ✅ examples pass `mxcli check` (`✓ Syntax OK`).** The pattern is sound Mendix best practice and aligns with the existing CLAUDE.md microflow idioms (append-on-merge / data-processing guidance).

I also cleared the two things that could have been problems:
- **No embed drift.** The embedded copy under `cmd/mxcli/skills/` is gitignored and regenerated by `make build` (`sync-skills`), so editing only the `mendix/` source is exactly right — nothing else needs committing.
- **Emoji style is consistent** — the file already uses ❌/✅ 18 times, so the emoji-consistency concern from the automated review is a non-issue.

**Minor (optional, non-blocking):**
- The diff leaves a double blank line before `## Object Operations` (one stray extra line).
- Completeness: "commit once after the loop" is correct versus the per-iteration antipattern, but for very large N a single commit of a huge list has its own costs (transaction size, lock duration, memory). Mendix guidance sometimes recommends chunked batch commits — a one-line cross-reference to `patterns-data-processing.md` (batch processing) would round it out. Not wrong as written.

**Recommendation: approve.** Accurate, well-scoped, syntax-validated, and stylistically consistent. The two notes are polish, not blockers.

PR #700, opened 2026-06-26
fix(lint): respect lint-config.yaml in mxcli lint CLI and options in Starlark rules

Andries-Smit:
## Summary

- `mxcli lint` CLI command was silently ignoring `lint-config.yaml` — `FindConfigFile`/`LoadConfig`/`ApplyConfig` were only wired into the executor (REPL/MDL script) path, not the CLI subcommand
- Rule `options` from YAML were stored in `RuleConfig` but never delivered to rules (no mechanism existed to pass them)
- `ModuleRoles()` and `RoleMappings()` iterators in `LintContext` did not apply `IsExcluded()`, so Starlark rules using `module_roles()` reported violations from excluded modules

## Changes

**`cmd/mxcli/cmd_lint.go`**
- Wire `FindConfigFile` → `LoadConfig` → `ApplyConfig` after Starlark rules are loaded; `excludeModules` from config merges with `--exclude` flag; `--rules` allowlist still wins last

**`mdl/linter/linter.go`**
- Add `Configurable` interface; `Run` calls `Configure(options)` on any rule implementing it before `Check`

**`mdl/linter/context.go`**
- Apply `IsExcluded()` in `ModuleRoles()` and `RoleMappings()` iterators

**`mdl/linter/starlark.go`**
- `StarlarkRule` implements `Configurable` (stores options map)
- Add `get_option(key, default=None)` builtin so `.star` rules can read per-rule options from `lint-config.yaml` at check time

**`mdl/linter/rules/domain_size.go`**
- `DomainModelSizeRule` implements `Configurable` with `max_entities` option

**Tests**
- New `mdl/linter/config_test.go`: `LoadConfig` YAML parsing (`excludeModules`, `enabled`, `severity`, `options`), `ApplyConfig`, `FindConfigFile`
- Extended `linter_test.go`: `Configurable` interface called before `Check`, not called when options empty
- Extended `domain_size_test.go`: `Configure` with int/float64/unknown key, raises threshold end-to-end

## Test plan

- [ ] `go test ./mdl/linter/...` passes (56 tests)
- [ ] `mxcli lint -p app.mpr` with `lint-config.yaml` in project root respects `excludeModules`, disabled rules, severity overrides
- [ ] Starlark rule calling `get_option("key", default)` reads value from `options:` block in YAML
- [ ] `--exclude` flag still works independently when no config file present

## Yaml example

```yaml
excludeModules:
  - Administration
  - System
rules:
  MPR001:
    enabled: false
  SEC014:
    severity: error
    options:
      min_length: 12
```

https://claude.ai/code/session_01R1yYeYn9rosf8UdAimdatt

### DIFF
--- a/cmd/mxcli/cmd_lint.go
+++ b/cmd/mxcli/cmd_lint.go
@@ -166,6 +166,19 @@ Examples:
 			}
 		}
 
+		// Load lint config file and apply (excludedModules, rule severity/enabled overrides).
+		// Config ExcludeModules merges with --exclude flag values.
+		configPath := linter.FindConfigFile(projectDir)
+		if cfg, err := linter.LoadConfig(configPath); err == nil {
+			if len(cfg.ExcludeModules) > 0 {
+				merged := append(excludeModules, cfg.ExcludeModules...)
+				ctx.SetExcludedModules(merged)
+			}
+			cfg.ApplyConfig(lint)
+		} else {
+			fmt.Fprintf(os.Stderr, "Warning: failed to load lint config: %v\n", err)
+		}
+
 		// If --rules is specified, disable every rule not in the allowlist.
 		if len(onlyRules) > 0 {
 			allowed := make(map[string]bool, len(onlyRules))
--- a/mdl/linter/config_test.go
+++ b/mdl/linter/config_test.go
@@ -1,0 +1,208 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package linter
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func writeYAML(t *testing.T, content string) string {
+	t.Helper()
+	dir := t.TempDir()
+	path := filepath.Join(dir, "lint-config.yaml")
+	if err := os.WriteFile(path, []byte(content), 0o644); err != nil {
+		t.Fatalf("write YAML: %v", err)
+	}
+	return path
+}
+
+func TestLoadConfig_ExcludedModules(t *testing.T) {
+	path := writeYAML(t, `
+excludeModules:
+  - Administration
+  - Anonymous
+`)
+	cfg, err := LoadConfig(path)
+	if err != nil {
+		t.Fatalf("LoadConfig: %v", err)
+	}
+	if len(cfg.ExcludeModules) != 2 {
+		t.Fatalf("want 2 excluded modules, got %d", len(cfg.ExcludeModules))
+	}
+	if cfg.ExcludeModules[0] != "Administration" || cfg.ExcludeModules[1] != "Anonymous" {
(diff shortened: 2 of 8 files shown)


ako:
A genuine bug-fix PR (not just a new builtin), and I verified it end-to-end.

**Fixes three real bugs:**
1. `lint-config.yaml` ignored by the CLI — `Find/Load/ApplyConfig` were only wired into the executor/REPL path; now wired into `cmd_lint.go`.
2. Rule `options` never delivered — new optional `Configurable` interface; `Run()` calls `Configure(options)` before `Check`, only when options are non-empty.
3. `ModuleRoles()`/`RoleMappings()` ignored exclusions — both now apply `ctx.IsExcluded()`, consistent with the other iterators.

Plus a `get_option(key, default)` Starlark builtin and `DomainModelSizeRule` as the first `Configurable` rule.

**Verified:**
- Properly rebased onto current main (merge-base == main head), CLEAN.
- **No spurious warning on the no-config path** (the main regression risk): `FindConfigFile` returns `""`, `LoadConfig("")` hits `os.IsNotExist` and returns the default config with nil error. Confirmed by a targeted test and a real run.
- Comprehensive unit tests (config parsing, `ApplyConfig`, `Configurable` called/not-called, `Configure` int/float64/unknown-key, find-config in root and `.claude/`). Full `./mdl/linter/...` suite green.
- **The untested CLI glue, exercised manually** on `testdata/expr-checker/minimal.mpr` (baseline 40 warnings):
  - `excludeModules: [System]` → 40 → **1** ✓
  - `SEC001: {enabled: false}` → 40 → **2** ✓
  - malformed YAML → **warns + continues** with defaults ✓
  - no config → **40, silent** ✓

**Notes (one recurring, rest minor):**
1. **Docs gap (should-add):** the new `get_option()` builtin and the `lint-config.yaml` `options:`/CLI capability aren't in the user skill `.claude/skills/mendix/write-lint-rules.md` (only in `docs/11-proposals`). Same gap as #688/#692 — add `get_option()` to the builtins table and document the config feature.
2. **Malformed config = warn + silently lint with defaults.** A typo'd config emits only a stderr warning while the lint proceeds and its exit code reflects *default* config, not the user's intent — in CI this can mask a config that never applied. Consider a hard failure for a *present-but-invalid* config (distinct from absent).
3. **`goValueToStarlark` is scalar-only** — list/map option values from YAML silently become `None` in `get_option()`. Fine for current scalar options; worth documenting the limitation.
4. The CLI glue has no automated test (the underlying pieces are well-covered); I validated it manually.

**Recommendation: approve** — the fix is correct, well-tested, and I confirmed the end-to-end behavior including the no-config regression risk. The docs addition is the one thing genuinely worth doing before/with merge; the rest are minor.

PR #706, opened 2026-06-26
Fix GetRawUnit for v1 MPR files: convert UUID to GUID blob

### LINKED ISSUE #705: GetRawUnit fails on v1 MPR files (UUID not converted to GUID blob)
## Bug

`Reader.GetRawUnit()` in `sdk/mpr/reader_documents.go` fails silently on MPR v1 files (Mendix < 10.18) because it passes the UUID string directly to SQLite instead of converting it to a 16-byte GUID blob.

## Root Cause

In `reader_documents.go:466`:
```go
// V1: Read from database
row := r.db.QueryRow("SELECT Contents FROM Unit WHERE UnitID = ?", string(id))
```

The `UnitID` column in v1 MPR files stores 16-byte GUID blobs (with Microsoft byte-swapping). Passing a UUID string like `"b1d5e4c4-502d-4558-b258-952649650c9e"` to SQLite will never match the blob — SQLite treats text and blob as distinct types.

## Expected Fix

Use `types.UUIDToBlob()` before querying, consistent with `getUnitByIDV1()` in `reader_units.go:604-605`:

```go
blob := types.UUIDToBlob(string(id))
row := r.db.QueryRow("SELECT Contents FROM Unit WHERE UnitID = ?", blob)
```

## Note

`GetRawUnitBytes()` in `writer_security.go` already has the correct implementation using `uuidToBlob()`. Only `GetRawUnit()` is affected.

## Impact

Any caller that uses `ListPages()` / `ListSnippets()` / etc. to get IDs and then calls `GetRawUnit(id)` on a v1 MPR will always get `sql: no rows in result set`. This effectively makes `GetRawUnit` unusable on v1 projects.

## Versions Affected

Confirmed in v0.8.0 through v0.12.0.

densmoe:
## Summary

- `GetRawUnit()` in `sdk/mpr/reader_documents.go` passes the UUID string directly to SQLite on v1 MPRs, but `UnitID` is stored as a 16-byte GUID blob — the query never matches
- Added `types.UUIDToBlob()` conversion before the query, consistent with `getUnitByIDV1()` and `GetRawUnitBytes()` which already do this correctly

Fixes #705

## Impact

Any caller combining `ListPages()`/`ListSnippets()`/etc. with `GetRawUnit(id)` on a v1 project (Mendix < 10.18) gets `sql: no rows in result set` for every unit. This makes `GetRawUnit` unusable on v1 MPR files.

## Test plan

- Verified on real v1 MPR files (Mendix 9.24, 10.18, 10.24) ranging from 11MB to 111MB
- Confirmed `GetRawUnit` now returns valid BSON data for all page/snippet/layout units
- Confirmed v2 projects continue working unchanged (v2 path reads from mprcontents folder, unaffected)

### DIFF
--- a/sdk/mpr/reader_documents.go
+++ b/sdk/mpr/reader_documents.go
@@ -462,8 +462,9 @@ func (r *Reader) GetRawUnit(id model.ID) (map[string]any, error) {
 			return nil, fmt.Errorf("failed to read unit contents: %w", err)
 		}
 	} else {
-		// V1: Read from database
-		row := r.db.QueryRow("SELECT Contents FROM Unit WHERE UnitID = ?", string(id))
+		// V1: Read from database — convert UUID to GUID blob for the query
+		unitIDBlob := types.UUIDToBlob(string(id))
+		row := r.db.QueryRow("SELECT Contents FROM Unit WHERE UnitID = ?", unitIDBlob)
 		err = row.Scan(&contents)
 		if err != nil {
 			return nil, fmt.Errorf("failed to read unit from database: %w", err)


densmoe:
Replacing with a new PR that follows the contribution guidelines (correct branch naming, test, changelog entry).

PR #711, opened 2026-06-29
fix: convert UUID to GUID blob in GetRawUnit for v1 MPRs

### LINKED ISSUE #705: GetRawUnit fails on v1 MPR files (UUID not converted to GUID blob)

densmoe:
Closes #705

## What does it do?

`GetRawUnit()` in `sdk/mpr/reader_documents.go` passed the UUID string directly to SQLite on v1 MPR files (Mendix < 10.18), but `UnitID` is stored as a 16-byte GUID blob. The query never matched, silently returning "no rows" for every unit lookup.

The fix adds `types.UUIDToBlob()` before the query, consistent with `getUnitByIDV1()` in `reader_units.go` and `GetRawUnitBytes()` in `writer_security.go` which already have the correct implementation.

## Testing

- `go build ./sdk/mpr/` — passes
- `go vet ./sdk/mpr/` — passes
- `go test ./sdk/mpr/` — all tests pass (including new regression test)
- Regression test (`TestGetRawUnit_V1`) uses a trimmed Mendix 9.24 v1 MPR fixture and verifies all unit IDs returned by `ListAllUnitIDs()` can be retrieved via `GetRawUnit()`
- Manually verified on real v1 MPR files (Mendix 9.24 / 10.18 / 10.24, 11MB–111MB)
- Confirmed v2 MPR projects continue working unchanged (v2 path is unaffected)

## Mendix validation

N/A — this is a read-only code path, no BSON serialization changes.

## Agentic Code Testing

N/A — this is a library-level bug fix, not an MDL command.


ako (approved):
Approve — correct, minimal, well-tested fix. Verified the regression test fails without the fix and passes with it; full sdk/mpr package tests pass. The blob conversion is consistent with getUnitByIDV1() and GetRawUnitBytes().

PR #732, opened 2026-07-03
Fix formatting in migration README diagram

Andries-Smit:
fix diagram allignment


ako (approved):
Width-aware column check confirms every box border now aligns (vertical bars at identical columns across all rows; the final box sits under Phase 5). Fixes the too-wide `Source` box and the under-indented closing box. Swapping the ambiguous-width `▶` for ASCII `-->` also makes the diagram render consistently in monospace. Thanks!

PR #733, opened 2026-07-03
fix(docs-site): remove duplicated src/ in GitHub edit link

Andries-Smit:
The "Suggest an Edit" button Top right on the documentation site does not work.
https://www.mxcli.org/

## Summary
- The docs site's "Edit this page" link pointed to `.../edit/main/docs-site/src/src/<page>` (duplicated `src`) instead of `.../edit/main/docs-site/src/<page>`.
- mdBook resolves `{path}` in `edit-url-template` relative to the book root (`docs-site/`), which already includes the `src/` prefix — so the template's hardcoded `docs-site/src/{path}` doubled it up.
- Fix: change the template to `docs-site/{path}` in `docs-site/book.toml`.

## Test plan
- [ ] `mdbook build docs-site` (or `mdbook serve docs-site`) and confirm the "Edit this page" link on a page (e.g. migration README) now points to `.../edit/main/docs-site/src/migration/README.md` with no duplicated `src`.


ako (approved):
Verified locally: building `docs-site` with the current template produces edit links with a duplicated `src` (`.../docs-site/src/src/...`) — in mdBook v0.5.2 the `{path}` placeholder already includes the `src/` prefix. This fix (`docs-site/{path}`) resolves to the correct file path (e.g. `docs-site/src/appendixes/changelog.md`), fixing the broken "Suggest an Edit" link. Thanks!

PR #792, opened 2026-07-27
Sync ako/main → upstream: RSS-reader + sudoku findings — authoring/check fixes, DESCRIBE round-trips, tracing & catalog observability

ako:
Brings the 17 commits on ako/main since the last sync (#788). Clean merge against
main; each fix has a Go regression test and (where applicable) an MDL repro under
`mdl-examples/bug-tests/`. Two rounds of findings, all verified by the reporters:

### Authoring / write-path fixes (RSS-reader)
- **ALTER PAGE `set X on <builtin>`** now matches property names case-insensitively
  (was a hard error on every built-in widget). `mdl/backend/pagemutator`
- **`call javascript action`** is serialized on the default modelsdk engine (was
  persisted as `-- Empty action` → CE0008/CE0109). Microflows + nanoflows.
- **`create or modify persistent entity`** preserves attribute identity — no more
  destructive DB column drop/re-add on an identical re-run (data loss).
- **`textbox` placeholder + onchange** are supported end-to-end (were silently dropped).
- **Widget action args** accept keyword names (`View`/`Source`/`Item`/`Page`/`Entity`)
  unquoted, matching `call microflow`.

### Checker (`check --references`)
- New **MDL-WIDGET13**: association traversal in a widget expression (`dynamicclasses`
  /`visibleif`/`editableif`) is flagged (CE0117) before the build.
- **JS/Java action parameter names** are validated against the action's declared
  parameters (a wrong/mis-cased name wrote a dangling ref → CE1613).
- Corrected **MDL001** (→ `FIND(...)`) and **MDL044** (aggregate-aware) hints; fixed the
  dead `mxcli syntax expressions` reference; documented fallback-chain CE0111.

### DESCRIBE round-trip
- **textbox placeholder/onchange** and **non-string dynamictext attributes**
  (`toString($currentObject/Attr)` → bare attribute) now round-trip — re-applying a
  DESCRIBE dump no longer corrupts the page (was CE1613).

### Diagnostics / docs
- Distinct **"string cannot span lines"** parser diagnostic (was the misleading
  "double your apostrophes"); corrected the generated **CLAUDE.md** template and the
  `$latestHttpResponse/Content` casing.

### Tooling & observability (sudoku)
- **`mxcli new`** hard-links the binary into new projects (no ~111 MB copy each).
- **`mxcli run --local --trace-otlp <endpoint>`** exports traces to an OTLP collector
  (the console exporter can't do flame charts). Implies `--trace`; user `OTEL_*` wins.
- **Catalog:** `xpath_expressions` is flagged as FULL-only (a fast-mode query warned
  for `activities`/`refs` but not xpath → silent empty results).
- New **`analyze-runtime`** skill (logs + metrics + traces + catalog + the cross-source
  "app warehouse" join) and a catalog direct-attach docs recipe.


ako:
@mkrouwel this PR contains functionality that makes all data queryable through duckdb: catalog + app data in Postgres + logs + metrics, so can query meta data + data + runtime meta data. Thought you might find it useful.

PR #793, opened 2026-07-28
docs(skills): warn that create-or-modify microflow drops Allowed roles (CE0106)

ront85:
## Problem

`CREATE OR MODIFY MICROFLOW` rebuilds a microflow from its MDL, but MDL has no syntax for the **Allowed roles** (`AllowedModuleRoles`) security property — so a round-trip silently resets it to empty. Under Production/Prototype security with Check Security on, any microflow triggered from a page/button/nanoflow/navigation/published service then fails validation with **CE0106** ("At least one allowed role must be selected…").

It's easy to miss because neither `describe microflow` nor `SHOW ACCESS ON MICROFLOW` surfaces this property — the loss only shows up when the Studio Pro build breaks. Hit this in the wild editing an existing action microflow.

## Changes (docs only)

- **write-microflows.md** — new Common Pitfall #8 (capture roles from the `.mxunit` before editing, re-grant with `grant execute` after), a Validation Checklist item, and a CE0106 row in the Common Studio Pro Errors table.
- **manage-security.md** — a note under Microflow Access that `grant execute` is what sets Allowed roles and is the CE0106 remedy after a round-trip.

No code changes.

### DIFF
--- a/.claude/skills/mendix/manage-security.md
+++ b/.claude/skills/mendix/manage-security.md
@@ -109,6 +109,8 @@ grant execute on microflow MyModule.ACT_Customer_Create to MyModule.User, MyModu
 revoke execute on microflow MyModule.ACT_Customer_Create from MyModule.User;
 ```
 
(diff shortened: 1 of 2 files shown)


ako:
Why not fix the bug, the dropped roles?

ront85:
Closing without merging — the underlying bug is already fixed in code, so this doc warning would be stale/unnecessary guidance.

`CREATE OR MODIFY MICROFLOW` preserving `AllowedModuleRoles` across a round-trip was fixed in commit 6e470e0 ("fix: preserve microflow UnitID across DROP + CREATE OR MODIFY"), which covers both the plain in-place modify path (`cmd_microflows_create.go` lines 70-81) and the DROP-then-recreate path. That fix shipped in v0.10.0 (2026-05-12) and is present in every release since, including current v0.15.0/v0.16.0.

Re-filing this as a manual workaround (re-run `grant execute` after every microflow edit) would train people/agents to do unnecessary busywork for a bug that no longer reproduces on current mxcli.

PR #964, opened 2026-08-25
docs(mcp): onboard Studio Pro 11.14 — the release a name diff would miss

ako:
Onboards **Studio Pro 11.14.0** per the procedure in
`docs/03-development/PED_MCP_CAPABILITIES.md` § *Onboarding a new Studio Pro
version*. Captured live over PED and diffed against the committed
`tools-11.13.json` fixture.

## The tool surface looks untouched — and that is the trap

Same **18 tools**, none added, none removed. This is the first release where a
name-only diff would report "nothing happened", and it would have been wrong on
three counts. All three live only in prose:

| | 11.13 | 11.14 |
|---|---|---|
| `ped_update_document` operations | set / add / remove | **+ `call`** (invoke a schema-declared method) |
| `set` on an element-valued property | only while `null`/`undefined` | **may replace a non-null one** |
| Atomicity | "applied atomically; stops on first error" | **"applied independently … removing, overwriting, or renaming can have side effects that persist even when another operation fails"** |

**The argument scan is a measured negative**: across all 18 tools, no added or
removed `inputSchema` property, no changed `default`, no changed `required` set.
So 11.13's `pg_read_page` `depth` trap did **not** recur and no new
`Client.SupportsToolArg` gate is needed. The scan is now written into the
onboarding procedure so the next release can re-run it verbatim.

## Better error info (the visible half)

Failures are reported **per operation**, with index and path, instead of stopping
at the first:

```
ERROR: Update operation failed with errors (in format [<operation index>]<operationPath>: <output>):
'[0]/entities': Property 'entities' is an array. Use 'add' or 'remove' operation instead.
```

Schemas now carry the **Mendix error code** a violated constraint will raise, so
the CE number is knowable *before* the write:

```
name: string & MinLength<1, 'CE0725'>;          // Workflows$Workflow
description: string & MinLength<1, 'MW0006'>;   // Workflows$WorkflowEventHandler
```

`ped_check_errors` itself is unchanged — the improvement is in the write and
schema paths, not the checker.

## Two gaps moved, both probed without mutating the open project

The technique is now documented: send a value that **cannot** succeed and read
*which layer* rejects it.

- **In-place attribute-type change** — the structural refusal is **gone**. Setting
  a bogus attribute type is now a *value-validation* error (`"not a valid concrete
  subtype of DomainModels$AttributeType"`), not 11.13's structural `"only allowed
  to set primitive or reference properties directly"`. Recorded as **moved, not
  closed**: proving it needs a successful mutation, and the data-safety question
  behind the gap (does an in-place type change migrate the column or drop it?) is
  unanswered and is the real gate on `ALTER ENTITY … MODIFY ATTRIBUTE`.
  **`capabilities.yaml` is deliberately unchanged.**
- **Attribute-array reads** now carry `type.$Type` and `documentation` inline, but
  still drop the type's *parameters* — `length` appears only in the leaf read, so
  `enrichReconstructedEntities` is still required. Measured on one String
  attribute; flagged as needing a wider sample before anyone exploits it.

## 11.14 advertises three document types it does not support

`Settings$ProjectSettings`, `Security$ProjectSecurity` and
`Security$ModuleSecurity` are named in the tool descriptions (`ped_find_document`
even says "Read them directly instead") and `ped_get_schema` returns full, richly
documented element schemas for all three. The document API rejects every one.

The **control** is what makes this a finding rather than a guess — in the same
probe, `Navigation$NavigationDocument` reaches the operation layer and reports a
per-operation error, while the three are refused at the *document* layer before
any operation is examined. Documented as a trap, because a reader trusting the
11.14 tool text alone would "fix" the capability notes in exactly the wrong
direction.

## Code change: one stale comment

`navigation.go` claimed `ped_update_document` "applies atomically". That is no
longer true. **The code is fine** — its menu clear and rebuild are already
separate calls, forced by PED's same-array rule — so this corrects the comment
and writes down *why the split must stay*: consolidating it into one call would
put the menu at risk of being left cleared when the rebuild fails.

## Not attributed to 11.14

`ped_get_schema`'s response format changed from structured JSON to TypeScript-like
source text, but the only two samples are the **11.11** fixture and this 11.14
capture, and 11.13's addition of `kind` makes 11.13 the likelier culprit. Recorded
as "release unknown — do not cite as an 11.14 change" rather than guessed at. Both
fixtures are kept side by side.

## Also new

`glob` gained two file domains — `/version-control` (status, commit history with
per-commit model changes, pending changes) and `/edc-schemas` (live DB schema for
`DatabaseConnection` documents). Neither is used by the backend; `/version-control`
is a plausible source for a future `--mcp` drift command.

## Untested, and why

Whether `ped_update_document` can now **write** a page (the system prompt dropped
the "pages must use `pg_*`" rule, and `ped_read_document` reads a page fine) needs
a real mutation. The backend keeps using `pg_*`, which still works. Flagged in the
doc with the exact probe to run against a throwaway project.

## Validation

`make build`, `go test ./...`, `make lint`, `make check-mdl`, `make
check-skill-mdl` — all green.


ako:
Filed against the wrong repository by mistake — this targets `ako/mxcli`, not upstream. Reopening there; it will reach mendixlabs through the usual sync PR.

PR #1106, opened 2026-09-15
fix(workflow): resolve ALTER WORKFLOW activity refs by name before caption

ako:
## Problem

Every `ALTER WORKFLOW` op that addresses an activity (e.g. `insert boundary event on bugSplitJump …`) failed with

```
ambiguous activity "bugSplitJump" (2 matches); use @N to disambiguate
```

whenever the workflow also contained `jump to bugSplitJump`. Measured live over the MCP backend against Studio Pro 11.14 (TestApp, `workflow.ZzMxcliBug_TimerBoundaryEvents2`); `bugSplitJump@1` worked.

## Cause

`buildJumpTo` (`mdl/executor/cmd_workflows_write.go`) names every jump `JumpTo` and, when the MDL gives no caption, sets `Caption` to the target's name. Both activity resolvers — `mcpWorkflowMutator.searchActivities` (MCP) and `findActivitiesRecursive` / `findActivityIndexRecursive` (MPR, `wfmutator`) — matched `name == ref || caption == ref` into a single pool, so the jump's caption shadowed its target's name. The MPR backend had the same defect; it just had not been reported.

## Fix

Both backends now resolve in two tiers: activities whose **name** equals the reference win; **captions** are consulted only when no activity carries that name. Ambiguity and `@N` are judged within the chosen tier, in the unchanged DESCRIBE traversal order.

Behaviour change: a reference that was ambiguous only because of a caption now resolves without `@N`. `X@2`, which previously reached the jump, now reports not found — the jump stays addressable by its own name.

The jump's default caption is deliberately left alone: existing projects already carry that shape (so the resolver had to be fixed regardless), and changing the default would rewrite every mxcli-authored jump on the next re-run. DESCRIBE already suppresses the derived caption.

## Tests

- `mdl/backend/mcp`: `TestWFResolve_NameBeatsJumpCaption` (fake PED) — boundary event lands on the named activity; caption fallback still works; two caption-only matches remain ambiguous and `@2` selects the second.
- `mdl/backend/wfmutator`: `TestWorkflowMutator_FindActivity_NameBeatsJumpCaption` — covers both `findActivityByCaption` and `findActivityIndex`.

**Control:** both tests were written first and fail against the pooled resolver with the reported `ambiguous activity` message; they pass with the fix. `mcp`, `wfmutator` and executor workflow/jump tests pass; `gofmt`, `go vet`, `make check-findings` clean.

Finding recorded in `.claude/skills/fix-issue/findings/mdl-backend.jsonl`.

Not yet re-verified live against Studio Pro after the fix.


ako:
Opened against the wrong repository by mistake (intended for the ako/mxcli fork). Closing — sorry for the noise.

Issue #643, opened 2026-06-05
Datasource-typed pluggable-widget property passes 'check' but is silently dropped at 'exec' (CE0642)


marclehane-creator:
**mxcli version:** `v0.11.0 (2026-05-21)` · **OS:** Windows 11 · **Mendix project:** runtime 11.10.0 · Combo Box widget `com.mendix.widget.web.combobox.Combobox` installed + `mxcli widget init` run

## Summary
A `datasource`-typed property on a `PLUGGABLEWIDGET` is accepted by `mxcli check` (even with `-p` against the project's real widget defs) and applied by `mxcli exec` without error, but is **not persisted** — `mx check` then reports the property as missing. Concretely, an association Combo Box (the Mendix 11 replacement for the Reference Selector) cannot be authored for a child-context reference.

## Setup
Association `Administration.AccountPasswordData_Account` (parent `AccountPasswordData`, child `Account`). The Combo Box pluggable is installed and `mxcli widget init -p <App.mpr>` has been run.

## Minimal repro
Apply with `mxcli exec -p <App.mpr> repro.mdl`, then `mx check`:

```sql
create or modify page Administration.MxfCombo (
  Title: 'T', Layout: Atlas_Core.Atlas_Default,
  Params: { $Obj: Administration.AccountPasswordData }
) {
  dataview dv (DataSource: $Obj) {
    pluggablewidget 'com.mendix.widget.web.combobox.Combobox' cb (
      optionsSourceType: association,
      attributeAssociation: Administration.AccountPasswordData_Account,
      optionsSourceAssociationDataSource: Administration.Account,
      optionsSourceAssociationCaptionAttribute: Administration.Account.FullName
    )
  }
}
```

## Expected
`optionsSourceAssociationDataSource: Administration.Account` is stored as the combo's options datasource and the widget validates.

## Actual
- `mxcli check` (incl. `-p`) → passes.
- `mxcli exec` → applies cleanly, no error.
- `mx check` → `CE0642 "Property 'Entity' is required." at Combo box 'cb'` — i.e. the datasource property was dropped, so the options source has no entity.

## Related
`optionsSourceAssociationCaptionType` is rejected by `check` as an unknown property *even with* `-p` (it is a valid combo property that defaults to `attribute`), which suggests the derived widget-property surface for `datasource`/conditional properties is incomplete.

## Impact
[...]

PR #1115, opened 2026-09-16
fix: support named datasource mappings for multi-source widgets

### LINKED ISSUE #1109: Follow-up to #643: support named datasource mappings for multi-source pluggable widgets
Follow-up to #643, whose safety fix intentionally rejected named datasource values and retained the generic DataSource clause as a workaround.

The remaining capability gap is now independently reproduced and implemented on a fork: a pluggable-widget schema may declare several independent list datasource properties. A single generic DataSource value cannot represent them without fanning the same source out to every mapping, and DESCRIBE PAGE cannot round-trip them without retaining each schema key.

## Expected behaviour

- Accept a real datasource expression under a declared datasource property key; continue rejecting scalar lookalikes.
- Resolve each datasource mapping by its key or alias before falling back to the generic single-source convenience form.
- Select datasource-dependent widget modes from the mappings for that mode, rather than from an unrelated widget-wide slot.
- Maintain entity context per mapping so dependent attributes bind to the correct entity.
- Emit named datasource keys from DESCRIBE PAGE whenever collapsing to one DataSource clause would lose source identity.

## Scope

List datasources only. Mendix permits several list datasource properties but at most one object datasource property.

## Validation

The fork implementation includes unit coverage for several simultaneous named sources, a public ComboBox regression fixture, a real Mendix 11.12.3 build with zero consistency errors, make build/test/lint/check-findings, and a round-trip assertion.

Proposed implementation: https://github.com/tjorim/mxcli/pull/2

This is deliberately a follow-up rather than a duplicate: #643 closed the silent-data-loss defect; this restores the deferred general named/multi-source capability safely.

tjorim:
Closes #1109

Follow-up to #643. That issue eliminated silent data loss by rejecting named datasource values and retaining a generic DataSource workaround; this PR safely implements the deferred general capability.

## What changes

- Resolves a datasource by its declared mapping key or alias before the generic fallback.
- Allows real datasource expressions in named datasource properties, while rejecting scalar lookalikes.
- Activates datasource-dependent modes from the mappings belonging to that mode.
- Preserves independent entity context for datasource-bound properties.
- Emits named datasource keys from DESCRIBE PAGE when a generic DataSource clause would lose identity.
- Prevents the generic fallback from copying one datasource into every slot of a multi-source mode.
- Preserves unsupported named datasource values as an explanatory DESCRIBE comment.

The generic DataSource form remains backward-compatible for a single-datasource mode.

## Validation

- make build, make test, make lint, and make check-findings passed on the final rebased branch.
- Regression and unit coverage cover named datasource mode selection, scalar alias rejection, distinct multi-source mappings, unsupported-source describe output, and round-trip MDL parsing.
- Public ComboBox fixture executed successfully.
- Tested with Mendix 11.12.3 on a disposable project copy: mxcli docker check --no-update-widgets reported 0 errors.

## Agentic use

The custom-widget skill documents the neutral multi-source pattern, and the regression fixtures avoid application-specific schemas.


tjorim:
I want to record my disappointment with how this was handled.

I opened #1109 before implementation began, documented the intended behavior and validation, and linked a working fork PR. The parallel implementation that landed through #1117 covers substantially the same scope—its commit history even references #1109—yet there was no discussion of the existing implementation, no request to collaborate, and no acknowledgement of it.

That leaves this PR conflicted and largely redundant after significant investigation, implementation, tests, documentation, and Mendix validation. The repository’s CONTRIBUTING guide says the issue/approval/assignment workflow exists to prevent duplicate work; regardless of the missing human approval/assignment on #1109, this is precisely the outcome that process is meant to avoid.

Please clarify how contributors should avoid this in future, especially when an issue already contains a linked implementation or open PR.

